PR #5, opened 2019-12-27
Re-consider graceful shutdown

marshall-lee:
- Dispatch OS signals to do graceful/fast shutdown.
- `App.Wait()` now really waits for everything to finish using WaitGroup/Cond.
- `App.Close()` does the force shutdown.
- `ctx` is removed as a struct member. Passing it as a parameter instead.

Don't know good it or not but App is now fully re-startable so we can do `Start()`/`Wait()` with something like exponential backoff.
@fspmarshall what do you think?

### DIFF
--- a/access/slackbot/slackbot.go
+++ b/access/slackbot/slackbot.go
@@ -26,8 +26,10 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"os/signal"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/gravitational/kingpin"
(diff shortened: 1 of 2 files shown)


access/slackbot/slackbot.go
```diff
@@ -91,101 +93,213 @@ func run(configPath string) error {
...
-	<-a.ctx.Done()
+	a.Lock()
+	defer a.Unlock()
+	if cond := a.waitCond; cond != nil {
+		cond.Wait()
+	}
+}
+
+// Error returns the error returned by watcher/http server or them both
+func (a *App) Error() error {
```
fspmarshall: This method should be protected by mutex.

PR #8, opened 2019-12-30
Reconnect watcher when the stream closes or server shuts down

### LINKED ISSUE #2: Consider watcher re-connection in Slackbot
Problem: when the watcher connection breaks, the whole bot process fails.
We should somehow re-establish a broken connection.

marshall-lee:
Closes #2.

There're two cases when watcher stream could be closed:

- When the cached enabled for Auth Server. It initiates stream closing every 600 seconds. We check this case using `IsEOF(err)`.
- When Auth Server goes offline. Then we should start another watcher. We check this case using `IsConnectionProblem(err)`.

Further, we should consider doing `GetRequests(PENDING)` after every successful reconnection not to miss any requests created between disconnection and reconnection. Not sure how to do this using the current API though. Must we fetch all the pending requests and match them against their presence in the cache? How we can be sure that there's not so large amount of records?

Another thing I want to discuss is that I've set `grpc.WaitForReady(true)` as a default call option for all the grpc calls. It's disabled by default in the grpc library, and as a library's default it's totally sane, but since the `access` package is a wrapper, it could be convenient to not have disconnection errors on every call. With this setting, if server goes offline the call like `GetRequest()` would block, but it's a caller's responsibility to control it using a `ctx` with timeout. You can see how I do this for the entire incoming http request and return 503 code in the case when timeout is exceeded. Perhaps, we should even change 503 to 200 and answer to the user with a slack *ephemeral* message saying "I'm sorry, Teleport is offline, cannot process your request". Please let me know what do you think about setting `WaitForReady(true)` as a default option for the `access` package!

### DIFF
--- a/access/slackbot/slackbot.go
+++ b/access/slackbot/slackbot.go
@@ -109,7 +109,6 @@ func serveSignals(app *App) {
 	signal.Notify(sigC,
 		syscall.SIGQUIT, // graceful shutdown
 		syscall.SIGTERM, // fast shutdown
-		syscall.SIGKILL, // fast shutdown
 		syscall.SIGINT,  // graceful-then-fast shutdown
 	)
 	var alreadyInterrupted bool
@@ -426,7 +425,7 @@ func (a *App) OnCallback(ctx context.Context, cb slack.InteractionCallback) (sla
 		status,
 	)
 	message := cb.OriginalMessage
-	message.Blocks.BlockSet = []slack.Block { msgSection(msg) }
+	message.Blocks.BlockSet = []slack.Block{msgSection(msg)}
 	return message, nil
 }
 
@@ -470,7 +469,10 @@ func (a *App) ActionsHandler(ctx context.Context, w http.ResponseWriter, r *http
 	} else {
 		w.Header().Add("Content-type", "application/json")
 		w.WriteHeader(http.StatusOK)
-		json.NewEncoder(w).Encode(&msg)
+		err = json.NewEncoder(w).Encode(&msg)
+		if err != nil {
+			log.Error("Failed to write JSON: ", err)
+		}
 	}
 }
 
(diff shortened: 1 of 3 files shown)


access/slackbot/slackbot.go
```diff
@@ -273,8 +272,19 @@ func (a *App) Start(ctx context.Context) error {
...
 		defer shutdown()
 		defer workers.Done()
-		if err := a.watchRequests(ctx); err != nil {
-			watcherErr = trace.Wrap(err)
+		for {
+			log.Info("Starting a request watcher...")
+			if err := a.watchRequests(ctx); err != nil {
+				if trace.IsEOF(err) || trace.IsConnectionProblem(err) {
+					log.Error("Watcher connection closed: ", err)
+					continue
```
fspmarshall: I think a backoff mechanism is warranted here, could be something simple like a linearly increasing sleep which resets after a short cooldown.

fspmarshall:
>  Please let me know what do you think about setting WaitForReady(true) as a default option for the access package!

This seems fine to me, so long as the timeouts used are fairly short (I'd rather the client be over-sensitive and need to be restarted than have it suppress important errors).

>  it could be convenient to not have disconnection errors on every call.

Was this happening to you, or is this a hypothetical?  This sounds like a deeper issue.

> we should consider doing GetRequests(PENDING) after every successful reconnection not to miss any requests created between disconnection and reconnection.

I agree, but I think this is a lower priority feature.  Features like persistent backend and user identity logging are much more important.  Missing a request on restart just means that the user will need to re-send it, which is a minor pain point since this should be a fairly rare occurrence.

When we do get around to it, I think you're right that comparing with the cache would be easiest.  We'd probably want to limit the requests displayed in this manner by time (e.g. any request created more than 30-40 seconds in the past can probably be ignored).  This would ensure continuity during brief outages without spamming the channel in cases of more serious failure.

access/slackbot/slackbot.go
marshall-lee: grpc library has its internal backoff implementation. I configure it via `grpc.WithBackoffMaxDelay(time.Second*2)` option.

With `WaitForReady(false)` it has no effect but combining it with `WaitForReady(true)` this loop would block on next iteration until the server is up.

PR #16, opened 2020-01-20
fix typo in readme.md

cpanato:


cpanato:
sounds very good @benarent , i will take a look and see how it works

PR #17, opened 2020-01-22
Better config support for ./slackbot start

xnutsive:
When using `./slackbot start`, you currently have to add an unnamed arg with a path to the config file, like this: `./slackbot start config.toml`. It won't work with `-f config.toml`.

This is inconsistent with how other Teleport services work: 
- Teleport daemon expects /etc/teleport.yaml
- `tctl` actions expect an unnamed argument, or a `-f` argument.

1. Should we switch to yaml so all the configs are in the same format? 
2. Should Slackbot also expect a config in /etc/teleport-slackbot.yaml by default? 
3. Should it parse both unnamed and `-f` argument? 

/cc @fspmarshall, @klizhentas.


xnutsive:
@marshall-lee, I couldn't help myself and wrote a few words in golang.

PR #38, opened 2020-01-24
Migrate to go modules

DarthSim:
Initial `go.mod` is taken from https://github.com/gravitational/teleport/pull/3299.
Don't merge until https://github.com/gravitational/teleport/pull/3299 is merged


klizhentas:
we are not going to be migrating to go modules for teleport in the short term, so closing this and teleport #3299 one as well.

PR #44, opened 2020-01-29
Dependency management using dep

DarthSim:
`Gopkg.toml` & `Gopkg.lock` are taken from teleport mater.
teleport repository contains private git submodules, so I had to add a wrapper over dep and git that ignore submodules.


dev/git
```diff
@@ -0,0 +1,16 @@
...
+# github.com/gravitational/teleport contains private git submodules that can't be
+# checkouted by dep, so we have to use git wrapper that ignores submodules
+
+if [ -z $_GIT_ORIGIN ]; then
+  echo "Don't use this wrapper directly";
+  exit 1;
+fi
+
+if [ $1 == "submodule" ]; then
+  echo "Ignoring submodules";
```
marshall-lee: It's a neat hack 👍 
Though i'm not sure that it's okay to include these scripts in the repo.
Another thing that bothers me: need to limit disabling of `git submodule` command *only* to `github.com/gravitational/teleport`. Other repos could use `git submodule` too.

Gopkg.toml
```diff
@@ -0,0 +1,74 @@
...
+[[override]]
+  name = "github.com/cenkalti/backoff"
+  version = ">=1.0.0, <=19.0.0-g5d150e7"
+
+[[constraint]]
+  name = "github.com/gravitational/kingpin"
+  revision = "742f2714c145075a68df26ccc25bc05545310f3a"
+
+[[override]]
+  name = "github.com/prometheus/procfs"
```
marshall-lee: There're many Teleport-specific stuff here. I think it's not necessary to put all of this here. Lets just leave here only direct dependencies of plugins

dev/git
DarthSim: `go mod` doesn't support git submodules, so I don't think any of pkg would use them. I'd offer to keep it as is until we run into a problem.

Gopkg.toml
DarthSim: I left only teleport-specific `override`s here that affect nested dependencies. If you delete those overrides and try to do something like `dep ensure -add some_package`, `dep` will break versions compatibility.

`Gopkg.toml` is not like `Gemfile`, it contains mostly rules, not requirements. If it contains `override` or `constraint` of some package, this doesn't mean these package **will** be vendored.

DarthSim:
Moved wrappers to `scripts/`

PR #36, opened 2020-01-31
Update the install script to create /var/lib/teleport/plugins/ 

xnutsive:


access/slackbot/install
```diff
@@ -5,10 +5,18 @@
...
+# and certificates
+#
+DATADIR=/var/lib/teleport/plugins/slackbot
+
 [ ! $(id -u) != "0" ] || { echo "ERROR: You must be root"; exit 1; }
 cd $(dirname $0)
 mkdir -p $BINDIR
 cp -f slackbot $BINDIR/ || exit 1
 
+mkdir -p $DATADIR
```
marshall-lee: You forgot `|| exit 1` part :)

access/slackbot/install
```diff
@@ -5,10 +5,18 @@
...
 
+#
+# the directory where Teleport Plugins store their data files
+# and certificates
+#
+DATADIR=/var/lib/teleport/plugins/slackbot
+
 [ ! $(id -u) != "0" ] || { echo "ERROR: You must be root"; exit 1; }
 cd $(dirname $0)
 mkdir -p $BINDIR
```
marshall-lee: Here need to check the result too

access/slackbot/install
marshall-lee: Also it's better to place this command *after* the `Teleport Slackbot binaries have been copied to $BINDIR` message.

marshall-lee: Either you could do `mkdir $BINDIR $DATADIR` at once. It case of failure, `mkdir` prints what particular write operation haven't succeed

xnutsive: Thanks for catching this, it'd make the script more elegant.

xnutsive:
@marshall-lee thank you for reviewing it and pointing out possible issues! Fixed both.

PR #47, opened 2020-02-01
Check cert expiration date on launch

### LINKED ISSUE #37: Check certificate validity on startup
I came across the same issue as #21 again, and it was due to exporting a very short license. Can we add a check that'll check the certificate validity, and inform the user that the plug cert has expired. 

```
teleport-421-proxy ~/teleport-access-slackbot: openssl x509 -in  /var/lib/teleport/plugin/plug.crt -text -noout
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            99:f4:67:be:80:2d:28:bb:a9:ba:21:35:b4:9d:d4:f4
    Signature Algorithm: sha256WithRSAEncryption
        Issuer: O = teleport-421-auth, CN = teleport-421-auth, serialNumber = 15083136544899289883549370232550266955
        Validity
            Not Before: Jan 21 22:33:50 2020 GMT
            Not After : Jan 22 10:34:50 2020 GMT
```

```
INFO[0000] Starting a request watcher...
ERRO[0005] Watcher connection error: watcher initialization timed out. Reconnecting...
ERRO[0010] Watcher connection error: watcher initialization timed out. Reconnecting...
```

xnutsive:
I just wanted to play around with some go code, so please don't take this too seriously. However, in case the code makes sense, and in case it's an OK place to put the certificate check, this should fix #37. 

What do you think?

### DIFF
--- a/access/slackbot/main.go
+++ b/access/slackbot/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"os"
 	"os/signal"
@@ -60,9 +61,9 @@ func main() {
 	utils.InitLogger()
 	app := kingpin.New("slackbot", "Teleport plugin for access requests approval via Slack.")
 
-	app.Command("configure", "Prints an example .TOML configuration file")
+	app.Command("configure", "Prints an example .TOML configuration file.")
 
-	startCmd := app.Command("start", "Starts a bot daemon")
+	startCmd := app.Command("start", "Starts the Teleport Slack plugin")
 	path := startCmd.Flag("config", "TOML config file path").
 		Short('c').
 		Default("/etc/teleport-slackbot.toml").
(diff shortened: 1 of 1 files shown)


access/slackbot/main.go
```diff
@@ -225,6 +226,16 @@ func (a *App) Run(ctx context.Context) error {
 
 	ctx, cancel := context.WithCancel(ctx)
 	a.cancel = cancel
+	
+	clientCert, err := x509.ParseCertificate(a.tlsConf.Certificates[0].Certificate[0])
```
marshall-lee: I don't feel it's the right place to do this check. Here, we must not assume that certificate is located in `Certificates[0]`. Having arbitrary `tls.Config`, we cannot rely on the certificate chain length.

Instead of this, I'd rather check the certificate when it's being loaded from the file.

access/slackbot/main.go
```diff
@@ -225,6 +226,16 @@ func (a *App) Run(ctx context.Context) error {
...
+	if err != nil {
+		cancel()
+		return trace.Wrap(err)
+	}
+
+	if time.Now().After(clientCert.NotAfter) {
+		log.Error("Teleport Auth certificates are expired.")
```
marshall-lee: I think it should be `log.Warning` since we don't terminate the app on such error.

access/slackbot/main.go
marshall-lee: It's also good to save the result of `x509.ParseCertificate` to the `Leaf` field of the certificate. Having this, `tls` implementation won't parse this data twice.

marshall-lee:
@xnutsive Thanks for the inspiration, I took your idea but did everything differently. Look  3f1ef52

PR #53, opened 2020-02-12
Consider graceful shutdown for outgoing HTTP-requests too

marshall-lee:
We already have `httpServer.Shutdown()` to ensure that all incoming HTTP requests are finished before we terminate the process.

However, we don't do the same for outgoing HTTP requests. On termination, all the background operations are simply halted by context cancellation. To challenge this problem, I extended the use of existing `WaitGroup` and introduced the `spawn(func(ctx)...)` method to execute goroutines as jobs that could be waited to finish.

### DIFF
--- a/access/slackbot/main.go
+++ b/access/slackbot/main.go
@@ -134,34 +134,41 @@ func run(configPath string, insecure bool, debug bool) error {
 }
 
 func serveSignals(app *App) {
+	ctx := context.Background()
 	sigC := make(chan os.Signal)
 	signal.Notify(sigC,
-		syscall.SIGQUIT, // graceful shutdown
-		syscall.SIGTERM, // fast shutdown
+		syscall.SIGTERM, // graceful shutdown
 		syscall.SIGINT,  // graceful-then-fast shutdown
 	)
 	var alreadyInterrupted bool
+	gracefulShutdown := func() {
+		tctx, tcancel := context.WithTimeout(ctx, 2*time.Second)
+		defer tcancel()
+		log.Infof("Attempting graceful shutdown...")
+		if err := app.Shutdown(tctx); err != nil {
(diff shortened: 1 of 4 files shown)


access/slackbot/main.go
```diff
@@ -263,49 +260,71 @@ func (a *App) Run(ctx context.Context) error {
...
+			jobs.Done()
+		}()
+	}
+
+	jobs.Add(1) // Start the "app" job. We have to do it for Wait() not being returned beforehand.
+	doneCh := make(chan struct{})
+	go func() {
+		jobs.Wait()
+		close(doneCh)
+	}()
```
DarthSim: I'd move this inside of `shutdown` and use `jobs.Wait()` here https://github.com/gravitational/teleport-plugins/pull/53/files#diff-b0e8b9ea124c13ad8e2c86d4e5280a93R327. This way we won't cast 2 runtime-long goroutines. And code IMHO will be a bit more readable.

marshall-lee: I re-thinked the `Shutdown` method, now in accepts the context. I cannot do `jobs.Wait()` there.

access/slackbot/main.go
```diff
@@ -263,49 +293,63 @@ func (a *App) Run(ctx context.Context) error {
...
+		}()
+	}
+
+	jobs.Add(1) // Start the "app" job. We have to do it for Wait() not being returned beforehand.
+	a.doneCh = make(chan struct{})
+	doneCh := a.doneCh // Save the variable to access it when unlocked
+	go func() {
+		jobs.Wait()
+		close(doneCh)
+	}()
```
DarthSim: It looks like we don't even need this goroutine. You can just defer this channel closure and use `jobs.Wait()` in the end

marshall-lee: 👍

PR #73, opened 2020-04-07
Add Github Actions for running tests

webvictim:
Testing Github actions


xnutsive:
Why not just `ls -l access/` and run tests for all of the plugins? This would make sure all the plugins are tested, even if we forget to update the job. 

Also, why not run the CI on _all_ pull requests? If I read the code correctly, the current version only runs on PRs that have `release/*` in branch name? 

Overall, this is much needed — Force only tests slack and Jira plugins, and we already have 3 more.  

I think @marshall-lee would be willing to help with this, if you have too much stuff on your plate! Let us know ;) /fyi @benarent

webvictim:
Yeah, I actually wanted to do something like this, but it seems the example test is broken:

```
~/go/src/github.com/gravitational/teleport-plugins(master*) » cd access && go test ./...
?       github.com/gravitational/teleport-plugins/access[no test files]
# github.com/gravitational/teleport-plugins/access/example
example/example.go:73:15: assignment mismatch: 2 variables but client.WatchRequests returns 1 values
```

If you fancy fixing the test I’ll definitely change this so it runs all tests in the directory automatically.

Also FYI we’re moving away from Force - we’re in the process of rebuilding all our Teleport CI on a new platform, probably Drone. This will be handling the release pipeline for teleport-plugins too. I’m hoping to have it up and running in a couple of weeks.

xnutsive:
We’ll def fox the example test and we could merge something like this as a temporary solution, until the drone setup is ready.

I’ll ask Vlad to take a look or fix it myself on the weekend ;-)

Nate
(650) 933-7237

webvictim:
I got Drone up and running quicker than expected, so merged the pipeline in. It should run against all pushes/PRs to master.

PR #75, opened 2020-04-28
Verify client certificate for PagerDuty webhooks

marshall-lee:
Hi!

Unlike other services we have intergated, PagerDuty does not implement anything like HMAC signature of the webhook payload to ensure that webhook comes from PagerDuty but not from the bad guys. Instead of HMAC it provides a bunch of options:

1. IP Safelists.

https://support.pagerduty.com/docs/whitelisting-ips#section-webhooks

Requires polling of https://app.pagerduty.com/webhook_ips endpoint. We don't have it yet and I don't think we should. I don't find it reliable because when running a bot behind a proxy we must be sure that it sets `X-Forwarded-For` header correctly, and also this IP list itself could be changed without a notice.

2. HTTP Basic authentication.

In some sense it's weak too but it's good to have it if there's no other option. Basic auth support was added in 7b6df598159dceb4cf3815b0a9261743118088d0 and must work in all the plugins that we have, though I'm not sure that all services we've integrated support URLs in form of "http://user:password@example.com" but PagerDuty promises they do - https://developer.pagerduty.com/docs/webhooks/webhook-behavior/#basic-authentication.

To activate basic auth for PagerDuty, simply set `base-url` to something like `"https://username:password@example.com"` (cc @xnutsive it must be reflected in PagerDuty integration docs as one of the possible measures for authentication).

3. Mutual TLS aka client certificate verification.

This is the most interesting thing which is described here: https://developer.pagerduty.com/docs/webhooks/webhooks-mutual-tls

It appears that during TLS handshake we could require a client to provide a *client* certificate and if it's a valid certificate from trusted CA then we also check its `Subject` field for a specific value `"CN=webhooks.pagerduty.com,O=PagerDuty Inc,L=San Francisco,ST=California,C=US"`.

I added a certificate verification support in 4100863.

@xnutsive:

In the docs, we must insist on "Mutual TLS" as a **recommended** way to authenticate incoming requests from PagerDuty. There're two cases we must consider:

1. Customer runs `teleport-pagerduty` with HTTPS ON.

Then, to activate mutual tls one must set `verify-client-cert = true` in the `[http.tls]` section of the config file.

2. Customer runs `teleport-pagerduty --insecure-no-tls` with HTTPS OFF behind a reverse proxy that does SSL termination.

In this case certificate verification, if possible, must be done on the proxy side. Examples on how to set it up for NGINX and Apache are provided in PagerDuty docs about mutual tls and we should refer this link in our docs.

### DIFF
--- a/access/pagerduty/README.md
+++ b/access/pagerduty/README.md
@@ -59,6 +59,21 @@ _Note: by default, tctl auth sign produces certificates with a relatively short 
 
 In your Pagerduty dashboard, go to **Configuration -> API Access -> Create New API Key**, add a key description, and save the key. We'll use the key in the plugin config file later.
 
+
+### Securing Pagerduty webhooks
+
+Pagerduty doesn't have a mechanism to sign it's webhook payload. Instead, they provide two good ways for you to verify the integrity and origin of the webhook requests (i.e. that the webhook is actually sent by Pagerduty, not bo something else): 
+
+- Basic auth (not recommended)
+- Certificate verification (recommended and default).
+
+To setup Basic Auth, setup a usename and password in the config below. in `[http.basic-auth]` section. 
+
+If you intend to run `teleport-pagerduty` with TLS anyway, then to ensure mutual TLS verification, you need to setup `verify-client-cert = true` in the config below in `[http.tls]` section.
+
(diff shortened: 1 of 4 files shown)


xnutsive:
Thanks for the notes on how to change the configs and docs! I'll add the docs and merge the PR once I'm done, probably later tonight.

xnutsive:
@marshall-lee, please review my readme and example changes and merge this ;)

PR #77, opened 2020-05-04
Various fixes

webvictim:
This implements the same thing as https://github.com/gravitational/teleport-plugins/pull/78 with some extra parts to make CI better (Makefile targets for test, linting and releasing all plugins at once)

I’ve implemented the same lint rules as we’re using for Teleport and fixed the errors the linter found here (related to `goimports` and `bodyclose`)

Drone changes:
- lint and run tests on every PR
- lint, run tests, build artefacts and push to S3 for every tag published

~The only part I’m not happy with is the change to `utils.GetFreeTCPPorts` under `vendor` - this should be made in the Teleport repo and then this codebase revendored to pick it up, but as a number of Teleport integration tests depend on this method too we need to be careful when changing it.~


Makefile
```diff
@@ -14,14 +16,55 @@ access-mattermost:
...
-.PHONY:
+.PHONY: access-example
 access-example:
 	go build -o build/access-example ./access/example
 
-.PHONY:
+# Run all tests
+.PHONY: test
+test:
+	go test -count 1 ./...
```
awly, line 26: nit: consider adding `-race`

Makefile
```diff
@@ -1,4 +1,6 @@
-.PHONY:
+GO_LINTERS ?= "unused,govet,typecheck,deadcode,goimports,varcheck,structcheck,bodyclose,staticcheck,ineffassign,unconvert,misspell"
```
awly, line 1: consider adding `golint`

access/jirabot/jirabot_test.go
```diff
@@ -98,7 +98,7 @@ func (s *JirabotSuite) SetUpTest(c *C) {
 }
 
 func (s *JirabotSuite) TearDownTest(c *C) {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*250)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*1)
```
awly, line 101: remove `*1`

access/mattermost/mattermost_test.go
```diff
@@ -97,7 +97,7 @@ func (s *MattermostSuite) SetUpTest(c *C) {
 }
 
 func (s *MattermostSuite) TearDownTest(c *C) {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*250)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*1)
```
awly, line 100: remove `*1`

access/pagerduty/pagerduty_test.go
```diff
@@ -101,7 +101,7 @@ func (s *PagerdutySuite) SetUpTest(c *C) {
 }
 
 func (s *PagerdutySuite) TearDownTest(c *C) {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*250)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*1)
```
awly, line 104: remove `*1`

access/slackbot/bot.go
```diff
@@ -103,6 +103,10 @@ func (b *Bot) Respond(ctx context.Context, reqID string, reqData requestData, st
 		return trace.Wrap(err, "failed to send update: %v", err)
 	}
 
+	if err := rsp.Body.Close(); err != nil {
+		return trace.Wrap(err, "failed to close body: %v", err)
+	}
```
awly, line 108: this should be in a defer, otherwise `ioutil.ReadAll` below will fail.
also, consider adding a unit test for this

access/slackbot/slackbot_test.go
```diff
@@ -98,7 +98,7 @@ func (s *SlackbotSuite) SetUpTest(c *C) {
 }
 
 func (s *SlackbotSuite) TearDownTest(c *C) {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*250)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*1)
```
awly, line 101: remove `*1`

vendor/github.com/gravitational/teleport/lib/utils/utils.go
```diff
@@ -361,18 +361,23 @@ func (p *PortList) PopIntSlice(num int) []int {
...
+		}
+
+		listen, err := net.ListenTCP("tcp", addr)
+		if err != nil {
+			return nil, trace.Wrap(err)
+		}
+		defer listen.Close()
+
+		port := strconv.Itoa(listen.Addr().(*net.TCPAddr).Port)
+		list = append(list, port)
```
awly, line 380: Make a new function outside of `vendor/`. Otherwise, next time someone updates deps this will get overwritten.

webvictim: Added to `utils` in `teleport-plugins` and renamed to `GetFreeTCPPortsForTests`.

Makefile
webvictim: I did try this. Unfortunately it seems that there is actually a race in one of the tests, so I've reverted it for now. I will look into this in some more detail.

Makefile
```diff
@@ -14,14 +16,55 @@ access-mattermost:
...
+release/access-pagerduty:
+	make -C access/pagerduty clean release
+
+# Run all releases
+.PHONY: releases
+releases: release/access-slackbot release/access-jirabot release/access-mattermost release/access-pagerduty
+
+.PHONY: get-deps
+get-deps:
+	go get -v -t -d ./...
```
awly, line 51: isn't everything in this repo vendored?

access/slackbot/bot.go
```diff
@@ -51,6 +51,7 @@ func NewBot(conf *Config) *Bot {
 }
 
 // Post posts request info to Slack with action buttons.
+//nolint(golint)
```
awly, line 54: leave a comment explaining why this is suppressed

utils/utils.go
```diff
@@ -0,0 +1,62 @@
+package utils
```
awly, line 1: https://github.com/golang/go/wiki/CodeReviewComments#package-names
please don't start the same `utils` kitchen-sink package here.
how about `utils/nettest`?

also, https://github.com/golang/go/wiki/CodeReviewComments#package-comments

utils/utils.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"net"
+	"strconv"
+
+	"github.com/gravitational/trace"
+)
+
+// PortList is a list of TCP ports
+type PortList []string
```
awly, line 11: could you reuse this type from `teleport/lib/utils`?

utils/utils.go
```diff
@@ -0,0 +1,62 @@
...
+// ports were allocated
+func (p *PortList) PopIntSlice(num int) []int {
+	ports := make([]int, num)
+	for i := range ports {
+		ports[i] = p.PopInt()
+	}
+	return ports
+}
+
+// GetFreeTCPPortsForTests returns n free ports (which are suggested by the kernel)
```
marshall-lee, line 43: Does it really help running tests in parallel? I see there's still a chance for race condition.

Ideally, the issue should be solved if both Teleport auth server and plugin http server could listen to `:0` and somehow report the allocated port to the app. I even have a half-working prototype for this.

Perhaps leave it alone for now and run tests with `go test -parallel 1`?

awly: This solution doesn't fix the problem, but makes it much less likely.
Actually listening on port 0 would certainly be preferable.

I added some code in https://github.com/gravitational/teleport/pull/3713/commits/13c0b8aceb59c2d9715e44706120746ba1519d4a to expose actual ports from `TeleportProcess` if it helps.

Makefile
webvictim: Yep. My understanding of what vendoring does is still a little broken, I think. Essentially I should only need to use `go get` if there are things which aren't supplied by `vendor/`?

utils/utils.go
webvictim: My goal with this PR was basically to fix the unit tests to the point where they would run in CI without error the majority of the time (unless there was a bug introduced by a PR) which is now the case.

As I understand it, Teleport's use of `gocheck` pretty much prevents us from running tests in parallel anyway so the `-parallel` flag wouldn't really make any difference? Certainly, I ran `go test -count 1 ./...` and `go test -count 1 -parallel 1 ./...` a few times and nothing seemed to be different between the two at all.

@marshall-lee If you'd like to fix the problem 'properly' in future by actually passing Teleport's listening port through to the tests that would be awesome. For now, I'd like to get this solution merged so we have 'working' CI.

Makefile
awly: Yes. In practice you should only be in one of two states:
- nothing is vendored and you use `go get` (possibly with go modules)
- everything is vendored
  - if using go modules, you use `go get` to add new dependencies or change versions of existing dependencies; this is only done by humans and checked in via PRs, automation never calls `go get`

Vendoring is preferred.
Since this repo already does vendoring, all dependencies should be there.

utils/utils.go
awly: `-parallel` affects the number of tests *within* a package, if they call `t.Parallel()` (not a thing with gocheck).
`-p` affects the number of *different* packages tested concurrently, which is the source of these race conditions.

You can use `go test -p=1` as a bandaid for CI while a proper fix is in progress.
Note that it also prevents parallel compilation, so the whole CI pipeline will be slower with this.

marshall-lee: @webvictim do you you need a review from anyone else or I can merge this branch?

Makefile
webvictim: Got it. Removed the `go get` commands.

utils/utils.go
webvictim: @marshall-lee It's all good now, I'll merge it!

access/slackbot/main.go
```diff
@@ -47,14 +47,14 @@ type requestData struct {
 	roles []string
 }
 
-type slackData struct {
+type SlackData struct {
```
awly, line 50: this probably doesn't need to be exported

utils/nettest/main.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"net"
+	"strconv"
+
+	utils "github.com/gravitational/teleport/lib/utils"
+	"github.com/gravitational/trace"
+)
+
+// GetFreeTCPPortsForTests returns n free ports (which are suggested by the kernel)
+func GetFreeTCPPortsForTests(n int) (utils.PortList, error) {
```
awly, line 14: you could revert this and related changes, up to you

access/slackbot/main.go
webvictim: The linter complained because it wasn't exported, figured it was easier just to export it. I can just revert it and add a comment for the nolint.

utils/nettest/main.go
webvictim: I'll leave it in.

access/slackbot/main.go
marshall-lee: It seems that it doesn't really matter is anything exported or not since it's an executable binary. So eventually, I rewrote everything to be private.
However, in some cases I had to make some types exported to use with `text/template`. I was confused and started to make everything public in new plugins but forgot to fix it in this place.

PR #78, opened 2020-05-13
Fix example code and address linter complaint

webvictim:
Added a quick fix for the example code so that the linter doesn't complain. At the same time, I also fixed another `ineffassign` issue from the linter.


webvictim:
The tests also don't run properly because of port reuse :(

PR #84, opened 2020-05-20
Rename slackbot -> slack and jirabot -> jira

webvictim:
- Rename `slackbot` to `slack` to fit in with other plugins.
- Rename `jirabot` to `jira` to fit in with other plugins.
- Tidy up docs and code by removing trailing whitespace throughout.


benarent:
👍

PR #86, opened 2020-05-21
Add release and promotion pipelines to Drone

webvictim:
Adds release pipelines to Drone which will be triggered when publishing a tag:
```
$ git tag teleport-pagerduty-v0.0.1-alpha.1
$ git push origin teleport-pagerduty-v0.0.1-alpha.1
```
This will run the linter, run tests, build the `teleport-pagerduty` artifact and upload it to the staging S3 bucket.

We can then run this command (which will download the built artifact from the staging S3 bucket and push it to the production S3 bucket (so it can be accessed by our download hosting):

```
$ drone build promote gravitational/teleport-plugins <build number> production
```


webvictim:
Yep, I'll do it once I've validated that they work.

PR #89, opened 2020-05-26
Update gitlab release version to 0.1.0-alpha.1

webvictim:
To match the other plugins and our general versioning scheme.


webvictim:
@benarent Not sure if you want to test this yourself before we cut a release for it?

PR #101, opened 2020-06-02
Add build status badge

webvictim:


awly:
Any way we could send build failures as slack messages?
I suspect few of us regularly pay attention to repo home pages.

webvictim:
Yes, this is possible.

PR #103, opened 2020-06-05
Refactor tests & write user identity to audit log

### LINKED ISSUE #1: All teleport plugins should log the identity of users approving / declining requests
The slackbot should log the identity of users when they approve or deny requests.

Once https://github.com/gravitational/teleport/issues/3204 has been resolved, user identity should be supplied to the Auth Server when setting request state.

marshall-lee:
Tests improvements:

 - Introduce a 0.5sec context for each test run.
 - Introduce access/integration helpers package similar to Teleport's one but access/integration starts ONLY the Auth Server and has misc helper methods dealing with access requests.
 - Overall reduce of boilerplate in tests.
 - Bind Teleport Auth Server on port 0 to run CI tests in parallel (!!!)
 - Move httptest fake servers to separate files.
 - Provide additional tests for plugin data.

Audit log:

 - Fill `Delegator` field of `SetAccessRequestState` to the value `<plugin-name>:<user-identity>` where `<user-identity>` is usually the email of the user who performed some action with the access request. Closes #1.


webvictim:
I think that we should wait to merge this PR until Teleport 4.2.11 or 4.3 is released with the needed commits for delegator entries and :0 port binding. We can then revendor this PR based on that tag rather than using master and merge it in.

marshall-lee:
@webvictim yes this branch cannot be merged right now cause I had to temporarily disable minimum require server version in `access.go`. it must be changed back when teleport re-vendored.

benarent:
Once @webvictim pushed out https://github.com/gravitational/teleport-plugins/issues/110,

Let's add this into a 0.2.0 release, so we can tell users on version before 4.3 to use 0.1.0

marshall-lee:
@benarent @webvictim so for now this PR is blocked to be merged until 0.1.0 is released?

webvictim:
@marshall-lee `0.1.0` has now been released.

benarent:
Thanks @marshall-lee , if we cut a 0.2.0, am I correct that this will write to 4.3 audit log.

marshall-lee:
@benarent yes, audit log should work and also minimum required teleport version bumped to 4.3.0. plugins 0.2.0 won't work with older versions.

PR #105, opened 2020-06-12
Conteinerized manual QA plan

xnutsive:
## Containerized QA flow

This PR adds a `./docker/` directory that does simplifies starting the plugins locally, without requiring any external VMs to QA them. 

The PR doesn't add any new features to the plugins themselves, and doesn't add any new plugins. The PR also doesn't change the way any of the plugins work. It only contains some tooling that we can use internally to quickly QA the plugins, using the [test plan](/testplan.md) or not.

### What to review

I've based the whole thing on Teleport's Docker flow, but some of the things it does (like providing stub licenses) are not implemented and not used in teleport-plugins flow, so there might be some leftovers, especially in `Makefile` and `docker-compose.yaml`. 

Here's what to review thoroughly: 
1. Is the Readme readable? Does it give a clear picture of when one might want to use this, and how to get started? If not, what part can be clarified?
2. I'm new to writing `Makefile`s, might be some weird stuff in there. 
3. Manual testing: I'll try to add a screencast on how to get started step by step, but maybe @benarent will want to try running this, maybe specifically for Pagerduty plugin to debug the webhooks problem? That would help make both readme and the flow itself easier.  

###  TODOs

That's my WIP todo list, disregard that :)

- [x] Run Teleport, tctl, and Teleport Web UI with root access to them.
- [x] Test that we're able to create new roles and users
- [x] Test that we're able to sign and export certificates
- [x] Run Slack Plugin and provide a way to edit it's configuration
- [x] Run Mattermost Plugin and provide a way to edit it's configuration
- [x] Run Pagerduty Plugin and provide a way to edit it's configuration
- [x] Run Gitlab Plugin and provide a way to edit it's configuration
- [x] Run Jira Plugin and provide a way to edit it's configuration

- [x] Provide a runner for both regular and `--insecure-no-tctl` modes
- [x] Optional: Document how to setup public hosts for plugins with ngrok / provide a make command for this as well.
- [ ] Optional: Document how to setup your own certificates with certmanager / LE

- [ ] Optional: Provide a way to run Jira Server as one of the Dockerized services, and run two different versions of Jira plugin (Server & Cloud).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,9 @@ release/access-gitlab:
 .PHONY: releases
 releases: release/access-slack release/access-jira release/access-mattermost release/access-pagerduty release/access-gitlab
 
+.PHONY: build-all
+build-all: access-slack access-jira access-mattermost access-pagerduty access-gitlab
+
 #
 # Lint the Go code.
 # By default lint scans the entire repo. Pass FLAGS='--new' to only scan local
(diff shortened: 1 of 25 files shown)


xnutsive:
Still work in progress, I'll make sure, role, certificate management works as expected, add 4 more integrations, and ask for your feedback <3

xnutsive:
Progress report: 

- The system now uses enterprise version of Teleport, but expects the license from the user to be placed in `data/var/lib/teleport/`. I can use the license from the fixtures in `e`, but then I'll shoot myself in the leg, since I don't have access to it. This can be tweaked later though. 
- Teleport version can be set in `docker/Makefile` or overridden in `docker/teleport/Dockerfile`. Any version available on the website will work. 
- `make config`expects that the network and all the docker requirements are already met, starts the containers, creates all required users and roles for testing, exports plugin certificates to `/mnt/shared/certs/` and exits. 
- The plugin can't yet start: complaints at the self-generated certificate not being there. Investigating. 

I haven't updated the README yet to account for all the changes, too. But this can already be reviewed — your advice on best practices would be useful ;-)

xnutsive:
#111 messes up the flow a bit — we'll have to create the directories for plugin certs manually, but that shouldn't be a problem.

xnutsive:
@benarent, I've made the flow work, but since it builds up on teleport's own containerised flow, and since that builds on builbox, the documentation for this might be messy. Please review if the docker/readme.md makes sense at all. 

If it does, try launching this and let me know if it works for you or if you hit any errors, even if you're able to quickly sort them out. 

This is more or less OK to merge, next steps would be: 
- Add 3 missing plugin configs and sevices
- [Maybe] add configs for all 5 services with shared app tokens for test purposes (!?) and use something like a vault to commit them, but not share them out in the open. This would make it possible to write an end to end test that would start up Teleport + a plugin with a config that would actually work. 

After that, I would consider adding some of the services that offer docker images as test instances to the flow. Speficically, we can run mattermost in a docker container, and same for Jira Server. It takes some setup, but again, would make testing possible end to end.

xnutsive:
Updated the PR description with more details on how to test and review this ;)

benarent:
Thanks @xnutsive the Readme is a bit of a beast but I'll try it out tomorrow.

xnutsive:
I've tested this on the Slack plugin tonight. Caught and fixed a couple of bugs. I'll do a screencast, and we can promote this as an easy way to sandbox / test plugins for customers if you feel like it then. 

If you don't mind, I'll merge this so I have a better way to test at least two other PRs: 
1. Your proposed default config changes (ports in public_addr) and 
2. My PR with Slack plugin's notification_only mode.

PR #112, opened 2020-07-28
Pagerduty: implement auto-approval

### LINKED ISSUE #72: Pagerduty discussion: should we do the auto-approve flow?
As of right now, Pagerduty integration allows to manually approve or deny Teleport permission requests. 

@Ben proposed implementing an auto-approve flow for situations when an on-call engineer requests the permission themselves. This behavior would be opt-in on a config level.

**Here's how the flow should work**

- A user runs `tsh login --request-roles %roles%`
- Teleport broadcasts the request via the GRPC API. 
- `teleport-pagerduty` plugin uses the User ID provided in `access.Request.User`
- `teleport-pagerduty` uses [Schedules API](https://api-reference.pagerduty.com/#!/Schedules/get_schedules) to determine the list of Pagerduty Users who are on call. 
- `teleport-pagerduty` uses `Request.User` to match against Pagerduty users. If the request user matches the user on call, and if they're the only on call user, then `teleport-pagerduty` immediately approves the request, automatically. 

**We need several decisions before we can implement this flow:** 

1. Should the request be automatically approved only if the requester is _the only_ user who's on call, or if the requester is _one of the users on call_? I.e. if there's someone else on call who can review their request, should we still auto-approve? Should this be optional?
2. On my local VM, I'm getting my Linux username as `Request.User` for my permission requests. To match Teleport cluster users to Pagerduty users, we'll either need to match their emails, or other metadata that can be stored in both systems. Pagerduty alllows “contact methods” that we can use, but that approach feels.. unreliable. 
3. When auto-approve occurs, should the plugin still create an incident and auto-resolve it to live a log / trace that a certain permission request was created and auto-approved, or leave it completely out of Pagerduty? 

**Engineering considerations** 

Whatever way we choose to match teleport user to pagerduty users, Pagerduty's Schedule API will only return a bunch of _user IDs_. So unless we store them in Teleport and provide them in Request.PagerdutyUserID, the plugin will need to iterate over all of the schedules that are current, and grab all users one by one, and then match their information to teleport's user info. That's O(n) API requests just to search for a matching user record, N is the count of users on call. Might as well do /users/all and cache the results in the plugin at start time and update it from time to time. 

**Next steps** 
Let's first discuss the desired behavior and decide on urgency / priority of this behavior. I'll update the issue with the spec once we have that. We'll then schedule the implementation ;)

marshall-lee:
Closes #72

### DIFF
--- a/access/integration/helpers.go
+++ b/access/integration/helpers.go
@@ -495,9 +495,7 @@ func (i *TeleInstance) GetAccessRequest(ctx context.Context, reqID string) (serv
 	return requests[0], nil
 }
 
-func (i *TeleInstance) PollAccessRequestPluginData(ctx context.Context, plugin, reqID string, timeout time.Duration) (map[string]string, error) {
-	ctx, cancel := context.WithTimeout(ctx, timeout)
-	defer cancel()
+func (i *TeleInstance) PollAccessRequestPluginData(ctx context.Context, plugin, reqID string) (map[string]string, error) {
 	auth := i.Process.GetAuthServer()
 	filter := services.PluginDataFilter{
 		Kind:     services.KindAccessRequest,
--- a/access/pagerduty/app.go
+++ b/access/pagerduty/app.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"net/url"
+	"regexp"
 	"strings"
 	"time"
 
@@ -32,6 +33,8 @@ func NewApp(conf Config) (*App, error) {
 	return app, nil
 }
 
+var emailRegex = regexp.MustCompile("^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$")
+
 // Run initializes and runs a watcher and a callback server
 func (a *App) Run(ctx context.Context) error {
 	// Initialize the process.
@@ -53,12 +56,13 @@ func (a *App) PublicURL() *url.URL {
 	return a.webhookSrv.BaseURL()
 }
 
-func (a *App) run(ctx context.Context) (err error) {
+func (a *App) run(ctx context.Context) error {
+	var err error
 	log.Infof("Starting Teleport Access PagerDuty extension %s:%s", Version, Gitref)
 
 	a.webhookSrv, err = NewWebhookServer(a.conf.HTTP, a.onPagerdutyAction)
 	if err != nil {
-		return
+		return trace.Wrap(err)
 	}
 
 	a.bot = NewBot(a.conf.Pagerduty, a.webhookSrv)
@@ -71,7 +75,7 @@ func (a *App) run(ctx context.Context) (err error) {
 	if trace.Unwrap(err) == access.ErrInvalidCertificate {
 		log.WithError(err).Warning("Auth client TLS configuration error")
 	} else if err != nil {
-		return
+		return trace.Wrap(err)
 	}
 
 	a.accessClient, err = access.NewClient(
@@ -81,34 +85,34 @@ func (a *App) run(ctx context.Context) (err error) {
 		tlsConf,
 	)
 	if err != nil {
-		return
+		return trace.Wrap(err)
 	}
 	if err = a.checkTeleportVersion(ctx); err != nil {
-		return
+		return trace.Wrap(err)
 	}
 
 	log.Debug("Starting PagerDuty API health check...")
 	if err = a.bot.HealthCheck(ctx); err != nil {
 		log.WithError(err).Error("PagerDuty API health check failed")
-		return
+		return trace.Wrap(err)
 	}
 	log.Debug("PagerDuty API health check finished ok")
 
 	err = a.webhookSrv.EnsureCert()
 	if err != nil {
-		return
+		return trace.Wrap(err)
 	}
 	httpJob := a.webhookSrv.ServiceJob()
 	a.SpawnCriticalJob(httpJob)
 	httpOk, err := httpJob.WaitReady(ctx)
 	if err != nil {
-		return
+		return trace.Wrap(err)
 	}
 
 	log.Debug("Setting up the webhook extensions")
 	if err = a.bot.Setup(ctx); err != nil {
 		log.WithError(err).Error("Failed to set up webhook extensions")
-		return
+		return trace.Wrap(err)
 	}
 	log.Debug("PagerDuty webhook extensions setup finished ok")
 
@@ -120,7 +124,7 @@ func (a *App) run(ctx context.Context) (err error) {
 	a.SpawnCriticalJob(watcherJob)
 	watcherOk, err := watcherJob.WaitReady(ctx)
 	if err != nil {
-		return
+		return trace.Wrap(err)
 	}
 
 	a.mainJob.SetReady(httpOk && watcherOk)
@@ -178,10 +182,7 @@ func (a *App) onWatcherEvent(ctx context.Context, event access.Event) error {
 }
 
 func (a *App) onPagerdutyAction(ctx context.Context, action WebhookAction) error {
-	log := log.WithFields(logFields{
-		"pd_http_id": action.HTTPRequestID,
-		"pd_msg_id":  action.MessageID,
-	})
+	log := utils.GetLogger(ctx)
 
 	keyParts := strings.Split(action.IncidentKey, "/")
 	if len(keyParts) != 2 || keyParts[0] != pdIncidentKeyPrefix {
@@ -208,7 +209,7 @@ func (a *App) onPagerdutyAction(ctx context.Context, action WebhookAction) error
 		return trace.Wrap(err)
 	}
 
-	log = log.WithField("pd_incident_id", action.IncidentID)
+	ctx, log = utils.WithLogField(ctx, "pd_incident_id", action.IncidentID)
 
 	if pluginData.PagerdutyData.ID != action.IncidentID {
 		log.WithField("plugin_data_incident_id", pluginData.PagerdutyData.ID).Debug("plugin_data.incident_id does not match incident.id")
@@ -225,41 +226,25 @@ func (a *App) onPagerdutyAction(ctx context.Context, action WebhookAction) error
 			userName = agent.Name
 		}
 	}
-	log = log.WithFields(logFields{
+
+	ctx, _ = utils.WithLogFields(ctx, logFields{
 		"pd_user_email": userEmail,
 		"pd_user_name":  userName,
 	})
 
-	var (
-		reqState   access.State
-		resolution string
-	)
-
 	switch action.Name {
 	case pdApproveAction:
-		reqState = access.StateApproved
-		resolution = "approved"
+		return a.setRequestState(ctx, req.ID, action.IncidentID, userEmail, access.StateApproved)
 	case pdDenyAction:
-		reqState = access.StateDenied
-		resolution = "denied"
+		return a.setRequestState(ctx, req.ID, action.IncidentID, userEmail, access.StateDenied)
 	default:
 		return trace.BadParameter("unknown action: %q", action.Name)
 	}
-
-	if err := a.accessClient.SetRequestState(ctx, req.ID, reqState, userEmail); err != nil {
-		return trace.Wrap(err)
-	}
-	log.Infof("PagerDuty user %s the request", resolution)
-
-	if err := a.bot.ResolveIncident(ctx, reqID, pluginData.PagerdutyData, resolution); err != nil {
-		return trace.Wrap(err)
-	}
-	log.Infof("Incident %q has been resolved", action.IncidentID)
-
-	return nil
 }
 
 func (a *App) onPendingRequest(ctx context.Context, req access.Request) error {
+	ctx, _ = utils.WithLogField(ctx, "request_id", req.ID)
+
 	reqData := RequestData{User: req.User, Roles: req.Roles, Created: req.Created}
 
 	pdData, err := a.bot.CreateIncident(ctx, req.ID, reqData)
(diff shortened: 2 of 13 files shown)


access/pagerduty/bot.go
```diff
@@ -209,46 +273,162 @@ func (b *Bot) CreateIncident(ctx context.Context, reqID string, reqData RequestD
...
+		Post(b.APIURL("incidents", incidentID, "notes"))
 	if err != nil {
 		return trace.Wrap(err)
 	}
-	_, err = client.ManageIncidents(b.from, []pd.ManageIncidentsOptions{
+	if resp.IsError() {
+		return responseError(resp)
+	}
+
+	// TODO: PUT /incidents/{id}
```
awly, line 317: Is this TODO still relevant?

access/pagerduty/bot.go
```diff
@@ -209,46 +273,162 @@ func (b *Bot) CreateIncident(ctx context.Context, reqID string, reqData RequestD
...
+	return &userRes.User, nil
+}
+
+func (b *Bot) GetUserByEmail(ctx context.Context, userEmail string) (*pd.User, error) {
+	usersOpts := pd.ListUsersOptions{
+		Query: userEmail,
+		APIListObject: pd.APIListObject{
+			Limit: pdListLimit,
+		},
+		// Total: true, // TODO:
```
awly, line 367: Is this TODO relevant?

access/pagerduty/bot.go
```diff
@@ -209,46 +273,162 @@ func (b *Bot) CreateIncident(ctx context.Context, reqID string, reqData RequestD
...
+		SetResult(&onCallRes).
+		Get(b.APIURL("oncalls"))
+	if err != nil {
+		return false, trace.Wrap(err)
+	}
+	if resp.IsError() {
+		return false, responseError(resp)
+	}
+
+	return len(onCallRes.OnCalls) > 0, nil
```
awly, line 431: also confirm that the oncalls list contains a user with the right email
it's pagerduty API's job, just still good to double-check in case, e.g. `userID` was empty

access/pagerduty/bot.go
```diff
@@ -267,3 +447,9 @@ func (b *Bot) buildIncidentBody(reqID string, reqData RequestData) (string, erro
 	}
 	return builder.String(), nil
 }
+
+func responseError(resp *resty.Response) error {
+	result := resp.Error().(*ErrorResult)
```
awly, line 452: use a checked type assertion here to avoid a panic

access/pagerduty/bot.go
```diff
@@ -267,3 +447,9 @@ func (b *Bot) buildIncidentBody(reqID string, reqData RequestData) (string, erro
...
+	err := fmt.Sprintf("http error code=%v, err_code=%v, message=%v, errors=[%v]", resp.StatusCode(), result.Code, result.Message, strings.Join(result.Errors, ", "))
+	return trace.Errorf(err)
```
awly, line 454: `return trace.Errorf("http error code=%v, err_code=%v, message=%v, errors=[%v]", resp.StatusCode(), result.Code, result.Message, strings.Join(result.Errors, ", "))`

access/pagerduty/bot.go
```diff
@@ -41,80 +45,109 @@ func init() {
...
+	if err != nil {
+		return trace.Wrap(err, "failed to fetch PagerDuty service info: %v", err)
+	}
+	if resp.IsError() {
+		if contentType := resp.Header().Get("Content-Type"); contentType != "application/json" {
+			return trace.Errorf("wrong content_type=%q", contentType)
+		}
+		if code := resp.StatusCode(); code == http.StatusUnauthorized {
+			return trace.Errorf("got %v from API endpoint, perhaps PagerDuty credentials are not configured well", code)
+		}
```
awly, line 117: All of this should live in `func responseError`

access/pagerduty/app.go
```diff
@@ -288,12 +280,64 @@ func (a *App) onDeletedRequest(ctx context.Context, req access.Request) error {
...
+func (a *App) setRequestState(ctx context.Context, reqID, incidentID, userEmail string, state access.State) error {
+	log := utils.GetLogger(ctx)
+	var resolution string
+
+	switch state {
+	case access.StateApproved:
+		resolution = "approved"
+	case access.StateDenied:
+		resolution = "denied"
+	}
```
fspmarshall, line 301: Best have a `default` case here.  Either return an `error` indicating an unexpected input, or fallback to `state.String()`.

access/integration/helpers.go
```diff
@@ -520,6 +520,19 @@ func (i *TeleInstance) PollAccessRequestPluginData(ctx context.Context, plugin,
...
+func (i *TeleInstance) PollAccessRequestStateChange(ctx context.Context, reqID string, expectedState services.RequestState) error {
+	for {
+		req, err := i.GetAccessRequest(ctx, reqID)
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		if req.GetState() == expectedState {
+			return nil
+		}
+		time.Sleep(25 * time.Millisecond)
```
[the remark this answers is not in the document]
marshall-lee: @awly I think I'll replace it with `runtime.Gosched()`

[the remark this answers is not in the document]
marshall-lee: I think it's worth replacing with event watcher. This was a quick implementation just to make sure that my code works.

access/pagerduty/pagerduty_test.go
```diff
@@ -303,6 +314,31 @@ func (s *PagerdutySuite) TestDenial(c *C) {
 	c.Assert(request.GetState(), Equals, services.RequestState_DENIED)
 }
 
+func (s *PagerdutySuite) TestAutoApprovalWhenOnCall(c *C) {
+	s.appConfig.Pagerduty.AutoApprove = true
+	s.userName = s.me.Username + "@example.com"
+	s.fakePagerduty.AddOnCall(pd.OnCall{})
```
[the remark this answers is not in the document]
marshall-lee: @awly i thought it's an overkill to implement a real OnCall filtering by multiple parameters in `FakePagerduty`. And this approach works well since in my code I just check `len(onCalls) > 0` and trust that Pagerduty API works correctly doesn't return bad data such an empty `pd.OnCall{}`.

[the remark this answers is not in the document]
marshall-lee: If we rely on third-party API correctness all the time why we shouldn't trust it in this case?

Anyway, I'll write that double check of onCall data so I have to fix this one too.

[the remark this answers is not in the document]
awly: This is a sensitive piece of code, since it enables privilege escalation.
I'd prefer to be more paranoid about all the 3rd party responses.

access/pagerduty/bot.go
```diff
@@ -41,80 +45,109 @@ func init() {
...
+	resp, err := b.NewRequest(ctx).
+		Get(b.APIURL("services", b.serviceID))
+	if err != nil {
+		return trace.Wrap(err, "failed to fetch PagerDuty service info: %v", err)
+	}
+	if resp.IsError() {
+		if contentType := resp.Header().Get("Content-Type"); contentType != "application/json" {
+			return trace.Errorf("wrong content_type=%q", contentType)
+		}
+		if code := resp.StatusCode(); code == http.StatusUnauthorized {
```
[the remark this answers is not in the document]
marshall-lee: It's a method to check if an api endpoint is set up correctly in the config. So here I need to check first does it seem like a json endpoint at all. If it looks like an "API" then lets check the code next.

access/pagerduty/bot.go
```diff
@@ -4,20 +4,24 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 	"text/template"
 	"time"
 
 	pd "github.com/PagerDuty/go-pagerduty"
+	"github.com/go-resty/resty"
```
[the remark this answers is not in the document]
marshall-lee: It was decided to gradually transition all the plugins to use `go-resty` for the unification. I tried `go-resty` and I found its API very convenient and simple yet very flexible for anything I need. We have 5 plugins and all of them used different client libraries for each of the service. All that libraries have different API and different conventions and some of them lack the things I need. For example, I didn't find how to pass a `ctx` in `go-pagerduty` so I had to do some hacks for this. Also some of the libraries contain mistakes and even lack some of the methods. Yes, everything is fixable and I could just send a PR but problems like `ctx` passing are architectural and not always easy.

I don't think that we'll miss something crucial transitioning from specialized libraries to `go-resty`. All that libraries actually do nothing special — they just wrap `http.Client` and do JSON encoding/decoding. It's a common task that `go-resty` handles well. The only thing about such libraries which is useful is type definitions. Well, I could use them as a reference and simply "copy" some of them. I find this tradeoff acceptable cause typically I don't need very much of these type defs.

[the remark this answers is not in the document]
awly: Sounds reasonable.
It just made this PR review much longer, the auto-approval changes are drowned out by refactoring.
For future PRs, please do large refactoring separate from functional changes (either separate PRs or at least separate commits).

access/pagerduty/bot.go
```diff
@@ -39,90 +40,112 @@ func init() {
 	}
 }
 
-// Bot is a wrapper around pd.Client that works with access.Request
+// Bot is a wrapper around resty.Client.
```
awly, line 43: nit: this is an implementation detail.
prefer doc comments like "Bot is a client for PagerDuty API"

access/pagerduty/bot.go
```diff
@@ -39,90 +40,112 @@ func init() {
...
+		req.SetError(&ErrorResult{})
+		return nil
+	})
+	client.OnAfterResponse(func(_ *resty.Client, resp *resty.Response) error {
+		if resp.IsError() {
+			result := resp.Error()
+			if result, ok := result.(*ErrorResult); ok {
+				return trace.Errorf("http error code=%v, err_code=%v, message=%v, errors=[%v]", resp.StatusCode(), result.Code, result.Message, strings.Join(result.Errors, ", "))
+			}
+			return trace.Errorf("Unknown error result %#v", result)
```
awly, line 80: nit: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

access/pagerduty/bot.go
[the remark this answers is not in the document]
awly: Fair enough.
If you some proxy that gives you a non-200 response code with non-json body, complaining about `Content-Type` first might make it harder to debug.
But no big deal.

access/pagerduty/bot.go
```diff
@@ -39,90 +40,112 @@ func init() {
...
-	client := b.NewClient(ctx)
-
-	if _, err := client.GetService(b.serviceID, nil); err != nil {
-		return trace.Wrap(err, "failed to fetch pagerduty service info: %v", err)
+	var result ServiceResult
+	resp, err := b.client.NewRequest().
+		SetContext(ctx).
+		SetResult(&result).
+		Get(utils.BuildURLPath("services", b.serviceID))
+	if resp != nil && resp.IsError() {
```
awly, line 98: Check this after checking `err`

access/pagerduty/bot.go
```diff
@@ -39,90 +40,112 @@ func init() {
...
+		if contentType := resp.Header().Get("Content-Type"); contentType != "application/json" {
+			return trace.Errorf("wrong Content-Type in PagerDuty response: %q", contentType)
+		}
+		if code := resp.StatusCode(); code == http.StatusUnauthorized {
```
awly, line 102: What about other non-403 error codes?

access/pagerduty/pagerduty_test.go
```diff
@@ -169,24 +192,31 @@ func (s *PagerdutySuite) shutdownApp(c *C) {
...
+	rawData, err := s.teleport.PollAccessRequestPluginData(s.ctx, "pagerduty", reqID)
 	c.Assert(err, IsNil)
 	return DecodePluginData(rawData)
 }
 
-func (s *PagerdutySuite) postAction(c *C, incident pd.Incident, action string) {
+func (s *PagerdutySuite) postActionAndCheck(c *C, incident Incident, action string) {
+	response, err := s.postAction(incident, action)
+	c.Assert(response.Body.Close(), IsNil)
+	c.Assert(err, IsNil)
```
awly, line 215: check `err` first, otherwise `response` could be `nil`

access/pagerduty/fake_pagerduty_test.go
```diff
@@ -8,60 +8,74 @@ import (
...
-	newExtensions    chan pd.Extension
-	incidents        sync.Map
-	newIncidents     chan pd.Incident
-	incidentUpdates  chan pd.Incident
-	newIncidentNotes chan pd.IncidentNote
+	srv *httptest.Server
+
+	objects sync.Map
+	// Extensions
+	extensionIDCounter uint64
```
awly, line 24: are these counters used anywhere?

marshall-lee: Yes, for instance look at `StoreExtension`.

access/pagerduty/bot.go
marshall-lee: There was a function `func responseError(resp *resty.Response) error` that I used to convert an error code to Go `error`.
Now I moved it to `OnAfterResponse` callback (see `NewBot` function). So for every code > 399 the `err` will always be non-nil.

access/pagerduty/bot.go
marshall-lee: I added a status code to `error` conversion in `OnAfterResponse` callback. So if check `err` first, this `resp.IsError()` will be unreachable.

marshall-lee: ping @awly

awly: OK, please leave a comment here explaining this.

access/pagerduty/bot.go
awly: Please leave a comment explaining this, the next person to read this code might have the same question

PR #117, opened 2020-08-09
Implements Version subcommand:

### LINKED ISSUE #113: Teleport Plugins should support 'version'
```bash
~/plugins/teleport-access-pagerduty: $ teleport version
Teleport Enterprise v4.3.0-alpha.2git:v4.3.0-alpha.2-0-gd9e436bdc go1.13.2

~/plugins/teleport-access-pagerduty: $ teleport-pagerduty version
ERRO   Terminating due to error error:expected command but got "version" utils/bail.go:12
```

also if a commend isn't found it should be more graceful.

e.g. 

```
$ teleport x
error: expected command but got "x"
# vs
$ teleport-pagerduty x
ERRO   Terminating due to error error:expected command but got "x" utils/bail.go:12
``

### PR DESCRIPTION
**Pull request by xnutsive**
- Closes #113
- Works for all current plugins: Slack, Jira, Mattermost,
  Pagerduty, and Gitlab.
- Print func in utils to dry things up a bit.

Disclaimer: I was just curious about go and wrote a few lines, please go extra thorough on the review. 

Next steps: I noticed that most of the `main.go` in all of the plugin is the same, save for the plugin name. I think it would be nice to have a `runner` utility struct that would have a func to declare a standard Kingpin app with three shared commands (start, configure, version), and accept any extra commands / expose the app in case a new integration needs more flexibility. 

I'll play around with this next if you don't mind ;-)

### DIFF
--- a/utils/runner.go
+++ b/utils/runner.go
@@ -1,0 +1,15 @@
+package utils
+
+import (
+	"fmt"
+	"runtime"
+)
+
+// PrintVersion prints the specified app version to STDOUT
+func PrintVersion(appName string, version string, gitref string) {
+	if gitref != "" {
+		fmt.Printf("%v v%v git:%v %v\n", appName, version, gitref, runtime.Version())
+	} else {
+		fmt.Printf("%v v%v %v\n", appName, version, runtime.Version())
+	}
+}
--- a/access/gitlab/main.go
+++ b/access/gitlab/main.go
@@ -23,6 +23,7 @@ func main() {
 	app := kingpin.New("teleport-gitlab", "Teleport plugin for access requests approval via GitLab.")
 
 	app.Command("configure", "Prints an example .TOML configuration file.")
+	app.Command("version", "Prints teleport-gitlab version and exits.")
 
 	startCmd := app.Command("start", "Starts a Teleport GitLab plugin.")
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: utils/runner.go [hunk 1]
```diff
@@ -0,0 +1,14 @@
...
+	"fmt"
+)
+
+// Prints the specified app version to STDOUT
+func PrintVersion(appName string, version string, gitref string) {
+	if gitref != "" {
+		fmt.Printf("%v version %v, gitref: %v\n", appName, version, gitref)
+	} else {
+		fmt.Printf("%v version %v\n", appName, version)
+	}
```
webvictim, line 13: Can we make this match the `teleport version`/`tctl version`/`tsh version` output format?

Something more like:

```go
// PrintVersion prints the specified app version to STDOUT
func PrintVersion(appName string, version string, gitref string) {
	if gitref != "" {
		fmt.Printf("%v v%v git:%v %v\n", appName, version, gitref, runtime.Version())
	} else {
		fmt.Printf("%v v%v %v\n", appName, version, runtime.Version())
	}
}
```

xnutsive:
Sure thing, I'll tweak that and request the re-review when done.

Nate
(650) 933-7237

xnutsive:
That should do it ;)

PR #118, opened 2020-08-09
Prettier all the READMEs

xnutsive:
Just cleaned up the readme files, and included a .prettierc to enforce 80 characters per line and proper word wrapping. @benarent, please take another look in case I messed them up. There should be no visible changes to README files when you view them, but the diffs should look like they're just reformatted. If there's a difference in how they look when they're rendered, that might be a problem ;)

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,2 @@
+printWidth: 80
+proseWrap: "always"
--- a/README.md
+++ b/README.md
@@ -7,44 +7,48 @@ Set of plugins for Teleport
 ## Access API
 
 The [access](./access) package exposes a simple API for managing access requests
-which can be used for writing plugins.  More info can be found in
-[access/README.md](./access/README.md), including instructions on how to properly
-provision necessary certificates.
+which can be used for writing plugins. More info can be found in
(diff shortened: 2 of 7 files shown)


benarent:
This looks good, but for the setup instructions we should replace the readmes and link to the Gravitational docs for the plugins so we've one canonical source of docs. 

https://gravitational.com/teleport/docs/enterprise/workflow/ssh_approval_slack/ 
https://gravitational.com/teleport/docs/enterprise/workflow/ssh_approval_mattermost/
https://gravitational.com/teleport/docs/enterprise/workflow/ssh_approval_jira_cloud/
https://gravitational.com/teleport/docs/enterprise/workflow/ssh_approval_jira_server/
https://gravitational.com/teleport/docs/enterprise/workflow/ssh_approval_pagerduty/

xnutsive:
Would it be better to preserve the readmes, but add the setup instructions with explicit comment in the beginning of the readmes?

Nate
(650) 933-7237

xnutsive:
@benarent, take a look — I've added the links to both the root README and each plugin's readme file.

PR #124, opened 2020-08-14
Ensure certs directory exists before trying to write self-signed certs

### LINKED ISSUE #111: Plugins fail to save a self-generated TLS certificate if their workdir doesn't exist
If the slack plugin, for example, works with `/var/lib/teleport/plugins/slack`, but for some reason the directory is not there, the plugin won't start without valid certs, or `--insecure-no-tls`. 

Here's the start log: 

```
teleport_slack_plugin | WARN   No TLS Keys provided, using self signed certificate. logrus/exported.go:137
teleport_slack_plugin | WARN   Generating self signed key and cert to /var/lib/teleport/plugins/slack/server.key /var/lib/teleport/plugins/slack/server.crt. utils/http.go:326
teleport_slack_plugin | ERRO   Terminating due to error error:error writing key PEM utils/bail.go:12
teleport_slack_plugin | DEBU   "\nERROR REPORT:\nOriginal Error: *os.PathError open /var/lib/teleport/plugins/slack/server.key: no such file or directory\nStack Trace:\n\t/gopath/src/github.com/gravitational/teleport-plugins/utils/http.go:339 github.com/gravitational/teleport-plugins/utils.(*HTTP).EnsureCert\n\t/gopath/src/github.com/gravitational/teleport-plugins/access/slack/callback_server.go:59 main.(*CallbackServer).EnsureCert\n\t/gopath/src/github.com/gravitational/teleport-plugins/access/slack/app.go:106 main.(*App).run\n\t/gopath/src/github.com/gravitational/teleport-plugins/utils/process.go:208 github.com/gravitational/teleport-plugins/utils.NewServiceJob.func1\n\t/gopath/src/github.com/gravitational/teleport-plugins/utils/process.go:229 github.com/gravitational/teleport-plugins/utils.(*serviceJob).DoJob\n\t/gopath/src/github.com/gravitational/teleport-plugins/utils/process.go:82 github.com/gravitational/teleport-plugins/utils.NewProcess.func2.1\n\t/opt/go/src/runtime/asm_amd64.s:1358 runtime.goexit\nUser Message: error writing key PEM\n" utils/bail.go:12
```

An easy fix is to create the dir `/var/lib/teleport/plugins/slack` in my case. But in general, I think it would be a better practice to create those dirs if they're missing automatically. I agree the common practice is to provide certs, but hey.

xnutsive:
If the default certs directory doesn't exist yet, try creating it with 0644 permissions. Closes #111.

Cleaned up `utils/http.go` a little bit, too — trying to make go-lint happier, one file at a time. My goal with this is to heal my minor annoyance and learn. How does it look, what can I improve in the code I committed?

### DIFF
--- a/utils/http.go
+++ b/utils/http.go
@@ -123,6 +123,7 @@ func (conf *HTTPConfig) Check() error {
 	return nil
 }
 
+// ServeHTTP processes one http request.
 func (auth *HTTPBasicAuth) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 	username, password, ok := r.BasicAuth()
 
@@ -316,7 +317,7 @@ func (h *HTTP) NewURL(subpath string, values url.Values) *url.URL {
 // but --insecure-no-tls is not invoked, it'll automatically
 // generaate a self-signed ceritificate and store it in the
 // provided default certificate path.
-func (h *HTTP) EnsureCert(defaultCertPath string) (err error) {
+func (h *HTTP) EnsureCert(defaultPath string) (err error) {
 	if h.Insecure {
 		return nil
 	}
@@ -326,32 +327,33 @@ func (h *HTTP) EnsureCert(defaultCertPath string) (err error) {
 		return err
 	}
 
-	return h.GenerateSelfSignedCert(defaultCertPath)
+	return h.GenerateSelfSignedCert(defaultPath)
 }
 
 // GenerateSelfSignedCert crates a self-signed certificate pair
 // and save it to the provided path.
 // It uses Teleport's utils.GenerateSelfSignedCert under the hood.
-func (h *HTTP) GenerateSelfSignedCert(defaultCertPath string) (err error) {
+func (h *HTTP) GenerateSelfSignedCert(defaultPath string) (err error) {
 	log.Warningf("No TLS Keys provided, using self signed certificate.")
 
 	// If files are not specified, try to fall back on self signed certificate.
-	h.CertFile = defaultCertPath + ".crt"
-	h.KeyFile = defaultCertPath + ".key"
+	h.CertFile = defaultPath + ".crt"
+	h.KeyFile = defaultPath + ".key"
 	_, err = tls.LoadX509KeyPair(h.CertFile, h.KeyFile)
 	if err == nil {
-		// self-signed or another cert is already y ,the path,d th
+		// self-signed or another cert is already in the defaault self-signed
+		// cert path, safe to quit.
 		return nil
 	}
 	if !os.IsNotExist(err) {
-		return trace.Wrap(err, "unrecognized error reading certs")
+		return trace.Wrap(err, "unrecognized error reading self-signed certs")
 	}
 
 	log.Warningf("Generating self signed key and cert to %v %v.", h.KeyFile, h.CertFile)
 
-	certDir := path.Dir(defaultCertPath)
+	certDir := path.Dir(defaultPath)
 	if _, err = os.Stat(certDir); os.IsNotExist(err) {
-		log.Warningf("Self-signed TLS certs directory %v doesn't exist, creating it now", certDir)
+		log.Debugf("Self-signed TLS certs directory %v doesn't exist, creating it now", certDir)
 
 		err := os.MkdirAll(certDir, 0644)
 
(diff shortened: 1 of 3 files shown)


utils/http.go
```diff
@@ -298,25 +311,36 @@ func (h *HTTP) NewURL(subpath string, values url.Values) *url.URL {
...
 }
 
-// EnsureCert checks cert and key files consistency. It also generates a self-signed cert if it was not specified.
-func (h *HTTP) EnsureCert(defaultPath string) (err error) {
+// EnsureCert checks cert and key files consistency.
+// If the certificate path is not provided in the config,
+// but --insecure-no-tls is not invoked, it'll automatically
+// generaate a self-signed ceritificate and store it in the
+// provided default certificate path.
+func (h *HTTP) EnsureCert(defaultCertPath string) (err error) {
```
marshall-lee, line 319: Why you specified `defaultPath` to `defaultCertPath`? We don't work here with any other types of paths.

utils/http.go
```diff
@@ -325,6 +349,17 @@ func (h *HTTP) EnsureCert(defaultPath string) (err error) {
 
 	log.Warningf("Generating self signed key and cert to %v %v.", h.KeyFile, h.CertFile)
 
+	certDir := path.Dir(defaultCertPath)
+	if _, err = os.Stat(certDir); os.IsNotExist(err) {
```
marshall-lee, line 353: It's rather defensive approach and note that non-existent directory is not the only reason the file write could fail. The file permissions are the reason too, directory could exist but be invisible for current user, entire filesystem could be mount readonly, etc.

I think it's better to just add more detailed error text describing the possible reasons of fail, e.g. change `"error writing key PEM"` to something like `"error writing key PEM, please check that directory exists and has write permissions"`.

utils/http.go
```diff
@@ -325,6 +349,17 @@ func (h *HTTP) EnsureCert(defaultPath string) (err error) {
...
+		log.Warningf("Self-signed TLS certs directory %v doesn't exist, creating it now", certDir)
+
+		err := os.MkdirAll(certDir, 0644)
```
marshall-lee, line 356: Nevertheless, if you think it's worth to have this `mkdir -p`, lets add it. I just wonder who ever needs these self signed certs, cloud service like Slack or Pagerduty refuse to send webhooks to endpoints with self signed certs.

utils/http.go
xnutsive: Just wanted to be extra-explicit about what that path is about. It's also the only argument our function takes, but we're not naming it `a`, because that would be unreadable, right? ;)

Maybe you're right — I don't have a strong preference tbh, I think given that the function is named EnsureCert, it's pretty obvious that an arg named `defaultPath` should contain the cert. 

I'll rename it ;)

utils/http.go
xnutsive: I think you're right that the whole approach to create the certs path if it's not there is defensive. You're also right that using self-signed certs for plugins even in testing makes less sense than using self-signed certs in Teleport itself, since all of 3rd party service won't work with the self-signed certs anyway. 

In #105, I've just added a task that creates the directories explicitly one chooses to. 

You're also right that it would be an improvement to add some explicit error message in case the directory is not writable or does not exist. 

Creating default directories automatically on `--insecure-no-tls` just makes sure that _if_ the user uses that flag — than the certificates will be generated in 99% situations, unless the dirs exist but the permissions are broken. 

Again, I don't have a strong preference, maybe it's easier to not create the directories, but give an explicit error message and make sure the user understands that they should create a directory, and that self-signed certs will be generated and put there, instead of just silently creating them.

utils/http.go
xnutsive: This code basically does `mkdir -p` ;)

utils/http.go
marshall-lee: > Creating default directories automatically on --insecure-no-tls just makes sure that if the user uses that flag — than the certificates will be generated in 99% situations, unless the dirs exist but the permissions are broken.

`--insecure-no-tls` has nothing to do with self-signed certificates cause it disables TLS at all. This option exposes a plain HTTP port which is useful for development or if you deploy a plugin behind some sort of reverse proxy like Traefik.

utils/http.go
```diff
@@ -325,6 +349,17 @@ func (h *HTTP) EnsureCert(defaultPath string) (err error) {
 
 	log.Warningf("Generating self signed key and cert to %v %v.", h.KeyFile, h.CertFile)
 
+	certDir := path.Dir(defaultCertPath)
+	if _, err = os.Stat(certDir); os.IsNotExist(err) {
+		log.Warningf("Self-signed TLS certs directory %v doesn't exist, creating it now", certDir)
```
marshall-lee, line 354: I think it must be `log.Debugf`

utils/http.go
marshall-lee: Lets do both — keep this Mkdir *and* clarify the "pem write error" message.

xnutsive:
@marshall-lee, cleaned up the code, rebased with Master. ;)

PR #126, opened 2020-08-19
teleport-slack in notification-only mode

### LINKED ISSUE #121: Teleport Slack Plugin as notification system only
Customer request came in looking for feasibility to only pass a notification via Slack channel when a request comes in and have it only state `pending request`. They still want to have the approve/deny happen via tctl. 

Use case:

Their Slack is managed by their IT so their IT has access to any channel, so basically anyone from IT can request admin role and then go to the Slack channel and approve it themselves.

With the current setup of the Slack plugin 2 concerns were brought up:

1. It potentially allows people to bypass teleport access control via slack
2. You don’t have the possibility to log who actually take the action

xnutsive:
This PR adds a read only (no buttons, i.e. notification only) mode for the Teleport Slack plugin, as @aelkugia and @benarent suggested. Closes #121.

In includes several behavior changes: 
1. The plugin doesn't render buttons in Slack notifications if `notification_only = true` in the config. 
2. The default is `notifications_only = false`. 
3. The plugin still starts up the http service, but doesn't process and errors out if a webhook from Slack is invoked in notifcation_only mode. 
4. The readme documents this + the `update` verb in role config. 

@marshall-lee, I've added an extra arg to `startApp` in tests, and I'm not sure if that is reasonable approach or not (everything else just takes `c *C`). Otherwise, this is ready for review.

TODOs:
- [x] Update example config so that `teleport-slack configure` works correctly
- [x] Add tests for readonly mode
- [x] Update the readme and make sure the docs will be updated, too

After the initial round of feedback: 
- [x] Rename readonly to notification_only in the config. 
- [x] Add a comment to the readme file that one could just not provide `update` permissions to the user.

### Checklist before merging

- [x] Check that this PR works as expected manually (@xnutsive)
- [ ] Confirm that we want this behavior in the plugin (@benarent, @aelkugia).

### DIFF
--- a/access/slack/README.md
+++ b/access/slack/README.md
@@ -24,6 +24,7 @@ root_cas = "/var/lib/teleport/plugins/slack/auth.cas"    # Teleport cluster CA c
 token = "api-token"         # Slack Bot OAuth token
 secret = "secret-value"     # Slack API Signing Secret
 channel = "channel-name"    # Message delivery channel
+readonly = false            # Whether run the plugin in notification mode (Don't allow approval/denial via a Slack button)
 
 [http]
 # listen_addr = ":8081" # Network address in format [addr]:port on which callback server listens, e.g. 0.0.0.0:443
--- a/access/slack/README.md
+++ b/access/slack/README.md
@@ -24,7 +24,7 @@ root_cas = "/var/lib/teleport/plugins/slack/auth.cas"    # Teleport cluster CA c
 token = "api-token"         # Slack Bot OAuth token
 secret = "secret-value"     # Slack API Signing Secret
 channel = "channel-name"    # Message delivery channel
-readonly = false            # Whether run the plugin in notification mode (Don't allow approval/denial via a Slack button)
+notification_only = false   # Whether run the plugin in notification only mode (Don't allow approval/denial via a Slack button)
 
 [http]
 # listen_addr = ":8081" # Network address in format [addr]:port on which callback server listens, e.g. 0.0.0.0:443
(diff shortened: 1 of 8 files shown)


access/slack/README.md
```diff
@@ -24,6 +24,7 @@ root_cas = "/var/lib/teleport/plugins/slack/auth.cas"    # Teleport cluster CA c
 token = "api-token"         # Slack Bot OAuth token
 secret = "secret-value"     # Slack API Signing Secret
 channel = "channel-name"    # Message delivery channel
+readonly = false            # Whether run the plugin in notification mode (Don't allow approval/denial via a Slack button)
```
benarent, line 27: I would change this to `notification_only = true` to be more explicit of what's being turned off, we should keep it as false as default and not require it to be defined.

marshall-lee:
We didn't get a confirmation is this feature really needed. I believe the case described in #121 could be solved using private channels only so for me `notify_only` option is a solution to a different problem

xnutsive:
@benarent, I'll rename the option, np. Before we decide to merge this in, though — let's decide on the right approach in #121.
Most of my PRs are hobby exercises in golang, so no hard feelings if we decide to shelf any of them.

xnutsive:
@fspmarshall @webvictim This one is ready for your review ;) 

The idea here is to do both: tell users that they can just not provide permissions to the user (that might result in an error if they later on decide to run with buttons), and also implement the notification_only mode that doesn't show the buttons, and has an internal check in the webhook server, too.

access/slack/README.md
xnutsive: Done!

xnutsive:
@benarent, I checked if this works using the docker flow ;) It works. 

You commented the other day on a different PR that you'd like to have notification_only with additions to readme — this PR seems to be in good shape and implements that, but before I merge this, I want to make sure this behavior is something you want in the plugin. 

To summarize: 
- teleport-slack now has notification_only = true/false config option. 
- If notification_only = true, Slack notification messages when new request is filed will not have the approval/denial buttons
- If for any reason the plugin receives a webhook in notification_only mode, it'll log it as error and not process it. 
- The Readme has more info on extra security and reducing the plugin user permission verbs to just list/read.

access/slack/config.go
```diff
@@ -6,18 +6,21 @@ import (
...
+// SlackConfig holds Slack-specific configuration options.
 type SlackConfig struct {
-	Token   string `toml:"token"`
-	Secret  string `toml:"secret"`
-	Channel string `toml:"channel"`
-	APIURL  string
+	Token            string `toml:"token"`
+	Secret           string `toml:"secret"`
+	Channel          string `toml:"channel"`
+	NotificationOnly bool   `toml:"notification_only"`
```
marshall-lee, line 22: nit: i'd prefer the adjective form more: `NotifyOnly`, `notify_only`, etc

Issue #120, opened 2020-08-12
Teleport Plugins 0.1.0 Webhook doesn't seem to be working. 


benarent:
I have been testing with 0.1.0 and I'm having issues with the new webhook setup. 

```
DEBU   DEBUG logging enabled logrus/exported.go:117
INFO   Starting Teleport Access PagerDuty extension 0.1.0:teleport-pagerduty-v0.1.0-0-g22dc5d2 pagerduty/app.go:74
DEBU   Checking Teleport server version pagerduty/app.go:152
DEBU   Starting PagerDuty API health check... pagerduty/app.go:107
DEBU   PagerDuty API health check finished ok pagerduty/app.go:112
DEBU   Starting secure HTTPS server on 66.220.18.10:8883 utils/http.go:235
DEBU   Setting up the webhook extensions pagerduty/app.go:125
DEBU   PagerDuty webhook extensions setup finished ok pagerduty/app.go:130
DEBU   Watcher connected access/service_job.go:62
INFO   PagerDuty incident created pd_incident_id:PUZCFSH request_id:17aadf67-24bb-48d5-be8b-838015763a1b pagerduty/app.go:282
```
## PD Setup
It looks like the extensions might be incorrectly setup? 
![image](https://user-images.githubusercontent.com/559288/90052497-112f4a00-dc8e-11ea-97f2-45e0ccccf04a.png)

https://dev-teleport-test.pagerduty.com/service-directory/P4IC1VD/undefined?addonId=PZ5RU2P
![image](https://user-images.githubusercontent.com/559288/90052556-27d5a100-dc8e-11ea-8254-8f0393cbe094.png)

## Teleport Config File
```
 # example teleport-pagerduty configuration TOML file
[teleport]
auth_server = "teleport-ent-4-3.practice.io:3025"                            # Teleport Auth Server GRPC API address
client_key = "/var/lib/teleport/plugins/pagerduty/auth.key" # Teleport GRPC client secret key
client_crt = "/var/lib/teleport/plugins/pagerduty/auth.crt" # Teleport GRPC client certificate
root_cas = "/var/lib/teleport/plugins/pagerduty/auth.cas"   # Teleport cluster CA certs

[pagerduty]
api_key = "fzPbHPxxx"               # PagerDuty API Key
user_email = "ben@gravitational.com" # PagerDuty bot user email (Could be admin email)
service_id = "P4IC1VD"        # PagerDuty service id

[http]
[...]

PR #127, opened 2020-08-19
Making config examples more explicit

benarent:
Helps make https://github.com/gravitational/teleport-plugins/issues/120 clearer, as port mapping can be confusing.

### DIFF
--- a/access/gitlab/config.go
+++ b/access/gitlab/config.go
@@ -46,7 +46,7 @@ project_id = "1812345"                     # GitLab Project ID
 webhook_secret = "your webhook passphrase" # A secret used to encrypt data we use in webhooks. Basically anything you'd like.
 
 [http]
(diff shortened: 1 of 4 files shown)


xnutsive:
I think I had a problem with explicitly putting the port number in slack's public_addr field. Let me test this for a couple plugins and make sure it works as we think it works ;)

benarent:
@xnutsive Did you get a chance to test?

xnutsive:
Nope, but done all the preparation steps. I’ll test tomorrow with the docker flow, and also share the test apps/credentials I’m using.

What’s the best way to store a few secret config so that they’re available to the team? Team 1Password or absolve vault in the repo with a shared password?

Nate
(650) 933-7237

benarent:
I'm going to close this PR as we've made a large amount of progress with the addons.

PR #128, opened 2020-08-19
WaitForReady false by default

### LINKED ISSUE #102: `WaitForReady` should be configurable.
The `WaitForReady` option is always set to `true` for the GRPC client in `access/access.go`.  This ends up suppressing a lot of useful errors, making debug difficult.  We should make it configurable, and possibly off by default, in order to support an easier debug workflow.

xnutsive:
- Closes #102.
- Makes WaitForReady false by default on access.go NewClient()
- Also removes the default grcp.WithMaxBackoffTime(time.Second*2)
- Provides a way to supply additional grpc.DialOptions if needed
- All of the plugins use the previous options for now, to make
  sure this commit doesn't break any existing behavior.

### DIFF
--- a/access/access.go
+++ b/access/access.go
@@ -35,7 +35,7 @@ import (
 	"github.com/gravitational/trace"
 )
 
-// Minimal Teleport version the plugin supports.
+// MinServerVersion is the minimal teleport version the plugin supports.
 const MinServerVersion = "4.3.0"
 
 // State represents the state of an access request.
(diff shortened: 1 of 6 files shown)


access/access.go
```diff
@@ -136,13 +136,14 @@ type clt struct {
 	cancel context.CancelFunc
 }
 
-func NewClient(ctx context.Context, plugin string, addr string, tc *tls.Config) (Client, error) {
+// NewClient creates a new Teleport GRPC API client and returns it.
+//
+// TODO: Nate've removed the 2 seconds back-off time,
```
xnutsive, line 141: Uh oh, I'll need to delete those commends if we keep the default options via the plugins app.go.

access/gitlab/app.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	"github.com/gravitational/teleport-plugins/access"
 	"github.com/gravitational/teleport-plugins/utils"
+	"google.golang.org/grpc"
```
xnutsive, line 10: If the default options are fine — then the client application doesn't have to import grpc.

benarent:
Let's make it notification only, it reduces the authorisation scope for companies that really want to use tctl

xnutsive:
@benarent, looks like you've meant to post this on another PR, I'll copy it there ;) 

@fspmarshall, Ironically, I've been playing around with another plugin idea for Teleport, and hit exactly that problem (the grpc client hanging with wrong certificates, without an error message). 

I'll merge this PR and make another one with @marshall-lee, he seemed to use this behavior somehow, I want to make sure I won't break things too bad and those error messages will be wrapped and logged correctly. 

Thanks for your feedback a ton! <3

PR #136, opened 2020-09-04
Introducing teleport-webhooks 

xnutsive:
## WTF is this

`teleport-plugins` allows building the approval flows with webhooks. 

## Zomg why did you do it

I think this will be extremely useful in demo purposes for Gravitational clients and their engineering teams. Plus, I wanted to understand the plugins better, and write some Go.

## Any surprises? 

Yeah, there's no unit tests yet. But there's the end to end docker test flow, and it works.

Tagging @fspmarshall, @marshall-lee, @klizhentas, and @benarent for review. What do you think?


benarent (approved):
This looks like a helpful addition, It'll be a helpful base for others to learn about the plugin.

Sorry for the delayed review.

xnutsive:
Thanks for the feedback! I'll rebase and merge this, and write more docs and an article a bit later.

PR #143, opened 2020-11-16
Sample Tomcat AAP Example

stevenGravy:
Example Tomcat application that uses the Teleport JSON Web Token (JWT) to allow access and display the user info.


benarent (approved):
@stevenGravy I tried this application today and I found a few typos / issues with it running. I just went ahead and added these changes to your branch.

benarent:
@stevenGravy Can you PTAL at my changes, I think this should be good to go? Just confirming that you don't have any last min since rolling out this example.

PR #163, opened 2021-01-21
Node App Example Updates for more error and verification info

stevenGravy:
Updated to the following

Provides where the JWT token was verified against
image

Also gives Access Denied info if JWT token is missing or can't be verified.

### DIFF
--- a/apps/example-nodejs-jwt/app.js
+++ b/apps/example-nodejs-jwt/app.js
@@ -77,7 +77,7 @@ app.get("/", function (req, res) {
        <li><%= role %></li>
      <% }); %>
    </ul>
-     <p> Your JSON Web Token (JWT) was verified against <a href="<%= proxyAddr %><%= jwtResource %>"><%= proxyAddr %><%= jwtResource %></a> </p>
+     <p> Your JSON Web Token (JWT) was sent via <a href="<%= proxyAddr %>"><%= proxyAddr %></a> and verified against <%= proxyAddr %><%= jwtResource %></p>
 `,
       { username: username, roles: roles, proxyAddr: proxyAddr, jwtResource: jwtResource }
     )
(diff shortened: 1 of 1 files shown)


apps/example-nodejs-jwt/app.js
```diff
@@ -48,12 +49,12 @@ function getKey(header, callback) {
...
   const token = req.headers["teleport-jwt-assertion"];
   if (!token) {
-    res.status(403).send("Access denied");
+    res.status(403).send("Access denied. No JWT Token present.");
     return;
   }
   jwt.verify(token, getKey, function (err, decoded) {
     if (err) {
-      res.status(403).send("Access denied");
+      res.status(403).send("Access denied. Error in verifying token.");
```
benarent, line 57: +1 for a making the error message clearer.

apps/example-nodejs-jwt/app.js
```diff
@@ -75,12 +77,13 @@ app.get("/", function (req, res) {
        <li><%= role %></li>
      <% }); %>
    </ul>
+     <p> Your JSON Web Token (JWT) was verified against <a href="<%= proxyAddr %><%= jwtResource %>"><%= proxyAddr %><%= jwtResource %></a> </p>
```
benarent, line 80: I think it's better to provide a link back to root cluster, but I've kept in the jwks url but removed the hyperlink for it. 
```suggestion
     <p> Your JSON Web Token (JWT) was sent via <a href="<%= proxyAddr %>"><%= proxyAddr %></a> and verified against <%= proxyAddr %><%= jwtResource %></p>
```

benarent (approved):
@stevenGravy PTAL at my suggestion, I don't see much value in deep linking to the jwks page but printing it out is useful for debugging and learning more.

PR #135, opened 2020-09-01
Enable go test -race

marshall-lee:
This PR enables `-race` option for `go test` and adds some racy tests emulating some load.

PR #178, opened 2021-01-27
Fix Darwin build

marshall-lee:


marshall-lee:
Closed in favor of #135

PR #196, opened 2021-02-22
New Access workflow for Slack plugin

### LINKED ISSUE #183: Split teleport-slack OSS and teleport-slack Enterprise versions
The first part of upgrading Teleport Slack integration to the new Access Workflows API. 

**Teleport Enterprise** will provide a web UI to work with Access Workflows. With that, we can reduce the amount of trusted services to just Teleport, and all of our integrations will work just as a notification channels, only require `requests_read` permission to function. 

This issue covers splitting out teleport-slack Enterprise and teleport-slack OSS. 

### Teleport Slack OSS

Should largely work as it works now, with some updates to the Access Request API described in #184. It would still post either just a notification of a new access request, or a notification with approve / decline buttons to a shared channel. 

### Teleport Slack Enterprise

- [ ] **The plugin should be able to post access requests to multiple people DMs, based on who on Slack has the permission to accept or deny the request.**

The DMs should only include a link back to Teleport Web UI to a specific access request. The user will click the link, open the request in the web UI, and approve it from Teleport. 

This will require the plugin to know the Slack to Teleport user mapping, so the plugin can figure out who to DM. This is not currently implemented. 

- [ ] **Send the accept / denied notification to a specified channel, optionally.** 

The user should be able to specify if they want notifications in any channels, and if they do — do they want all notifications, including new requests, or just the approved / denied updates to the requests.

- [ ] **Update the documentation in install.md**

- [ ] **Update the documentation on the website**

marshall-lee:
Closes #183 
Related to #205


klizhentas (commented):
* Remove approval buttons from slack completely
* Use api method to detect if it's Enterprise or OSS https://github.com/gravitational/teleport/issues/5652
* If it's OSS print CLI instructions, otherwise print Public Addr from discovery endpoint

benarent:
@marshall-lee Do you have an expected timeframe to address the comments. I have to test Slack plugins as part of the 6.0 test plan and it would be good to try / test with the latest plugin if it's ready. https://github.com/gravitational/teleport/issues/5654

marshall-lee:
@benarent I'm planning to finish the most of my job this week but i largely depend on https://github.com/gravitational/teleport/issues/5652

access/slack/bot.go
```diff
@@ -148,17 +177,30 @@ func (b *Bot) msgSections(reqID string, reqData RequestData, status string) []sl
...
+		msgFieldToBuilder(&builder, "Reason", reqData.RequestReason)
+	}
+	if b.webProxyURL != nil {
+		reqURL := *b.webProxyURL
+		reqURL.Path = lib.BuildURLPath("web", "requests", reqID)
+		msgFieldToBuilder(&builder, "Link", reqURL.String())
+	} else {
+		if status != "EXPIRED" {
+			msgFieldToBuilder(&builder, "Approve", fmt.Sprintf("tctl requests approve %s", reqID))
+			msgFieldToBuilder(&builder, "Deny", fmt.Sprintf("tctl requests deny %s", reqID))
```
fspmarshall, line 190: For teleport version`6.1` (coming out in ~3 weeks) onwards the recommended method of approval/denial via CLI will be:
- `tsh request review --aprove <request-id>`
- `tsh request review --deny <request-id>`

Probably best to just start recommending that now.

access/slack/app.go
```diff
@@ -2,28 +2,27 @@ package main
...
+	"github.com/gravitational/teleport/api/types"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/backoff"
 
 	"github.com/gravitational/trace"
 )
 
 // MinServerVersion is the minimal teleport version the plugin supports.
-const MinServerVersion = "4.3.0"
+const MinServerVersion = "6.0.0-alpha.2"
```
fspmarshall, line 18: Are we deliberately keeping compatibility with `6.0` or is this just what was being used to test against?  Seems to me that these changes make more sense if they come alongside `6.1` since that is when we're adding approval/denail to the Web UI.

klizhentas: We are targeting 6.1 version, so compatibility with 6.0 is not required. Our plugins will follow the same versioning scheme as Teleport, so the next plugins version will be 6.1

marshall-lee: @fspmarshall @klizhentas 

It's temporarily! Teleport version in this branch is still `6.0.0-alpha.2` and until Teleport 6.1 is released, the plugin will refuse to start.

access/access.go
```diff
@@ -211,16 +220,16 @@ func (c *clt) GetRequests(ctx context.Context, fltr Filter) ([]Request, error) {
 }
 
 func (c *clt) CreateRequest(ctx context.Context, user string, roles ...string) (Request, error) {
-	req := &services.AccessRequestV3{
-		Kind:    services.KindAccessRequest,
-		Version: services.V3,
-		Metadata: services.Metadata{
+	req := &types.AccessRequestV3{
```
Joerger, line 223: Use `services.NewAccessRequest(user, roles...)`, which is functionally equivalent.

fspmarshall (approved):

klizhentas (approved):

PR #199, opened 2021-03-05
add dockerfile for building and using slack plugin

khous:
Is there any interest in building and publishing plugins as docker images?
This will make it a lot easier for kubernetes users to use these plugins.

### DIFF
--- a/Dockerfile.slack
+++ b/Dockerfile.slack
@@ -1,0 +1,14 @@
+FROM golang:1.16.0-alpine AS build
+RUN mkdir -p /go/src/github.com/gravitational/teleport-plugins/
+WORKDIR /go/src/github.com/gravitational/teleport-plugins/
+COPY . .
+RUN apk update && apk add make
+RUN cd access/slack
+RUN make && ls access/slack/*
+
+FROM alpine:3.12 as plugin
+RUN mkdir -p tmp/slack
+WORKDIR /tmp/slack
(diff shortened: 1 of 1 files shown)


klizhentas:
@khous great idea, I think we should do it. We'd probably choose a different base image, but otherwise agree 100%

klizhentas:
@webvictim can you give us guidance on what stack we should use for setting up the publishing?

khous:
@klizhentas Thanks for your enthusiasm. What is the preferred base image?

klizhentas:
We'd probably want to keep the image similar to what we use for standard Teleport:

https://github.com/gravitational/teleport/blob/master/build.assets/Dockerfile#L9

khous:
> We'd probably want to keep the image similar to what we use for standard Teleport:
> 
> https://github.com/gravitational/teleport/blob/master/build.assets/Dockerfile#L9

Thanks, I'll take a look

webvictim:
As @klizhentas says we use `ubuntu:18.04` for our Teleport build box (for a lower glibc version requirement on the built binaries) but our [public Teleport images use `ubuntu:20.04` as a base](https://github.com/gravitational/teleport/blob/master/build.assets/charts/Dockerfile).

Joerger:
This is now being handled in https://github.com/gravitational/teleport-plugins/pull/356, thank you for your contribution and suggestion.

PR #200, opened 2021-03-17
Unify linter settings with Teleport repo and add YAML config

webvictim:
Updates https://github.com/gravitational/teleport/issues/6013


klizhentas (approved):

.golangci.yml
```diff
@@ -0,0 +1,31 @@
+issues:
```
awly, line 1: nit: we used to set `--skip-dirs` instead of using the defaults because by default golangci-lint skips `examples/`

webvictim: I did notice this while creating the file, but https://github.com/golangci/golangci-lint/blob/master/.golangci.example.yml#L22-L34 makes it look like even if you set `skip-dirs` then `skip-dirs-use-default` is still set to `true` by default. I haven't validated this though.

If the original intention was to actively suppress that behaviour then I'll PR a new change:

```yaml
issues:
  skip-dirs:
    - vendor
  skip-dirs-use-default: false
```

PR #201, opened 2021-03-18
Re-adds --skip-dirs=vendor and disables default skipping behaviour

webvictim:
As per @awly's comment in https://github.com/gravitational/teleport-plugins/pull/200#discussion_r596419420, the original intention here was to only skip the `vendor` directory and test everything else.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -28,4 +28,7 @@ output:
   uniq-by-line: false
 
 run:
+  skip-dirs:
+    - vendor
+  skip-dirs-use-default: false
   timeout: 5m
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,6 @@
 issues:
   exclude-rules:
-    - linters: staticcheck
+    - linters: gosimple
       text: "S1002: should omit comparison to bool constant"
   exclude-use-default: true
   max-same-issues: 0


.golangci.yml
```diff
@@ -1,6 +1,6 @@
 issues:
   exclude-rules:
-    - linters: staticcheck
+    - linters: gosimple
```
awly, line 3: was this intentional?
`S1002` is from staticcheck: https://staticcheck.io/docs/checks#S1002

awly (approved):

.golangci.yml
webvictim: It looks like [the S1002 check is still technically part of `gosimple`](https://github.com/dominikh/go-tools/blob/master/simple/analysis.go#L20-L23), but everything is now bundled under the `staticcheck` name because `gosimple` merged with `staticcheck` (see https://github.com/dominikh/go-simple -> https://github.com/dominikh/go-tools)

awly: Oh, in that case should we drop `gosimple` from our list of linters?

webvictim: If I remove `gosimple` from the list, it seems like `staticcheck` doesn't run the same checks as there are no S1002 warnings, even when I remove the filter. I wouldn't be confident that it's running everything `gosimple` did.

awly: Woah, something is really odd. When I run `staticcheck` manually, it finds a lot more stuff:
```
$ staticcheck ./... | ag -v S1002
e/api/cloud/client.go:13:2: package "google.golang.org/grpc" is being imported more than once (ST1019)
	e/api/cloud/client.go:14:2: other import of "google.golang.org/grpc"
e/lib/auth/plugin.go:9:2: package "github.com/gravitational/teleport/lib/auth" is being imported more than once (ST1019)
	e/lib/auth/plugin.go:10:2: other import of "github.com/gravitational/teleport/lib/auth"
examples/go-client/client.go:18:15: error strings should not be capitalized (ST1005)
integration/helpers.go:1006:5: ineffective break statement. Did you mean to break out of the outer loop? (SA4011)
integration/integration_test.go:4813:5: ineffective break statement. Did you mean to break out of the outer loop? (SA4011)
integration/kube_integration_test.go:1169:2: tlsConfig.BuildNameToCertificate has been deprecated since Go 1.14: NameToCertificate only allows associating a single certificate with a given name. Leave that field nil to let the library select the first compatible chain from Certificates.  (SA1019)
lib/auth/auth_test.go:36:2: package "github.com/gravitational/teleport/lib/auth/testauthority" is being imported more than once (ST1019)
	lib/auth/auth_test.go:37:2: other import of "github.com/gravitational/teleport/lib/auth/testauthority"
lib/auth/middleware.go:219:28: this comparison is always true (SA4023)
	lib/auth/middleware.go:219:13: the lhs of the comparison is the 1st return value of this function call
	lib/multiplexer/tls.go:108:1: (*github.com/gravitational/teleport/lib/multiplexer.TLSListener).Serve never returns a nil interface value
lib/auth/password_test.go:40:2: package "gopkg.in/check.v1" is being imported more than once (ST1019)
	lib/auth/password_test.go:41:2: other import of "gopkg.in/check.v1"
lib/client/api.go:2701:16: error strings should not be capitalized (ST1005)
lib/client/api.go:2710:16: error strings should not be capitalized (ST1005)
lib/client/api.go:2715:16: error strings should not be capitalized (ST1005)
lib/config/configuration.go:1474:4: ineffective break statement. Did you mean to break out of the outer loop? (SA4011)
lib/events/slice.pb.go:6:8: package github.com/golang/protobuf/proto is deprecated: Use the "google.golang.org/protobuf/proto" package instead.  (SA1019)
lib/events/slice.pb.go:53:10: xxx_messageInfo_SessionSlice.Marshal is deprecated: Do not use; this method existed for intenal-use only.  (SA1019)
lib/events/slice.pb.go:64:2: xxx_messageInfo_SessionSlice.Merge is deprecated: Do not use; this method existed for intenal-use only.  (SA1019)
lib/events/slice.pb.go:70:2: xxx_messageInfo_SessionSlice.DiscardUnknown is deprecated: Do not use; this method existed for intenal-use only.  (SA1019)
lib/events/slice.pb.go:73:34: proto.InternalMessageInfo is deprecated: Do not use; this type existed for intenal-use only.  (SA1019)
lib/events/slice.pb.go:136:10: xxx_messageInfo_SessionChunk.Marshal is deprecated: Do not use; this method existed for intenal-use only.  (SA1019)
lib/events/slice.pb.go:147:2: xxx_messageInfo_SessionChunk.Merge is deprecated: Do not use; this method existed for intenal-use only.  (SA1019)
lib/events/slice.pb.go:153:2: xxx_messageInfo_SessionChunk.DiscardUnknown is deprecated: Do not use; this method existed for intenal-use only.  (SA1019)
lib/events/slice.pb.go:156:34: proto.InternalMessageInfo is deprecated: Do not use; this type existed for intenal-use only.  (SA1019)
lib/events/slice.pb.go:208:2: proto.RegisterType is deprecated: Use protoregistry.GlobalTypes.RegisterMessage instead.  (SA1019)
lib/events/slice.pb.go:209:2: proto.RegisterType is deprecated: Use protoregistry.GlobalTypes.RegisterMessage instead.  (SA1019)
lib/events/slice.pb.go:976:15: proto.RegisterFile is deprecated: Use protoregistry.GlobalFiles.RegisterFile instead.  (SA1019)
lib/jwt/jwt.go:34:2: package "gopkg.in/square/go-jose.v2/jwt" is being imported more than once (ST1019)
	lib/jwt/jwt.go:35:2: other import of "gopkg.in/square/go-jose.v2/jwt"
lib/kube/proxy/forwarder.go:1670:2: tlsConfig.BuildNameToCertificate has been deprecated since Go 1.14: NameToCertificate only allows associating a single certificate with a given name. Leave that field nil to let the library select the first compatible chain from Certificates.  (SA1019)
lib/utils/cli.go:35:2: package "github.com/sirupsen/logrus" is being imported more than once (ST1019)
	lib/utils/cli.go:36:2: other import of "github.com/sirupsen/logrus"
```

awly: OK, looks like:
- any `ST*` checks are considered a part of `stylecheck` linter in `golangci-lint` and is somehow filtered out
  - we need to enable `stylecheck`
- any `S[0-9]*` checks are considered `gosimple` and similarly filtered out from `staticcheck` somehow
- `SA4011` is filtered by the default exclude rules in `golangci-lint`
  - we should use our own exclude list instead of the default one
- `SA1019` I don't know but also seems useful (detecting deprecated types/functions)

Anyway, we can merge this PR and I'll tweak the config afterwards.

PR #212, opened 2021-04-16
jira: Rewrite plugin to use new API client and support new Teleport features

### LINKED ISSUE #207: Move JIRA plugin to teleport/api

marshall-lee:
- Use new `teleport/api` client.
- Add comments to the issue when new access review is submitted.
- Automatically change issue status once request is approved/denied.
- Manage plugin data state safely using compare-and-swap.
- Migrate tests to testify.
- Use `addr` config setting instead of deprecated `auth_server`.

Closes #207.


klizhentas:
@marshall-lee is that still work in progress?

marshall-lee:
@klizhentas yes i returned it to work to support more features as other plugins do

access/jira/README.md
```diff
@@ -39,7 +39,7 @@ make access-jira
 ```toml
 [teleport]
 auth_server = "example.com:3025"                       # Teleport Auth Server GRPC API address
```
Joerger, line 41:
```suggestion
auth_server = "example.com:3025"                       # Teleport Auth/Proxy Server address (should be port 443 for Teleport Cloud)
```

access/jira/README.md
```diff
@@ -39,7 +39,7 @@ make access-jira
...
-clientkey = "/var/lib/teleport/plugins/jira/auth.key"  # Teleport GRPC client secret key
```
Joerger, line 42: Add identity option

access/jira/README.md
```diff
@@ -66,6 +66,16 @@ server & manage access-requests. Use `tctl auth sign --format=tls` to generate
 the required PEM files, and make sure that the Auth Server's GRPC API is
 accessible at the address indicated by `auth_server`.
 
+`auth_server` can also be a proxy (e.g. when using a plugin with Teleport Cloud),
+in that case use `tctl auth sign --format=file` to export the identity file
+containing SSH keys and set the `identity` file path in a config.
+
+```TOML
+[teleport]
+auth_server = "yourproxy.teleport.sh"            # Teleport proxy address
```
Joerger, line 75:
```suggestion
auth_server = "yourproxy.teleport.sh:443"            # Teleport proxy address
```

access/jira/README.md
```diff
@@ -66,6 +66,16 @@ server & manage access-requests. Use `tctl auth sign --format=tls` to generate
 the required PEM files, and make sure that the Auth Server's GRPC API is
 accessible at the address indicated by `auth_server`.
 
+`auth_server` can also be a proxy (e.g. when using a plugin with Teleport Cloud),
```
Joerger, line 69: You could avoid adding this explanation (here and in #220) by making `file` format and `IdentityCredentials` the default/recommended credential method. You could leave the `tls` format and `KeyPairCredentials` just for backwards compatiblity.

access/jira/app.go
```diff
@@ -137,60 +112,123 @@ func (a *App) run(ctx context.Context) (err error) {
...
+	return nil
+}
+
+func (a *App) checkTeleportVersion(ctx context.Context) (proto.PingResponse, error) {
 	log := logger.Get(ctx)
 	log.Debug("Checking Teleport server version")
-	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
-	defer cancel()
-	pong, err := a.accessClient.Ping(ctx)
+	pong, err := a.apiClient.WithCallOptions(grpc.WaitForReady(true)).Ping(ctx)
```
Joerger, line 178: This shouldn't be needed with the new client
```suggestion
	pong, err := a.apiClient.Ping(ctx)
```

access/jira/app.go
```diff
@@ -297,82 +338,250 @@ func (a *App) onJIRAWebhook(ctx context.Context, webhook Webhook) error {
...
+		return trace.Wrap(err)
+	}
+
+	ctx, log := logger.WithFields(ctx, logger.Fields{
+		"jira_issue_id":  data.IssueID,
+		"jira_issue_key": data.IssueKey,
+	})
+	log.Info("JIRA Issue created")
+
+	_, err = a.modifyPluginData(ctx, reqID, func(existing *PluginData) (PluginData, bool) {
```
Joerger, line 431: Same comment as for Pagerduty - https://github.com/gravitational/teleport-plugins/pull/220#discussion_r655796923
Can you name these anonymous functions, perhaps with a comment, to make it clear what their purpose is?

access/jira/config.go
```diff
@@ -67,18 +67,6 @@ func LoadConfig(filepath string) (*Config, error) {
 }
 
 func (c *Config) CheckAndSetDefaults() error {
-	if c.Teleport.AuthServer == "" {
```
Joerger, line 70: Can you make the same changes here that I noted in Pagerduty
https://github.com/gravitational/teleport-plugins/pull/220#discussion_r655791392
https://github.com/gravitational/teleport-plugins/pull/220#discussion_r655798486

Joerger (approved):

access/jira/README.md
marshall-lee: `*.teleport.sh` names work without specifying a port

access/jira/app.go
```diff
@@ -297,82 +375,272 @@ func (a *App) onJIRAWebhook(ctx context.Context, webhook Webhook) error {
...
+	}
+	log.Info("Successfully resolved the issue")
 
 	return nil
 }
 
-func (a *App) getPluginData(ctx context.Context, reqID string) (PluginData, error) {
-	dataMap, err := a.accessClient.GetPluginData(ctx, reqID)
+// modifyPluginData performs a compare-and-swap update of access request's plugin data.
+func (a *App) modifyPluginData(ctx context.Context, reqID string, fn func(data *PluginData) (PluginData, bool)) (bool, error) {
```
tcsc, line 588: Might be worth mentioning in the docstring that `fn` may be invoked many times and shouldn't do anything with a heap of side-effects. Also might be worth mentioning _why_ it needs to be a compare-and-set operation as well. 

(I don't know the plugin API well, so I assume its because there is a chance someone could sweep in and alter the server-side plugin config while this update is in transit?)

access/jira/app.go
```diff
@@ -1,23 +1,56 @@
...
+	// pluginName is used to tag PluginData and as a Delegator in Audit log.
+	pluginName = "jira"
+	// backoffMaxDelay is a maximum time GRPC client waits before reconnection attempt.
+	backoffMaxDelay = time.Second * 2
+	// initTimeout is used to bound execution time of health check and teleport version check.
+	initTimeout = time.Second * 10
+	// handlerTimeout is used to bound the execution time of watcher event handler.
+	handlerTimeout = time.Second * 5
+	// maxModifyPluginDataTries is a maximum number of compare-and-swap tries when modifying plugin data.
+	maxModifyPluginDataTries = 5
```
tcsc, line 52: How was this value  arrived at?

access/jira/app.go
tcsc: Design question - would it make sense to keep trying to modify the config until it succeeds (or fails with a non-`CompareFailed` error) or the supplied `ctx`expires?

That way you can have one, top-level timeout controlling how long to try and start the plugin, rather than arbitrary retries scattered throughout the code.

Just a suggestion.

tcsc: Also, it would be nice to have what the update function documented (e.g., what does returning `false` imply for the behaviout of `modifyPluginData()`.

access/jira/app.go
```diff
@@ -223,55 +296,60 @@ func (a *App) onJIRAWebhook(ctx context.Context, webhook Webhook) error {
...
-		if trace.IsNotFound(err) {
-			log.WithError(err).Warning("Cannot process expired request")
-			return nil
-		}
 		return trace.Wrap(err)
 	}
+	if len(reqs) == 0 {
+		log.Debug("Cannot process expired request")
+	}
+	req := reqs[0]
```
tcsc, line 320: This will still explode on an empty `reqs`, though yes?

access/jira/app.go
marshall-lee: > I don't know the plugin API well, so I assume its because there is a chance someone could sweep in and alter the server-side plugin config while this update is in transit?

Yes cause there's some gap between `GetAccessRequests(...)` and `UpdatePluginData(...)` calls.

access/jira/app.go
marshall-lee: Removed it at all after moving to backoff algorithm

access/jira/app.go
marshall-lee: > Design question - would it make sense to keep trying to modify the config until it succeeds (or fails with a non-CompareFailed error) or the supplied ctx expires?

Oh, after some hesitation, I rewrote `modifyPluginData` to use exponential backoff!

lib/backoff/decorr.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"context"
+	"math/rand"
+	"time"
+
+	"github.com/gravitational/trace"
+)
+
+// Decorr is a "decorrelated jitter" inspired by https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/.
+type decorr struct {
```
tcsc, line 28: You might want to use something like the [Clockwork](https://github.com/jonboulle/clockwork) package to allow a mock clock to be injected for testing

lib/backoff/backoff_test.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func TestDecorr(t *testing.T) {
+	t.Parallel()
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	t.Cleanup(cancel)
+
+	base := 200 * time.Millisecond
+	cap := 2 * time.Second
+	delay := 125 * time.Millisecond // Lets have some delay because darwin builds on CI are a bit laggy.
```
tcsc, line 41: You really want to use a mock clock for this, rather than tweaking timeouts for different platforms. Teleport uses [Clockwork](https://github.com/jonboulle/clockwork) internally for this purpose, so I recommend that for now.

lib/backoff/decorr.go
```diff
@@ -0,0 +1,61 @@
...
+	base  int64
+	cap   int64
+	mul   int64
+	sleep int64
+}
+
+// Decorr initializes an alogoritm.
+func Decorr(base, cap time.Duration) Backoff {
```
tcsc, line 36: `NewDecorr()` for a constructor.

access/jira/app.go
```diff
@@ -59,320 +95,597 @@ func (a *App) WaitReady(ctx context.Context) (bool, error) {
...
-			return nil
-		}
 		return trace.Wrap(err)
 	}
 
-	if req.State != access.StatePending {
-		log.WithField("request_state", req.State).Warningf("Cannot process not pending request")
-		return nil
+	var req types.AccessRequest
+	if len(reqs) > 0 {
```
tcsc, line 323: Why do all the plugin data validation below if this condition could cause it to just fail (see line 341) anyway?

If that validation needs to happen regardless of whether `req == nil`, move this condition down closer to where the consequences happen. Alternatively, if all the validation can be avoided, bring the early exit up here.

tcsc (commented):
I like the new `modifyPluginData()` - just a few points about testing the new backoff service, but otherwise LGTM.

access/jira/app.go
marshall-lee: On the line 341 it's not a fail, it calls `resolveIssue`:

```go
	if req == nil {
		return trace.Wrap(a.resolveIssue(ctx, reqID, Resolution{Tag: ResolvedExpired}))
	}
```

lib/backoff/decorr.go
marshall-lee: Done! Clockwork was added.

lib/backoff/backoff_test.go
marshall-lee: Thanks for suggestion!

tcsc (approved):

PR #213, opened 2021-04-16
Support connecting to proxy using identity file credentials

### LINKED ISSUE #194: Slack/Mattermost plugin should work with cloud
Check/Fix if necessary and update docs

marshall-lee:
Closes #194.


access/slack/bot.go
```diff
@@ -41,6 +41,10 @@ func NewBot(conf Config, clusterName, webProxyAddr string) (Bot, error) {
 		if webProxyURL, err = url.Parse(webProxyAddr); err != nil {
 			return Bot{}, err
 		}
+		if webProxyURL.Scheme == "https" && webProxyURL.Port() == "443" {
+			// Cut off redundant :443
+			webProxyURL.Host = webProxyURL.Hostname()
```
klizhentas, line 46: Why check for a specific port, wouldn't .Hostname() always work? Even if there is no port, it should return itself?

klizhentas (changes requested):

access/slack/bot.go
klizhentas: Can you also please cover by tests

marshall-lee: > Even if there is no port, it should return itself?

Yes but if there's a port *and* it's not 443, I need to preserve it.

klizhentas: Why preserve it?

marshall-lee: It's not strictly necessary but links like `https://evilmartians.teleport.sh:443` look weird since `https` already means `443` by default. I wanted to slightly improve the UI by cutting off redundant `:443` part in HTTPS links. However if the port differs from 443 it must be preserved in the URL.

marshall-lee:
Ping @klizhentas @fspmarshall please merge

klizhentas:
closed, will merge as #244

PR #227, opened 2021-04-29
Increase timeout for heavy racy tests

marshall-lee:


klizhentas (approved):

access/gitlab/gitlab_test.go
```diff
@@ -520,7 +520,7 @@ func (s *GitlabSuite) TestRace(c *C) {
 	defer log.SetLevel(prevLogLevel)
 
 	s.cancel() // Cancel the default timeout
-	s.ctx, s.cancel = context.WithTimeout(context.Background(), 10*time.Second)
+	s.ctx, s.cancel = context.WithTimeout(context.Background(), 20*time.Second)
```
r0mant, line 523: Nit: I'd move it to constant so next time you need to adjust this, you'd have to do it once.

r0mant (approved):

PR #228, opened 2021-05-04
Update Slack path in install

inertial-frame:
According to the docs, the install step is run after `make`. However, the generated output path is `build/teleport-slack` not `teleport-slack` - while the fix is easy (`nano` the `install` file and correct the path). It'd probably be easier to update the path.

As is:

Screen Shot 2021-05-04 at 5 12 17 PM

With change:

Screen Shot 2021-05-04 at 5 13 56 PM

I might be mistaken but I think the `install` file is only used following `make`?


inertial-frame:
Confirmed that this remains an issue.

inertial-frame:
After further testing this applies solely to the second installation option - when downloading the source itself. We should instead add this as a note rather than a change to the install file.

PR #231, opened 2021-05-07
Update terraform getting started doc to use identity file.

Joerger:
Identity file is now the suggested credential loader for terraform because it can connect through a Teleport Proxy.

Fixes https://github.com/gravitational/teleport/issues/6566

### DIFF
--- a/terraform/GETTING_STARTED.md
+++ b/terraform/GETTING_STARTED.md
@@ -69,6 +69,8 @@ terraform {
 }
 
 provider "teleport" {
+  # Update addr to point to Teleport Auth/Proxy
+  # e.g. addr = "example.teleport.sh:3025"
   addr               = "localhost:3025"
   identity_file_path = "terraform-identity"
 }
--- a/terraform/GETTING_STARTED.md
+++ b/terraform/GETTING_STARTED.md
@@ -51,7 +51,7 @@ Run:
 
 ```
 tctl create terraform.yaml
-tctl auth sign --format=tls --user=terraform --out=tf --ttl=10h
+tctl auth sign --format=file --user=terraform --out=terraform-identity --ttl=10h
 ```
 
 # Create Terraform configuration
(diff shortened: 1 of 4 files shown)


terraform/GETTING_STARTED.md
```diff
@@ -69,10 +69,8 @@ terraform {
 }
 
 provider "teleport" {
-  addr         = "localhost:3025"
-  cert_path    = "tf.crt"
-  key_path     = "tf.key"
-  root_ca_path = "tf.cas"
+  addr               = "localhost:3025"
```
benarent, line 72:
```suggestion
  # Update addr to point to Teleport Auth/Proxy
  # e.g. addr = "example.teleport.sh:3025"
  addr               = "localhost:3025"
```

terraform/GETTING_STARTED.md
```diff
@@ -51,7 +51,7 @@ Run:
 
 ```
 tctl create terraform.yaml
-tctl auth sign --format=tls --user=terraform --out=tf --ttl=10h
+tctl auth sign --format=file --user=terraform --out=terraform-identity --ttl=10h
 ```
```
benarent, line 56: Tip: Make sure the user can impersonate the new terraform role and user. 

```
spec:
  allow:
    impersonate:
      roles:
      - terraform
      users:
      - terraform
```

Joerger: I'm not sure I understand. Why does the Terraform user/role need to impersonate itself?

benarent: > Why does the Terraform user/role need to impersonate itself?

I had to add it my role, `access`, meaning that the `access` role can impersonate the user. This is important for cloud users. As they'll want to impersonate as the user and not login and obtain the certs as that user.

klizhentas (approved):

r0mant (approved):

PR #238, opened 2021-05-19
Add configurable IssueType to Jira plugin

### LINKED ISSUE #237: Allow user to Choose the Issue Type with Jira Plugin
Currently, the Jira plugin requires that the Issue Type "Task" be used for the Jira plugin integration.

Request is to expose Jira Issue Type as a configurable option on the Jira Plugin so end users can choose which Issue Type they would like the plugin to use.

highb:
~So I haven't had time to test this out on a live system yet, but~ I believe this will fix #237 . ~I'm planning to try~ I pushed the binary out to my plugins server and it is able to create tickets with the issueType I've configured. It builds with `make access-jira` and tests pass when I do `make test`.

### DIFF
--- a/access/jira/bot.go
+++ b/access/jira/bot.go
@@ -26,6 +26,7 @@ const (
 type Bot struct {
 	client      *resty.Client
 	project     string
+	issueType   string
 	clusterName string
 }
 
@@ -126,7 +127,7 @@ func NewBot(conf JIRAConfig) *Bot {
 		}
 		return nil
 	})
-	return &Bot{client: client, project: conf.Project}
+	return &Bot{client: client, project: conf.Project, issueType: conf.IssueType}
 }
 
 func (b *Bot) HealthCheck(ctx context.Context) error {
@@ -206,7 +207,7 @@ func (b *Bot) CreateIssue(ctx context.Context, reqID string, reqData RequestData
 			},
 		},
 		Fields: IssueFieldsInput{
-			Type:        &IssueType{Name: "Task"},
+			Type:        &IssueType{Name: b.issueType},
 			Project:     &Project{Key: b.project},
 			Summary:     fmt.Sprintf("%s requested %s", reqData.User, strings.Join(reqData.Roles, ", ")),
 			Description: description,
--- a/access/jira/config.go
+++ b/access/jira/config.go
@@ -21,10 +21,11 @@ type Config struct {
 }
 
 type JIRAConfig struct {
-	URL      string `toml:"url"`
-	Username string `toml:"username"`
-	APIToken string `toml:"api_token"`
-	Project  string `toml:"project"`
+	URL       string `toml:"url"`
+	Username  string `toml:"username"`
+	APIToken  string `toml:"api_token"`
+	Project   string `toml:"project"`
+	IssueType string `toml:"issue_type"`
 }
 
 const exampleConfig = `# example jira plugin configuration TOML file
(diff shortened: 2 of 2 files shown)


klizhentas (approved):
@marshall-lee please review, config-wise looks good to me.

highb:
Just cross-compiled this, uploaded the binary and tested this out on one of my proxy servers. Here's the config (token omitted):
```
# example jira plugin configuration TOML file
[teleport]
auth_server = "engops-teleport-auth-e346bfd0d74897d5.elb.us-west-2.amazonaws.com:3025"                       # Teleport Auth Server GRPC API address
client_key = "/var/lib/teleport/plugins/jira/auth.key" # Teleport GRPC client secret key
client_crt = "/var/lib/teleport/plugins/jira/auth.crt" # Teleport GRPC client certificate
root_cas = "/var/lib/teleport/plugins/jira/auth.cas"   # Teleport cluster CA certs

[jira]
url = "https://opengovinc.atlassian.net"    # JIRA URL. For JIRA Cloud, https://[my-jira].atlassian.net
username = "jono.spiro@opengov.com"               # JIRA username
api_token = ""                 # JIRA API Basic Auth token, or our password in case you're using Jira Server.
project = "ONCALL"                  # JIRA Project key
issue_type = "Teleport Access Request"                 # JIRA Issue type

[http]
public_addr = "teleport.production.engops.opengov.zone" # URL on which callback server is accessible externally, e.g. [https://]teleport-proxy.example.com
# listen_addr = ":8081" # Network address in format [addr]:port on which callback server listens, e.g. 0.0.0.0:8081
https_key_file = "/var/lib/teleport/webproxy_key.pem"  # TLS private key
https_cert_file = "/var/lib/teleport/webproxy_cert.pem" # TLS certificate

[log]
output = "stderr" # Logger output. Could be "stdout", "stderr" or "/var/lib/teleport/jira.log"
```

Here's it running and opening up the custom issueType with no problems that I can see...
```
[root@ip-10-251-1-121 ec2-user]# ./teleport-jira start
INFO   Starting Teleport Access JIRAbot 6.1.1:teleport-jira-v6.1.1-6-g27a05a74-dirty jira/app.go:71
INFO   JIRA Issue created jira_issue_id:112592 jira_issue_key:ONCALL-3 request_id:ccba5f69-bd05-4841-872d-83fb4b652e24 request_op:put jira/app.go:333
```

And here's the ticket (I'm mousing over the ticket icon to show the type)...
image

We don't have the workflow configured to send back Jira events to the webhook yet, but this initial pass looks promising!

marshall-lee (approved):

highb:
After further testing, I've found that my changes accomplish the following:
1. When a user makes an access request, a new ticket it opened on our Service Project
2. When an access request expires, the plugin will do the right thing and transition the issue to "Expired".

However, the following parts of the workflow fail:
1. Moving the ticket to Approved or Denied will not propagate back to Teleport. By adding some extra debug info to the plugin, I was able to verify that Jira was delivering the payload. Looking at the structure of the JSON payload provided, it doesn't match what the `Webhook` object is expecting, especially because it lacks the `webHookevent` and `issue_event_type_name` fields that the `onJIRAWebhook` looks for.

highb:
This could also be due to me using the "Automation" feature provided by the Service Desk, rather than the actual Webhook implementation, which I suspect is locked away from me due to my lack of privileges on Jira. I'll check in with our resident Jira wizard and see what I can get going next week.

highb:
Final follow up before I duck out for the weekend; I think I need to get my Jira wizard to setup the webhook at the whole-org level, rather than my team-project level automation. I'll see where I get with that next week and report back.

highb:
Okay, I got my Jira wizard to help me out and confirmed the plugin:
- creates tickets when a request is made
- approves/denies access when the ticket is moved to the relevant status (the webhook sends correctly when we setup the JIra-wide webhook instead of the service desk Automation)

So I think this is good to go.

highb:
Any ideas on the timeline for reviewing this PR?

klizhentas:
@highb sorry about the delay. I will work on getting this merged today.

klizhentas:
merged in https://github.com/gravitational/teleport-plugins/pull/278

PR #220, opened 2021-04-22
Pagerduty plugin migration to new API client and support of new Teleport features

marshall-lee:
- Render resolve reason in the notification incident.
 - Render access reviews as a comments to notification incident.
 - Configure target services using request annotations.
 - Manage plugin data state safely using compare-and-swap.
 - Migrate tests to testify.

Closes #219.
Part of #209.

PR #241, opened 2021-05-21
Support resolve reason & review comments in Slack & Mattermost plugins

### LINKED ISSUE #184: Add Access Workflow Notes to Slack notifications
Improve the notifications content. This does not depend on #183 and can be shipped independently.

- [ ] Add the request note to the body of the Slack notification

Access Workflows added a way for a user to provide a note about why they need the roles they requested. We need to show that note in the notifications. 

- [ ] Add response note to the body of the Slack notification

If the request was processed outside of Slack and Slack received a notification about that, the notification message should contain the response reason note, and a list of accepted / denied roles.

- [ ] Update documentation

marshall-lee:
- Render resolve reason.
 - Post message replies when new access review submitted.
 - Manage plugin data state with compare-and-swap.
 - Migrate tests to testify.
 - Extract common testing.Suite behavior.
 - Limit the events queue concurrency and order the processing by resource

Closes #184.


access/mattermost/app.go
```diff
@@ -325,57 +412,115 @@ func (a *App) getPostRecipients(ctx context.Context, suggestedReviewers []string
...
+func (a *App) setNewPluginData(ctx context.Context, reqID string, data PluginData) error {
+	expect := make(map[string]string)
+	return a.apiClient.UpdatePluginData(ctx, types.PluginDataUpdateParams{
+		Kind:     types.KindAccessRequest,
+		Resource: reqID,
+		Plugin:   pluginName,
+		Set:      EncodePluginData(data),
+		Expect:   expect,
+	})
 }
```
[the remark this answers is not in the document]
marshall-lee: Ah I hadn't update the code yet. After gravitational/teleport#695 I fixed only #220 (pagerduty plugin has the same problem) but forgot to apply the change to slack & mattermost :(

Now I fixed `EncodePluginData` in both slack & mattermost plugins so that if `PluginData` is an empty struct it yields a non-empty map but with all values empty.

For example, before the fix `EncodePluginData` was encoding `0` as a `"0"` now it encodes `0` as `""`

lib/watcherjob.go
```diff
@@ -73,20 +104,92 @@ func (job *watcherJob) eventLoop(ctx context.Context) error {
...
+		if err := ctx.Err(); err != nil {
+			return trace.Wrap(err)
+		}
+
+		job.queueMu.Lock()
+		if queue, exist = job.queues[queueKey]; !exist {
+			if len(job.queues) >= job.config.MaxConcurrency {
+				job.queueMu.Unlock()
+				continue
+			}
```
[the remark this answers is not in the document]
marshall-lee: > Checking for the desired condition in a tight loop like this is bad for performance. 

Yes that definitely must be improved.

> Side note: why do we even need separate queues? Looks to me like all events end up being proceeded by the same function anyhow. Why not just use a single large queue that buffers all events?

@fspmarshall 

That was a temporary workaround against the `UpdatePluginData()` race. My concern was the following:

- If i'm processing events sequentially, it removes the race (at least for the case when there's only one plugin process).
- But processing events sequentially is slower than processing them concurrently.
- So why not have a mixed approach:
  - Events for different access requests: process them concurrently.
  - Events for the same access request: process sequentially to avoid races.

I added tests for both cases explaining the approach.

`UpdatePluginData` races in 99.9% cases are running around the same request so why not have a treatment for this?

As I said, that workaround was *temporary* and now it seems not to be required BUT I tend to keep it because it prevents potential race conditions just not allowing them to happen at least in the context of one process.

lib/watcherjob/watcherjob.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	defer func() {
+		if err := watcher.Close(); err != nil {
+			logger.Get(ctx).WithError(err).Error("Failed to close a watcher")
+		}
+	}()
+
+	if err := job.waitInit(ctx, watcher, 5*time.Second); err != nil {
+		return trace.Wrap(err)
+	}
```
fspmarshall, line 117: Nit: 5 seconds is a bit short.  Init events aren't always sent immediately.  E.g. if the teleport instance is using a cache and the watcher is created during a cache reset, then the init event is not sent until after the cache reset is complete.  Something like a 30 second timeout is probably better.

fspmarshall (approved):
LGTM overall.  I do still have misgivings about the queue system though.  It strikes me as something that will be difficult to debug.  I'd prefer if we could get rid of it entirely, but I understand that going full synchronous or full asynchronous both have their own potential downsides.

How do you feel about simplifying the queueing logic a bit?  IMO something like this would be a bit easier to reason about and should achieve the same desired effect:

```golang
func ProcessEvents(ctx context.Context, ch chan Event, fn EventFunc) {
	const maxEvents = 1024
	
	queues := make(map[eventQueueKey][]Event)
	donech := make(chan eventQueueKey)
	
	handleEvent := func(e Event) {
		fn(ctx, e)
		select {
		case donech <- newQueueKey(e):
		case <-ctx.Done():
		}
	}
	
	var events int
	
	for {
		eventch := ch
		if events >= maxEvents {
			// exert backpressure
			eventch = nil
		}
		select {
		case event := <-eventch:
			events++
			key := newQueueKey(event)
			queue, ok := queues[key]
			if ok {
				queues[key] = append(queue,event)
				continue
			}
			
			go handleEvent(event)
			// create empty queue
			queues[key] = []Event{}
			
		case key := <-donech:
			events--
			queue := queues[key]
			if len(queue) == 0 {
				delete(queues,key)
				continue
			}
			
			next, rest := queue[0], queue[1:]
			go handleEvent(next)
			queues[key] = rest
		case <-ctx.Done():
			return
		}
	}
}
```

lib/watcherjob/watcherjob.go
marshall-lee: @fspmarshall 

If we haven't get `OpInit` we don't proceed further to process the events.

Is it event possible to get `OpPut`, `OpPut`, ... `OpInit` event sequence? Then do we really need to wait for `OpInit` at all?

fspmarshall: Sorry, my wording was ambiguous.  When I say that `OpInit` isn't always sent immediately, I don't mean that it might arrive after other events.  It is always the first event.  I just mean that sometimes there is an extra delay between the time when the watcher is created, and the time when `OpInit` is sent.

In theory, it would be possible to go straight to the event processing loop, and simply ignore the `OpInit` event when it comes through.  Why we don't typically do this is because teleport instances which are unhealthy will queue up watchers, and then sent the `OpInit` event when the instance become healthy.  By enforcing a timeout on how long we're willing to wait for `OpInit`, we're limiting the amount of time that we wait for the instance to become healthy.  Since clients typically interact with teleport via a load balancer, having a reasonable timeout for `OpInit` allows to to attempt to "switch" to a healthy instance if the one we are currently connected to is taking too long to become healthy.

Joerger (approved):

access/slack/bot.go
```diff
@@ -18,6 +20,20 @@ import (
...
+func init() {
+	var err error
+	reviewReplyTemplate, err = template.New("review reply").Parse(
+		`{{.Author}} reviewed the request at {{.Created.Format .TimeFormat}}.
+Resolution: {{.ProposedStateEmoji}} {{.ProposedState}}.
+{{if .Reason}}Reason: {{.Reason}}.{{end}}`,
+	)
+	if err != nil {
+		panic(err)
+	}
```
fspmarshall, line 34: Nit: Use the provided [`template.Must(...)`](https://golang.org/pkg/text/template/#Must) helper.

fspmarshall (approved):
Queue rework looks good to me.

klizhentas:
closed in favor of rebased and signed-off https://github.com/gravitational/teleport-plugins/pull/258

PR #242, opened 2021-05-25
Release 6.2.0

webvictim:
Bumps supported plugin versions to 6.2.0.

~I ran `go get -d -v github.com/gravitational/teleport@v6.2.0 && go mod vendor` as it seemed the correct thing to do, no idea whether that's what we actually need.~ This turned out to be a bad idea because it breaks `go mod`'s versioning expectations. As such, we're just going to leave the imports as-is for now and bump the plugin versions.


vendor/github.com/BurntSushi/toml/session.vim
```diff
@@ -0,0 +1 @@
+au BufWritePost *.go silent!make tags > /dev/null 2>&1
```
russjones, line 1: Is this part of the toml package? Seems like an accidentally commit.

webvictim: I certainly didn't write it, seems like maybe they committed it accidentally.

https://github.com/BurntSushi/toml/blob/master/session.vim

russjones (approved):
Bot.

Joerger:
> I ran `go get -d -v github.com/gravitational/teleport@v6.2.0 && go mod vendor` as it seemed the correct thing to do, no idea whether that's what we actually need.

Info dump: there's no good way to import teleport after 4.3, after which teleport made an incomplete transition to go modules. It looks like the go.mod is still replacing the imported teleport version with `1.3.3`, so this shouldn't matter for now as long as the plugins are working. Eventually we should try to remove teleport dependencies from teleport-plugins by using `github.com/gravitational/teleport/api` or even making new sub modules if needed. I'm doing this for the `access` package for 7.0.

webvictim:
Yeah, everything builds fine with this PR, we just get `incompatible` in the `go.mod` because its version is >2 and it doesn't use tags correctly for versioning. In time we will address this and the imports should get better.

I can just revert the `go.mod` changes.

r0mant (approved):

webvictim:
Holding this temporarily while @klizhentas merges some extra fixes.

PR #245, opened 2021-05-26
Fluentd log forwarder

gzigzigzeo:
This is the initial release of log forwarder. It can forward events to fluentd.


fluentd-forward/README.md
```diff
@@ -0,0 +1,341 @@
...
+```sh
+tctl create -f fluentd-forward.yaml
+```
+
+### Export fluentd-forward Certificate
+
+Teleport Plugin use the fluentd role and user to read the events. We export the identity files, using tctl auth sign.
+
+```sh
+tctl auth sign --format=tls --user=fluentd-forward --out=fd --ttl=720h
```
klizhentas, line 51: will this work with format=file?

fluentd-forward/teleport_client.go
```diff
@@ -0,0 +1,232 @@
...
+		cl, err = newUsingIdentityFile(c)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		cl, err = newUsingKeys(c)
+		if err != nil {
+			return nil, err
+		}
+	}
```
Joerger, line 83: You don't need to switch on identity/keypair credentials, just provide both and `client.New` will use whichever one successfully connects.
```suggestion
	config := client.Config{
		Addrs: []string{c.TeleportAddr},
		Credentials: []client.Credentials{
			client.LoadKeyPair(c.TeleportCert, c.TeleportKey, c.TeleportCA),
			client.LoadIdentityFile(c.TeleportIdentityFile),
		},
	}
	if client, err = client.New(context.Background(), config); err != nil {
		return nil, err
	}
```

fluentd-forward/teleport_client.go
```diff
@@ -0,0 +1,232 @@
...
+		},
+	}
+
+	client, err := client.New(context.Background(), config)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	return client, nil
+}
```
Joerger, line 134: If you go with the above suggestion, you can just remove these helper functions.
```suggestion
```

Joerger (commented):
Just reviewed the API client usage.

fluentd-forward/README.md
gzigzigzeo: Yes, it is shown and described below in Cloud section.

fluentd-forward/teleport_client.go
gzigzigzeo: Great, good to know, will change this piece.

fluentd-forward/config.go
```diff
@@ -0,0 +1,361 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = c.validateStorage()
+	if err != nil {
+		return err
+	}
+
+	// If start time was not passed, use the beginning of time
```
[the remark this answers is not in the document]
gzigzigzeo: Comment was wrong, changed

fluentd-forward/fliuentd_client.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, trace.Wrap(err)
+	}
+	caCertPool := x509.NewCertPool()
+	caCertPool.AppendCertsFromPEM(caCert)
+
+	return caCertPool, nil
+}
+
+// Send sends event to fluentd
+func (f *FluentdClient) Send(obj interface{}) error {
```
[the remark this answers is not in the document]
gzigzigzeo: @xacrimon This method can accept any object which could be marshalled to JSON, so, I just did not want to tightly couple it with specific type.

fluentd-forward/teleport_client.go
```diff
@@ -0,0 +1,260 @@
...
+
+	log.WithFields(log.Fields{"id": t.id, "new_pos": t.pos}).Info("Skipping last known event")
+
+	return nil
+}
+
+// getEvents calls Teleport client and loads events
+func (t *TeleportClient) getEvents() ([]events.AuditEvent, string, error) {
+	return t.client.SearchEvents(
+		context.Background(),
```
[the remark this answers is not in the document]
gzigzigzeo: Added context to TeleportClient, just in case.

klizhentas:
@xacrimon @Joerger @r0mant @fspmarshall this is ready for another round of reviews, thanks team!

fluentd-forward/README.md
```diff
@@ -0,0 +1,355 @@
...
+key  = "fd.key"
+```
+
+## Setup fluentd
+
+For the purpose of testing, we will run the local fluentd instance first.
+
+### <a name="mtls"></a>Generate mTLS certificates
+
+For the purpose of security, we require mTLS to be enabled on the fluentd side. You are going to need [OpenSSL configuration file](example/ssl.conf). Put the following contents to `ssl.conf`:
```
klizhentas, line 117: This is going to be a slow down for getting started, can you please create a fluentd-forward configure command that works similar to:

https://goteleport.com/docs/application-access/getting-started/#step-23-install-and-configure-teleport

and generates all certificates and files for easy to get started? Let's aim for one liner that sets everything up for a local config. We can have advanced section that will show how to generate certificates, etc

gzigzigzeo: OK. Actually, there is one: `make gen-example-mtls`. It might be enough when plugin is built from source, but might not for binary version. I'll try to add the equivalent to binary itself (or possibly to separate bash script if that's okay).

klizhentas: Let's keep everything in one binary - that's what people will be downloading, you can generate certs using go much more conveniently than using external tools, see for example teleport code:

https://github.com/gravitational/teleport/blob/8922af0c6191af8fdf251acfb80f5e8789b07cd0/lib/utils/tls.go#L93

xacrimon (approved):

fluentd-forward/README.md
gzigzigzeo: This is done. mTLS certs are not self-signed, they require CA and separate client/server certs to be generated as well.

fspmarshall (approved):

klizhentas (changes requested):
Almost there. Great work. I was able to make it work in 10 minutes!
Let's bring it to 2 minutes and improve security by applying a couple of tweaks:

* [x] Rename plugin to `teleport-fluentd-forward` to match the naming scheme of all other plugins

* [x] Let's rename `teleport-fluentd-forward gen-certs` into more generic `teleport-fluentd-forward configure` to match `teleport configure`

* [x] `teleport-fluentd-forward configure` to generate password for the private key.

`teleport-fluentd-forward gen-certs` without password argument should generate a crypto random hex 32 byte password and save to `password.file`.

For people would like to generate their own pass read the password from stdin.

As a security precaution we should not accept passwords as a positional arguments to prevent them staying in the command line history.

`teleport fluentd-forward configure . --password < cat password`

* [x] Reduce the amount of steps and generate the `toml` file and role and user from your doc example.
* [x] Use new, `V4` role type in the spec, it is more secure.
* [x] Generate a sample fluent.conf.
* [x] If there a way to for fluent.conf to read the password from file instead of `private_key_passphrase` hardcode, use the file instead.
* [x] All generated secrets can should have permissions `0066`, not world-readable. Owners of the file should be able to write them.
* [x] Fix the output to remove extra INFO lines (that should be debug and omitted from stdout by default unless --debug is present)

Let's make the output more user-friendly:

```bash
$ teleport-fluentd-forward configure . --proxy=example.teleport.sh:443
[1] Generated ca.crt ca.key client.crt client.key in folder .
[2] Encrypted private key with a password. Stored password in password.file
[3] Generated sample fluentd file fluentd-forwarder.toml
[4] Generated sample teleport user and role fluentd-role.yaml
[5] Generated sample fluent.conf

Follow-along with our getting started guide:

https://goteleport.com/setup/guides/fluentd
```

* [ ] Always generate the config file that uses `identity` flow
and update the docs to always use the config designed for:

```bash
$ tctl auth sign --out ./identity --user fluentd-forward
```

This flow will always work for both cloud and on-prem, so let's make it a default.
Move the flow with `--format=tls` in the advanced section.

fluentd-forward/Makefile
```diff
@@ -0,0 +1,72 @@
+VERSION=0.0.1
```
russjones, line 1: Please add a `release` target to the Makefile similar to https://github.com/gravitational/teleport-plugins/pull/248.

gzigzigzeo:
Everything is done, except release target. Fluentd does not support password files. And you possibly meant 0600 perms?

klizhentas:
@fspmarshall @xacrimon @russjones can you please review once again?

klizhentas (changes requested):
Sorry for another rename, but I think it will be worth it

fluentd-forward/.gitignore
```diff
@@ -0,0 +1,5 @@
+build
+example/keys
+example/storage
+cover.out
+tmp
```
xacrimon, line 5: nit: missing an EOF newline

fluentd-forward/state.go
```diff
@@ -0,0 +1,198 @@
...
+	"time"
+
+	"github.com/gravitational/trace"
+	"github.com/peterbourgon/diskv"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	// cacheSizeMax max memory cache
+	cacheSizeMax = 1024
```
xacrimon, line 31: This name should include a unit

fluentd-forward/README.md
```diff
@@ -0,0 +1,350 @@
...
+```
+
+In sections below we assume that you have `teleport-fluentd-forward` executable available in your `$PATH` or cwd.
+
+## Generate example configuration
+
+Run:
+
+```sh
+teleport-fluentd-forward configure .
```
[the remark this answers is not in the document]
gzigzigzeo: Done. Mock arg for now. It might require different cli args for different targets in future, so, possibly it would be better to rename `configure` to `configure-fluentd`?

fluentd-forward/state.go
gzigzigzeo: Renamed to `cacheSizeMaxBytes`. I think, bytes might be assumed by default.

xacrimon: Should be named `cacheSizeMaxKiloBytes` I think. From what I gather from the docs it assumes kilobytes.

gzigzigzeo: I might be mistaken, but according to docs looks like it's bytes: `CacheSizeMax uint64 // bytes`

klizhentas:
closed in favor of buddy signed-off commit https://github.com/gravitational/teleport-plugins/pull/262

Issue #251, opened 2021-06-02
Plugin config example missing recipients field in config.go


inertial-frame:
The Slack Plugin is working great. However, the following config example lacks a `recipients` field:

https://github.com/gravitational/teleport-plugins/blob/a89a239e053414b4a085dd66a80b9f93407276d8/access/slack/config.go#L24-L37

WIthout this field added to the config I get the following after running:

```bash
sudo tctl request create test-user --roles=admin 3daf04b3-7780-4cff-a1ce-345c0318776f
```

```bash
No channel to post request_id:34cac0ad-580e-4b37-9a5c-85e7d90da7e7 request_op:put slack/app.go:209
```

Furthermore, the example should more helpfully indicate:

1. That the recipient format for users should be the email address.
2. That the recipient format for a channel should be the channel ID.

```
recipients=["person@place.com","AIBOEFJD"]
```

inertial-frame:
Verified this for 6.2.3:

```
WARN   No channel to post request_id:ab1903b8-98db-456d-89d6-71fa3418f1ae request_op:put slack/app.go:209
```

The TOML is also updated here: https://github.com/gravitational/teleport/pull/7041

inertial-frame:
Updated this to include `suggested_reviewers` as well.

berdee:
fixed in #259

PR #259, opened 2021-06-16
docs: update slack config example

inertial-frame:
Tweaked the config.go example:

1. The plugin does not require the `recipients` field to launch. But it does need it when you request a role.
2. I also updated the field to better depict what the values need to be per https://github.com/gravitational/teleport/pull/7041.

This is for issue: https://github.com/gravitational/teleport-plugins/issues/251


benarent (approved):

klizhentas (approved):

access/slack/config.go
```diff
@@ -22,14 +22,21 @@ type SlackConfig struct {
 }
 
 const exampleConfig = `# example slack plugin configuration TOML file
```
inertial-frame, line 24: This is the exact string that's used to generate the example configuration file when running: `teleport-slack configure > teleport-slack.toml` or `teleport-slack configure > /path/to/teleport-slack.toml`. Tested this. 

Decided to include the commands and previous configuration.

inertial-frame:
I made a few more changes here and wanted to get another review pass. Thanks and sorry for the extra review!

klizhentas (approved):

access/slack/INSTALL.md
```diff
@@ -151,15 +151,22 @@ certificates there and make sure the config points to them.
...
+
+# tctl auth sign --format=tls --auth-server=auth.example.com:3025 --user=access-plugin --out=auth --ttl=1h
+# client_key = "/var/lib/teleport/plugins/slack/auth.key" # Teleport GRPC client secret key ("tls" format")
+# client_crt = "/var/lib/teleport/plugins/slack/auth.crt" # Teleport GRPC client certificate ("tls" format")
+# root_cas = "/var/lib/teleport/plugins/slack/auth.cas"   # Teleport cluster CA certs ("tls" format")
 
 [slack]
-token = "api-token"       # Slack Bot OAuth token
+token = "xoxb-11xx"                                # Slack Bot OAuth token
+recipients = ["person@email.com","YYYYYYY"]        # Optional Slack Rooms
```
marshall-lee, line 169: `recipients` list is the old mode where recipients are set statically in config file.

Normally, people should specify "suggested reviewers" when they request the role. If any of the suggested reviewers match the slack user emails — that users will be notified.

marshall-lee: This is why `recipients` is not a required option.

inertial-frame: Ah, I was unaware that's the recommended field to configure. Will update, thanks!

PR #260, opened 2021-06-19
Added fluentd build and publish pipeline.

russjones:
**Description**

Added fluentd build and publish pipeline.

**Related Issues**

https://github.com/gravitational/teleport/issues/5350

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -160,7 +160,11 @@ trigger:
     - tag
   ref:
     include:
-      - refs/tags/teleport-*-v*
+      - refs/tags/teleport-gitlab-v*
+      - refs/tags/teleport-jira-v*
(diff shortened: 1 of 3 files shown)


russjones:
@webvictim If I wanted to test this pipeline, the way to do it is to create a tag off `rjones/fluentd-pipeline` and make sure it builds artifacts correctly on Drone right?

webvictim:
@russjones Yes, although obviously you’ll need to update the fluentd Makefile to add the release target too. It’d probably be good to also update the `build-all` target in the root Makefile to do test builds of the fluentd forwarder on push. I think the tests should be covered by the `test` target there already.

klizhentas (approved):

PR #263, opened 2021-06-28
Adds AuthPreference resource, upgrades Role to V4

gzigzigzeo:
- Adds AuthPreference resource
- Replaces RoleV3 with V4


klizhentas:
@fspmarshall @Joerger can you folks please review?

klizhentas (approved):

klizhentas:
@gzigzigzeo @berdee have you tested with the cloud? can you send a PR to add a new resource for terraform reference doc:

https://github.com/gravitational/teleport/blob/master/docs/pages/setup/reference/terraform-provider.mdx

terraform/gen_teleport.yaml
```diff
@@ -83,13 +90,13 @@ defaults:
...
 
     # Role
-    "types.RoleV3.Version": "v3"
-    "types.RoleV3.Kind": "role"
-    "types.RoleV3.SubKind": ""
-    "types.RoleV3.Spec.Options.MaxSessionTTL": "30h"
-    "types.RoleV3.Spec.Options.CertificateFormat": "standard"
-    "types.RoleV3.Spec.Allow.ReviewRequests.Where": ""
-    "types.RoleV3.Spec.Deny.ReviewRequests.Where": ""
+    "types.RoleV4.Version": "v4"
```
fspmarshall, line 93: IIUC this will cause existing terraform that used to emit a `v3` role to now emit a `v4` role if the user did not specify the version.  On its own, this would be a silently breaking change, since there is no change to role syntax, but behavior changes between `v3` and `v4`.  If anyone other than us is using this in prod already, the default value should probably be left as `v3` (@klizhentas may be able to comment on this).

fspmarshall (approved):
LGTM once the role versioning strategy is decided upon.

berdee:
> @gzigzigzeo @berdee have you tested with the cloud? can you send a PR to add a new resource for terraform reference doc:
> 
> https://github.com/gravitational/teleport/blob/master/docs/pages/setup/reference/terraform-provider.mdx

I didn't, but @gzigzigzeo can tell you more about his activities in this PR.

PR #266, opened 2021-06-30
pagerduty: Migrate to 6.2 client & rewrite the logic.

klizhentas:
- Render resolve reason in the notification incident.
 - Render access reviews as a comments to notification incident.
 - Configure target services using request annotations.
 - Manage plugin data state safely using compare-and-swap.
 - Migrate tests to testify.


russjones (approved):
Bot.

PR #294, opened 2021-08-23
Rewrite integration tests to run Teleport as a separate instance.

marshall-lee:
This PR introduces new `github.com/gravitational/teleport-plugins/lib/testing/integration` package useful for writing integration tests for plugins. It initializes a temporary data directory and runs a Teleport binary against it. Auth server is running as a separate OS process which the test suite can communicate with via the API.

Previously we used quite a different approach: teleport was running embedded inside a `go test` so it required an entire Teleport to be compiled along with plugin code and its tests. Similar approach is used in the integration tests of the Teleport itself and it's totally fair there because while testing the Teleport itself you have to compile it anyway.

However, Teleport's approach for writing integration tests is not all good for plugins. The multitude of Teleport-specific dependencies burdens the build and even more we sometimes get some errors when importing new version of Teleport to the teleport-plugins project. Teleport is a large executable, not a library - it's just badly suited to be used *as a module*.

Also I think not depending on hard-vendored Teleport distribution makes it easy to run the tests against different Teleport versions. And not embedding the Teleport process into the plugin test process makes things a little more fair: in reality Teleport and its plugins don't share the memory space, right? :)

In the past we just didn't have another choice. We had to import the entire Teleport anyway just to get its latest generated GRPC client. But now we have a separate `teleport/api` module which implements the client.

So now we don't have a reason to import `github.com/gravitational/teleport` as a dependency anymore. It's not needed in the plugins code and the only place depending on this entire module was tests. This commit fixes it.

Many deps were removed, project is more lightweight now as it should be!

However, there're some **issues with the approach**. Some tests involving enterprise features such as access reviews are not testable anymore with the Teleport OSS binary since it doesn't have them but the Teleport Enterprise binary requires a license file to work. In other words, at least this approach requires **providing a license file on CI**.

So for project contributors now it depends. OSS users are still able to contribute but the tests involving enterprise features will be skipped while running the test suite. It's not critical if having the license CI will run full test suite anyway. Contributors with Enterprise license won't see any difference.


Joerger (commented):
LGTM overall, but it looks like CI is currently failing, @marshall-lee could you look into it?

marshall-lee:
@Joerger it's intended that CI fails because now the test suite requires both teleport binary and license.

Joerger:
@marshall-lee We should find a way to get CI working with the new approach. @russjones What do you think, how can we provide CI with a licence and the required binaries? Or do we need to take a different approach altogether?

tcsc:
I'm looking into how we can inject a license via drone secrets, so we can at least answer the technical side of this question.

Joerger (approved):

tcsc (approved):

PR #297, opened 2021-08-23
Add auth preference and update v3 to v4 role

klizhentas:


russjones (approved):
Bot.

r0mant (commented):
It looks like this PR has same many of the same parts as https://github.com/gravitational/teleport-plugins/pull/298 - are both needed?

gzigzigzeo:
> It looks like this PR has same many of the same parts as #298 - are both needed?

Yes, they are. #298 is based on this PR as like as a couple of upcoming.

PR #298, opened 2021-08-23
Adds aws_role_arns to role, updates schema to latest version

klizhentas:


russjones (approved):
Bot.

terraform/go.mod
```diff
@@ -4,6 +4,7 @@ go 1.15
 
 replace (
 	github.com/coreos/go-oidc => github.com/gravitational/go-oidc v0.0.3
+	github.com/gravitational/teleport/api => github.com/gravitational/teleport/api v0.0.0-20210817145431-a8c11241284b
```
r0mant, line 7: Curious why do we need this replace clause? The source is the same?

terraform/tfschema/types_terraform.go
```diff
@@ -52,10 +56,10 @@ var (
 	SchemaProvisionTokenV2 = GenSchemaProvisionTokenV2()
 	// SchemaMetaProvisionTokenV2 is schema metadata for ProvisionTokenV2 specifies provisioning token
 	SchemaMetaProvisionTokenV2 = GenSchemaMetaProvisionTokenV2()
-	// SchemaRoleV3 is schema for RoleV3 represents role resource specification
-	SchemaRoleV3 = GenSchemaRoleV3()
```
r0mant, line 56: Is this a backwards compatible change? This looks like we're removing V3 role from terraform spec completely unless I'm missing something. What happens if somebody is still using V3 roles?

r0mant (commented):
Question about backwards compatibility with removal of RoleV3, otherwise lgtm.

terraform/tfschema/types_terraform.go
klizhentas: @gzigzigzeo see the question above? both on api and V3 change

gzigzigzeo: @klizhentas @r0mant They will be automatically upgraded to v4. I've tested this. There's no way I can use both because API has no RoleV3 struct anymore, otherwise I would keep both.

terraform/go.mod
gzigzigzeo: It's an artifact :-( Needs to be removed, but I am not quite sure how. Should I update my original branch?

PR #302, opened 2021-08-26
Fixes SAML and Github connectors schema

gzigzigzeo:
What's inside:

- [x] Changed object constructors logic
- [x] Marked required SAML attributes as computed


Joerger (approved):

terraform/example/oidc_connector.tf
```diff
@@ -0,0 +1,19 @@
+# Teleport OIDC connector
```
tcsc, line 1: Why commented out?

terraform/example/trusted_cluster.tf
```diff
@@ -0,0 +1,20 @@
+# Teleport trusted cluster
```
tcsc, line 1: Also why commented out?

tcsc (approved):

terraform/example/oidc_connector.tf
gzigzigzeo: For testing purposes, uncommented it back

gzigzigzeo:
This conflict is about `protoc-gen-terraform` version. In this branch, it points to a correct version.

PR #308, opened 2021-09-02
[Buddy PR] Load identity file from env var

Joerger:
Closes https://github.com/gravitational/teleport-plugins/pull/299 (original PR)


russjones:
@gzigzigzeo @Joerger Is it intentional that master of Teleport is being vendored? I would expect it to be `branch/v7`?

gzigzigzeo:
@russjones It points to the #8132, otherwise would not work. This reference must be updated once #8132 gets back ported to v7.

russjones:
@Joerger Is this something that can be now be merged?

Joerger:
> @Joerger Is this something that can be now be merged since `teleport/lib` is not longer being imported?

This can be merged as soon as https://github.com/gravitational/teleport/pull/8186 is merged.

klizhentas (approved):

PR #295, opened 2021-08-23
gitlab: Rewrite plugin to use new API client.

marshall-lee:
- Use new `teleport/api` client.
- Add comments to the issue when new access review is submitted.
- Automatically change issue status once request is approved/denied.
- Manage plugin data state safely using compare-and-swap.
- Migrate tests to testify.
- Also update slack, mattermost and pagerduty plugins with a new `addr` config setting instead of deprecated `auth_server`.
- Add exponential backoff to `modifyPluginData` loop in slack, mattermost and pagerduty plugins.

Closes #188.
Closes #209.

PR #310, opened 2021-09-03
Teleport Email Plugin

gzigzigzeo:


access/email/bot.go
```diff
@@ -0,0 +1,303 @@
...
+
+var reviewReplyTemplate = template.Must(template.New("review reply").Parse(
+	`{{.Author}} reviewed the request at {{.Created.Format .TimeFormat}}.
+Resolution: {{.ProposedStateEmoji}} {{.ProposedState}}.
+{{if .Reason}}Reason: {{.Reason}}.{{end}}`,
+))
+
+// Bot is a email client that works with access.Request. Essentially it's not a bot, but let's
+// keep naming for the sake of uniformity.
+type Bot struct {
```
marshall-lee, line 32: Please don't call it `Bot`. It's unfortunate naming I'm trying to get rid of.

In pagerduty, jira and gitlab I already renamed it from `Bot` to `Client`.
In slack & mattermost it will probably remain a `Bot` but only because slack clients are actually bots :)
In your case I don't know what name is good, it's up to you to decide.

access/email/config.go
```diff
@@ -0,0 +1,134 @@
...
+[teleport]
+auth_server = "0.0.0.0:3025"                              # Teleport Auth Server GRPC API address
+
+# tctl auth sign --format=file --auth-server=auth.example.com:3025 --user=access-plugin --out=auth --ttl=1h
+identity = "/var/lib/teleport/plugins/email/auth"         # Teleport certificate ("file" format)
+
+# tctl auth sign --format=tls --auth-server=auth.example.com:3025 --user=access-plugin --out=auth --ttl=1h
+# client_key = "/var/lib/teleport/plugins/email/auth.key" # Teleport GRPC client secret key ("tls" format")
+# client_crt = "/var/lib/teleport/plugins/email/auth.crt" # Teleport GRPC client certificate ("tls" format")
+# root_cas = "/var/lib/teleport/plugins/email/auth.cas"   # Teleport cluster CA certs ("tls" format")
```
marshall-lee, line 52: From #295.

```suggestion
[teleport]
# Teleport Auth/Proxy Server address.
#
# Should be port 3025 for Auth Server and 3080 or 443 for Proxy.
# For Teleport Cloud, should be in the form "your-account.teleport.sh:443".
addr = "example.com:3025"

# Credentials.
#
# When using --format=file:
# identity = "/var/lib/teleport/plugins/email/auth_id"    # Identity file
#
# When using --format=tls:
# client_key = "/var/lib/teleport/plugins/email/auth.key" # Teleport TLS secret key
# client_crt = "/var/lib/teleport/plugins/email/auth.crt" # Teleport TLS certificate
# root_cas = "/var/lib/teleport/plugins/email/auth.cas"   # Teleport CA certs

```

access/email/app.go
```diff
@@ -0,0 +1,483 @@
...
+
+	"github.com/gravitational/teleport-plugins/lib"
+	"github.com/gravitational/teleport-plugins/lib/logger"
+	"github.com/gravitational/teleport-plugins/lib/stringset"
+	"github.com/gravitational/teleport-plugins/lib/watcherjob"
+	"github.com/gravitational/teleport/api/client"
+	"github.com/gravitational/teleport/api/client/proto"
+	"github.com/gravitational/teleport/api/types"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/backoff"
```
marshall-lee, line 15: Once #295 or #212 is merged, we'll have our own `github.com/gravitational/teleport-plugins/lib/backoff` package. Lets prepare for this

```suggestion
	grpcbackoff "google.golang.org/grpc/backoff"
```

access/email/app.go
```diff
@@ -0,0 +1,483 @@
...
+	ctx, cancel := context.WithTimeout(ctx, initTimeout)
+	defer cancel()
+
+	var (
+		err          error
+		webProxyAddr string
+		pong         proto.PingResponse
+	)
+
+	bk := backoff.DefaultConfig
```
marshall-lee, line 119:
```suggestion
	bk := grpcbackoff.DefaultConfig
```

access/email/app.go
```diff
@@ -0,0 +1,483 @@
...
+	bk.MaxDelay = backoffMaxDelay
```
marshall-lee, line 120:
```suggestion
	bk.MaxDelay = grpcBackoffMaxDelay
```

access/email/app.go
```diff
@@ -0,0 +1,483 @@
...
+	"github.com/gravitational/trace"
+)
+
+const (
+	// minServerVersion is the minimal teleport version the plugin supports.
+	minServerVersion = "6.1.0-beta.1"
+	// pluginName is used to tag PluginData and as a Delegator in Audit log.
+	pluginName = "email"
+	// backoffMaxDelay is a maximum time GRPC client waits before reconnection attempt.
+	backoffMaxDelay = time.Second * 2
```
marshall-lee, line 26:
```suggestion
	// grpcBackoffMaxDelay is a maximum time GRPC client waits before reconnection attempt.
	grpcBackoffMaxDelay = time.Second * 2
```

access/email/app.go
```diff
@@ -0,0 +1,483 @@
...
+	log.Infof("Marked request as %s and sent emails!", resolution.Tag)
+
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	return nil
+}
+
+// modifyPluginData performs a compare-and-swap update of access request's plugin data.
```
marshall-lee, line 413: From #295 

```suggestion
// modifyPluginData performs a compare-and-swap update of access request's plugin data.
// Callback function parameter is nil if plugin data hasn't been created yet.
// Otherwise, callback function parameter is a pointer to current plugin data contents.
// Callback function return value is an updated plugin data contents plus the boolean flag
// indicating whether it should be written or not.
// Note that callback function fn might be called more than once due to retry mechanism baked in
// so make sure that the function is "pure" i.e. it doesn't interact with the outside world:
// it doesn't perform any sort of I/O operations so even things like Go channels must be avoided.
// Indeed, this limitation is not that ultimate at least if you know what you're doing.
```

access/email/plugindata.go
```diff
@@ -0,0 +1,98 @@
...
+// EmailThread stores value about particular original message
+type EmailThread struct {
+	Email     string
+	MessageID string
+	Timestamp string
+}
+
+// NewRequestData converts types.AccessRequest to RequestData
+func NewRequestData(req types.AccessRequest) (string, RequestData) {
+	id := req.GetName()
```
marshall-lee, line 46: The caller of `NewRequestData` can easily `req.GetName()` too cause it's a trivial getter.
Why we do it here? It seems like an alien thing to `RequestData` construction.

access/email/app.go
```diff
@@ -0,0 +1,483 @@
...
+	dataMaps, err := a.apiClient.GetPluginData(ctx, types.PluginDataFilter{
+		Kind:     types.KindAccessRequest,
+		Resource: reqID,
+		Plugin:   pluginName,
+	})
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	if len(dataMaps) == 0 {
+		return nil, nil
```
marshall-lee, line 453:
```suggestion
		return nil, trace.NotFound("plugin data not found")
```

access/email/app.go
```diff
@@ -0,0 +1,483 @@
...
+	}
+	entry := dataMaps[0].Entries()[pluginName]
+	if entry == nil {
+		return nil, nil
```
marshall-lee, line 457:
```suggestion
		return nil, trace.NotFound("plugin data entry not found")
```

access/email/app.go
```diff
@@ -0,0 +1,483 @@
...
+
+	return nil
+}
+
+// modifyPluginData performs a compare-and-swap update of access request's plugin data.
+func (a *App) modifyPluginData(ctx context.Context, reqID string, fn func(data *PluginData) (PluginData, bool)) (bool, error) {
+	var lastErr error
+	for i := 0; i < maxModifyPluginDataTries; i++ {
+		oldData, err := a.getPluginData(ctx, reqID)
+		if err != nil {
```
marshall-lee, line 418:
```suggestion
		if err != nil && !trace.IsNotFound(err) {
```

access/email/app.go
```diff
@@ -0,0 +1,483 @@
...
+		return trace.Wrap(err)
+	}
+
+	if isNew {
+		if recipients := a.getEmailRecipients(ctx, req.GetSuggestedReviewers()); len(recipients) > 0 {
+			if err := a.sendNewThreads(ctx, recipients, reqID, reqData); err != nil {
+				return trace.Wrap(err)
+			}
+		} else {
+			log.Warning("No recipients to send")
```
marshall-lee, line 233:
```suggestion
			log.Warning("No recipients to send")
			return nil
```

WDYT? In a similar place of slack plugin I don't do it but now i think we can just exit here.

access/email/app.go
```diff
@@ -0,0 +1,483 @@
...
+		}
+		logger.Get(ctx).Infof("New review for request %v by %v is %v", reqID, review.Author, review.ProposedState.String())
+		logSentThreads(ctx, threadsSent)
+	}
+
+	return trace.NewAggregate(errors...)
+}
+
+// sendStatusUpdates updates the messages status and sends message when request has been resolved
+func (a *App) sendStatusUpdates(ctx context.Context, reqID string, resolution Resolution) error {
```
marshall-lee, line 369:
```suggestion
// sendResolution updates the messages status and sends message when request has been resolved
func (a *App) sendResolution(ctx context.Context, reqID string, resolution Resolution) error {
```

marshall-lee (changes requested):

access/email/app.go
```diff
@@ -0,0 +1,483 @@
...
+}
+
+// logSentThreads logs successfuly sent emails
+func logSentThreads(ctx context.Context, threads []EmailThread) {
+	for _, thread := range threads {
+		logger.Get(ctx).WithFields(logger.Fields{
+			"email":      thread.Email,
+			"timestamp":  thread.Timestamp,
+			"message_id": thread.MessageID,
+		}).Info("Successfully sent email!")
```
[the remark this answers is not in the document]
gzigzigzeo: They are sticked to the operations. State change and email delivery are two different operations. First, you'll see that state was change. Next, there would be email deliveries logged.

```INFO   New review by for request 4e82b933-5a7a-437a-bcba-51d50999cd4f by email-plugin is APPROVED request_id:4e82b933-5a7a-437a-bcba-51d50999cd4f request_op:put request_state:PENDING email/app.go:361
INFO   Successfully sent email! email:gzigzigzeo@gmail.com message_id:CE0BDT19OO3K.O8TSWLJJ34GI@gzigzigzeos-MacBook-Air.local request_id:4e82b933-5a7a-437a-bcba-51d50999cd4f request_op:put request_state:PENDING timestamp:2021-09-03 17:19:40.576815 +0300 MSK m=+74.884042960 email/app.go:481```

File: access/email/bot.go [hunk 15]
```diff
@@ -0,0 +1,303 @@
...
+	)
+
+	if webProxyAddr != "" {
+		if webProxyURL, err = lib.AddrToURL(webProxyAddr); err != nil {
+			return Bot{}, trace.Wrap(err)
+		}
+	}
+
+	if conf.Mailgun != nil {
+		mg = mailgun.NewMailgun(conf.Mailgun.Domain, conf.Mailgun.PrivateKey)
```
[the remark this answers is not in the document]
gzigzigzeo: Yep, I had this in mind, thank you.

access/email/bot.go
gzigzigzeo: Yes, you're absolutely correct, `Bot` sounds weird here. I just thought it's a convention. Will happily rename it to Client :-)

access/email/app.go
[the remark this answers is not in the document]
gzigzigzeo: Done

access/email/bot.go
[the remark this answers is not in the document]
gzigzigzeo: Done

access/email/config.go
gzigzigzeo: Done. Will replace auth_server with addr once base PR will be merged.

access/email/app.go
gzigzigzeo: Done

access/email/app.go
gzigzigzeo: Done

access/email/app.go
gzigzigzeo: Done

access/email/app.go
gzigzigzeo: Done

access/email/plugindata.go
gzigzigzeo: Yes. I do not see it as an alien thing. ID belongs to request data, except that for some reason it is not embedded into struct and is always a separate variable.

access/email/app.go
gzigzigzeo: Done

access/email/app.go
gzigzigzeo: Done

access/email/app.go
gzigzigzeo: Agreed.

klizhentas (approved):
U.X. looks good to me, will let the engineering team comment on the code part.

access/email/plugindata_test.go
```diff
@@ -0,0 +1,75 @@
...
+		"user":           "user-foo",
+		"roles":          "role-foo,role-bar",
+		"request_reason": "foo reason",
+		"reviews_count":  "3",
+		"resolution":     "APPROVED",
+		"resolve_reason": "foo ok",
+		"email_threads":  "E1/0000001/M1,E2/0000002/M2",
+	})
+	assert.Equal(t, samplePluginData, pluginData)
+}
```
Joerger, line 62: I think it's worth combining these to reduce hard coded values and to make sure that the roundtrip works as expected. Also more of a nit, but you could just compare the `map[string]string`s directly.
```suggestion
func TestEncodeDecodePluginData(t *testing.T) {
	dataMap := EncodePluginData(samplePluginData)
	assert.Equal(t, map[string]string{
		"user":           "user-foo",
		"roles":          "role-foo,role-bar",
		"request_reason": "foo reason",
		"reviews_count":  "3",
		"resolution":     "APPROVED",
		"resolve_reason": "foo ok",
		"email_threads":  "E1/0000001/M1,E2/0000002/M2",
	}
	pluginData := DecodePluginData(dataMap)
	assert.Equal(t, samplePluginData, pluginData)
}
```'

Side note, is there a reason why you are using `assert` here? In `teleport`, we are transitioning to `testify/require` and it looks like teleport-plugins is also split between the two.

File: access/email/email_test.go [hunk 17]
```diff
@@ -0,0 +1,653 @@
...
+func (s *EmailSuite) TestDenialByReview() {
+	t := s.T()
+
+	if !s.teleportFeatures.AdvancedAccessWorkflows {
+		t.Skip("Doesn't work in OSS version")
+	}
+
+	req := s.createAccessRequest([]string{s.userNames.requestor})
+
+	s.startApp()
```
Joerger, line 445: Should these be swapped in order?

TESTING.md
```diff
@@ -0,0 +1,133 @@
+# Using `github.com/gravitational/teleport-plugins/lib/testing` package for writing tests
+
+This package implements Teleport integration testing.
+
+For the complete examples, please refer to [email plugin tests](access/email/email_test.go).
+
+```go
+import . "github.com/gravitational/teleport-plugins/lib/testing"
```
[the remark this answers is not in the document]
gzigzigzeo: First of all, this doc is a bit outdated already. I am going to update it once we merge this PR or event handler, whatever comes first. I have fixed this in the test itself.

access/email/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Generate the certificate
+
+For the plugin to connect to Auth Server, it needs an identity file containing TLS/SSH certificates. This can be obtained with tctl:
+
+```bash
+$ tctl auth sign --auth-server=AUTH-SERVER:PORT --format=file --user=access-email --out=/var/lib/teleport/plugins/email/auth_id --ttl=8760h
+```
+
+Here, `AUTH-SERVER:PORT` could be `localhost:3025`, `your-in-cluster-auth.example.com:3025`, `your-remote-proxy.example.com:3080` or `your-teleport-cloud.teleport.sh:443`. For non-localhost connections, you might want to pass the `--identity=...` option to authenticate yourself to Auth Server.
```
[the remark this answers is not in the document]
gzigzigzeo: This section is equal in all plugin docs (Slack, Jira), I just used the same wording.

access/email/plugindata.go
```diff
@@ -0,0 +1,111 @@
...
+	result := make(map[string]string)
+
+	result["user"] = data.User
+	result["roles"] = strings.Join(data.Roles, ",")
+	result["request_reason"] = data.RequestReason
+	var reviewsCountStr string
+	if data.ReviewsCount > 0 {
+		reviewsCountStr = fmt.Sprintf("%d", data.ReviewsCount)
+	}
+	result["reviews_count"] = reviewsCountStr
```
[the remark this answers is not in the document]
gzigzigzeo: It is encoded the same way in Slack plugin. Not quite sure, I think, it's better to address this to @marshall-lee

access/email/plugindata.go
```diff
@@ -0,0 +1,111 @@
...
+// DecodePluginData deserializes a string map to PluginData struct.
+func DecodePluginData(dataMap map[string]string) (data PluginData) {
+	data.User = dataMap["user"]
+	if str := dataMap["roles"]; str != "" {
+		data.Roles = strings.Split(str, ",")
+	}
+	data.RequestReason = dataMap["request_reason"]
+	if str := dataMap["reviews_count"]; str != "" {
+		fmt.Sscanf(str, "%d", &data.ReviewsCount)
+	}
```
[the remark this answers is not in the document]
gzigzigzeo: I guess, we do not want to return errors from this method at all, even if plugin data gets corrupted - we still want to send emails. Let's ask @marshall-lee

access/email/plugindata_test.go
gzigzigzeo: 1. Good idea. Combined tests.
2. Yes, it is between the to. I prefer require. `assert` is needed sometimes, in constructs like `next = next && assert.True(t, loaded)`. but there are not a lot of places like these. Let's just agree on require from now on.

access/email/email_test.go
gzigzigzeo: It does not matter, I'll wait for all messages anyway. Swapped just for convenience.

access/email/plugindata.go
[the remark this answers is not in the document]
Joerger: Fair enough, just ignore the error but leave a comment. strconv.Atoi would still be much simpler in that case.

```
data.ReviewsCount, _ = strconv.Atoi(dataMap["reviews_count"])
```

Joerger (approved):

access/email/plugindata.go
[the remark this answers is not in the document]
marshall-lee: Yeah I agree that `strconv.Atoi` suits better

access/email/plugindata.go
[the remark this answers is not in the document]
marshall-lee: `UpdatePluginData` treats empty values as missing so this is important to encode empty values as `""` when data is passed as CAS expectation. Sorry that it's undocumented

access/email/README.md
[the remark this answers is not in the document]
marshall-lee: @Joerger how to `tsh login` as a plugin user that typically doesn't have a password?

access/email/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Save configuration file
+
+By default, configuration file is expected to be at `/etc/teleport-email.toml`.
+
+```toml
+# /etc/teleport-email.toml
+[teleport]
+auth_server = "example.com:3025"                               # Teleport Auth/Proxy/Tunnel Server Address
```
marshall-lee, line 64: `auth_server` setting is deprecated

```suggestion
# Teleport Auth/Proxy Server address.
#
# Should be port 3025 for Auth Server and 3080 or 443 for Proxy.
# For Teleport Cloud, should be in the form of "your-account.teleport.sh:443".
addr = "example.com:3025"
```

access/email/config.go
```diff
@@ -0,0 +1,214 @@
...
+	SMTP     *SMTPConfig        `toml:"smtp"`
+	Delivery DeliveryConfig     `toml:"delivery"`
+	Log      logger.Config      `toml:"log"`
+}
+
+// TODO: Replace auth_server with addr once it is merged
+const exampleConfig = `# Example email plugin configuration TOML file
+
+[teleport]
+auth_server = "0.0.0.0:3025"                              # Teleport Auth Server GRPC API address
```
marshall-lee, line 64:
```suggestion
# Teleport Auth/Proxy Server address.
#
# Should be port 3025 for Auth Server and 3080 or 443 for Proxy.
# For Teleport Cloud, should be in the form "your-account.teleport.sh:443".
addr = "example.com:3025"
```

marshall-lee (changes requested):

access/email/plugindata.go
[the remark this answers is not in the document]
gzigzigzeo: Changed

access/email/app.go
```diff
@@ -0,0 +1,516 @@
...
+// Otherwise, callback function parameter is a pointer to current plugin data contents.
+// Callback function return value is an updated plugin data contents plus the boolean flag
+// indicating whether it should be written or not.
+//
+// Note that callback function fn might be called more than once due to retry mechanism baked in
+// so make sure that the function is "pure" i.e. it doesn't interact with the outside world:
+// it doesn't perform any sort of I/O operations so even things like Go channels must be avoided.
+//
+// Indeed, this limitation is not that ultimate at least if you know what you're doing.
+func (a *App) modifyPluginData(ctx context.Context, reqID string, fn func(data *PluginData) (PluginData, bool)) (bool, error) {
```
tcsc, line 447: Not necessary for _this_ PR, but a version of this seems to exist for each plugin. Is there a way to factor it out into a common package - or are there subtle differences between the plugin versions that I'm not seeing?

tcsc (approved):

PR #318, opened 2021-09-13
Make part of Token sensitive

ReeceCrossland:
Teleport provision token resource has an ID that is the token. The issue is that this should be a sensitive value, we require this to be obscured within tf plans/apply runs.

To tackle this our solution is to split this and have a prefix as the ID instead, so the full token is not exposed.

Since there is no other ID exposed by the Teleport API, our options are limited.


terraform/provider/resource_teleport_provision_token.go
```diff
@@ -52,16 +166,34 @@ func resourceProvisionTokenCreate(ctx context.Context, d *schema.ResourceData, m
...
+	if !ok {
+		return diagFromErr(fmt.Errorf("Index is required"))
+	}
+	name, ok := d.GetOk("name")
+	if !ok {
+		return diagFromErr(fmt.Errorf("Name is required"))
+	}
+
+	token := index.(string) + name.(string)
+	metadata, ok := d.GetOk("metadata")
```
gzigzigzeo, line 179: What would you say if we rephrased this fragment a bit?

```
var m []interface{}

metadata, ok := d.GetOk("metadata")
if !ok {
	m = []interface{}{map[string]interface{}{"name": token}}
} else {
	m, ok := metadata.([]interface{})
	if !ok {
		return diagFromErr(trace.BadParameter("Can not convert %T to []interface{}", a))
	}
	md, ok := m[0].(map[string]interface{})
	if !ok {
		return diagFromErr(trace.BadParameter("Can not convert %T to map[string]interface{}", a))
	}

	md["name"] = token
}

d.Set("metadata", m)
```

terraform/provider/resource_teleport_provision_token.go
```diff
@@ -52,16 +166,34 @@ func resourceProvisionTokenCreate(ctx context.Context, d *schema.ResourceData, m
...
 	}
 
 	// Check if token already exists
-	_, err = c.GetToken(ctx, name)
+	_, err = c.GetToken(ctx, token)
 	if err == nil {
-		existErr := "token " + name + " exists in Teleport. Either remove it (tctl tokens rm " + name + ")" +
-			" or import it to the existing state (terraform import teleport_provision_token." + name + " " + name + ")"
+		existErr := "token " + token + " exists in Teleport. Either remove it (tctl tokens rm " + token + ")" +
+			" or import it to the existing state (terraform import teleport_provision_token." + index.(string) + " " + token + ")"
```
gzigzigzeo, line 196: What happens if `index.(string)` conversion fails?

terraform/provider/resource_teleport_provision_token.go
```diff
@@ -94,11 +226,12 @@ func resourceProvisionTokenCreate(ctx context.Context, d *schema.ResourceData, m
...
 
 	err = c.UpsertToken(ctx, tV2)
+
 	if err != nil {
-		return diagFromErr(describeErr(err, "token"))
+		return diagFromErr(fmt.Errorf("API Error: %v", err))
 	}
 
-	d.SetId(tV2.GetName())
+	d.SetId(index.(string))
```
gzigzigzeo, line 234: Could convert index to string somewhere at the beginning, handle conversion error and use the result in all cases like this?

terraform/provider/resource_teleport_provision_token.go
```diff
@@ -94,11 +226,12 @@ func resourceProvisionTokenCreate(ctx context.Context, d *schema.ResourceData, m
 	}
 
 	err = c.UpsertToken(ctx, tV2)
+
 	if err != nil {
-		return diagFromErr(describeErr(err, "token"))
+		return diagFromErr(fmt.Errorf("API Error: %v", err))
```
gzigzigzeo, line 231: Could you please explain why it's worth omitting `describeErr` here?

gzigzigzeo (changes requested):

Joerger:
This has been fixed in #341, thank you for your contribution.

PR #329, opened 2021-09-27
[Buddy PR] Add Proxy and SSH services support to testing/integration library + CA Pinning fix

Joerger:
This is a buddy PR for [this commit](https://github.com/gravitational/teleport-plugins/pull/329/commits/b721d1e955ce407d1ae065e9b56ec3d16966a8d0) which was included in [this PR](https://github.com/gravitational/teleport-plugins/pull/317).

This PR also adds support for CA-pinning in the integration tests, which fixes an issue with the darwin drone tests caused by insecure teleport process registration. For further clarification, the darwin drone test was trying to `lstat /private/var/lib/teleport/ca.cert`, which fails due to the permissions of the `/private` directory.

### DIFF
--- a/lib/testing/integration/authservice.go
+++ b/lib/testing/integration/authservice.go
@@ -34,11 +34,11 @@ import (
 
 var regexpAuthStarting = regexp.MustCompile(`Auth service [^ ]+ is starting on [^ ]+:(\d+)`)
 
-type AuthServer struct {
+type AuthService struct {
 	mu           sync.Mutex
 	teleportPath string
 	configPath   string
-	authAddr     string
+	authAddr     Addr
 	isReady      bool
 	readyCh      chan struct{}
 	doneCh       chan struct{}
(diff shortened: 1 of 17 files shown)


lib/testing/integration/authservice.go
```diff
@@ -270,6 +290,35 @@ func (auth *AuthService) IsReady() bool {
...
+
+	submatch := regexpAuthCAPin.FindStringSubmatch(string(output))
+	if submatch != nil {
+		auth.mu.Lock()
+		defer auth.mu.Unlock()
+		auth.caPin = submatch[1]
+	}
+	if auth.caPin == "" {
+		return trace.Errorf("failed to find CA Pin in auth status")
+	}
```
marshall-lee, line 318: Can this be a method in `tctl` package?

marshall-lee (approved):

lib/testing/integration/authservice.go
Joerger: Good idea, I moved all of the logic to `integration` and `setup` instead of `authservice`.

tcsc (approved):

PR #332, opened 2021-10-01
Add Kubernetes manager RFD

marshall-lee:


marshall-lee:
Ping @pierrebeaucamp @r0mant

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,100 @@
...
+
+```yaml
+apiVersion: instances.goteleport.com/v7
+kind: Instance
+metadata:
+  name: some-teleport-instance
+  namespace: teleport
+spec:
+  addr: some-teleport.svc.cluster.local:3080
+  secretName: some-creds
```
pierrebeaucamp, line 36: To follow some other k8s conventions, I'd rename this field to `secretRef` instead and make it an object:

```yaml
spec:
  secretRef:
    name: some-creds
    # optional namespace
    namespace: teleport
```

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,100 @@
...
+
+Custom resource definitions of Teleport resource look like this:
+
+```yaml
+apiVersion: resources.goteleport.com/v2
+kind: Role
+metadata:
+  name: folks
+  namespace: teleport
+instanceName: some-instance
```
pierrebeaucamp, line 65: I don't know if you can actually add new top-level fields to a CRD in k8s. 

You can probably achieve the same using [owner references](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/)

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,100 @@
...
+
+Deletion of resources is being handled using [resource finalizers](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/). Every acknowledged resource gets a finalizer named `"resources.goteleport.com/delete"` so when the deletion happens, Kubernetes doesn't actually delete it but sets a deletion timestamp. While the finalizer list is non-empty, the manager can still access the resource's fields, `instanceName` in particular. Deletion of the Kubernetes object leads to deletion of the associated resource in Teleport. Once a deletion in Teleport succeeds, the manager removes a finalizer from the Kubernetes object so it can be garbage-collected.
+
+### API Groups and Versions
+
+`Instance` resource resides in an API group called `instances.goteleport.com`, and all the Teleport resources reside in a group called `resources.goteleport.com`. Two different groups are used for a reason:
+
+- API Versions of `instances.goteleport.com` start from `v7` because the Teleport version is 7.x at the moment of writing this document.
+- API Versions of `resources.goteleport.com` are all different because they're tied to Teleport resource versions.
```
pierrebeaucamp, line 92: For simplicity, I'd probably just the same API group for both - but that's personal preference.

rfd/0001-kubernetes-manager.md
marshall-lee: > I don't know if you can actually add new top-level fields to a CRD in k8s.

I think top-level fields are not disallowed and such setup works fine in my draft implementation. `instanceName` was put on a top-level intentionally. Otherwise, definitions will look like this:

```yaml
metadata:
  name: folks
  namespace: teleport
spec:
  instanceName: some-instance
  spec:
    allow: ...
```

So it creates an unnecessary nesting and also it's confusing when you include `spec` inside `spec`. Okay `spec.spec` could be a `spec.role` but I wanted to achieve such a UX that whenever the user sees `spec` then it means the same thing as the `spec` in Teleport. This mitigates the confusion a lot.

> You can probably achieve the same using owner references

i thought about using owner references but are they supposed to be edited by the user? owner reference needs a resource UUID — where the user supposed to get it? and looks more verbose in YAML.

I always thought that owner references are for the stuff one creates programmatically.

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,100 @@
...
+For the task, the [Kubebuilder](https://kubebuilder.io/) framework should be used. It has great code generation capabilities and gives a complete setup to develop & deploy the manager.
+
+### Instance definitions
+
+To write something into Teleport, the manager needs to know what server to connect to and how. So it needs Teleport Auth/Proxy/Tunnel address and credentials to be recognized. 
+
+Instance definitions look like this:
+
+```yaml
+apiVersion: instances.goteleport.com/v7
```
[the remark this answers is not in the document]
marshall-lee: > You might want to add a version field to the spec describing the Teleport version instead.

I thought about it — not in a context of `Instance` resources but in a context of `resources.goteleport.com` resources and it's problematic.

1. `apiVersion` + `version` creates a confusion for a user.
1. different `version` would probably mean different `spec` schema. CRDs doen't allow to make such a divergence — openapi v3 schema of a CRD is static cause it's used for validation etc

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,100 @@
...
+kind: Instance
```
r0mant, line 30: "Instance" is a bit too generic IMO. If this is intended to represent Auth/Proxy, can it just have a respective Auth/Proxy kind as well?

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Details
+
+### Implementation
+
+For the task, the [Kubebuilder](https://kubebuilder.io/) framework should be used. It has great code generation capabilities and gives a complete setup to develop & deploy the manager.
+
+### Instance definitions
+
+To write something into Teleport, the manager needs to know what server to connect to and how. So it needs Teleport Auth/Proxy/Tunnel address and credentials to be recognized.
```
r0mant, line 24: Can K8s operator run as a part of a Teleport process (e.g. auth server)?

rfd/0001-kubernetes-manager.md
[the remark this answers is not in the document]
r0mant: Version here should be independent of Teleport version I think, otherwise we'd have to bump it for every release which is additional maintenance burden.

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Implementation
+
+For the task, the [Kubebuilder](https://kubebuilder.io/) framework should be used. It has great code generation capabilities and gives a complete setup to develop & deploy the manager.
+
+### Instance definitions
+
+To write something into Teleport, the manager needs to know what server to connect to and how. So it needs Teleport Auth/Proxy/Tunnel address and credentials to be recognized. 
+
+Instance definitions look like this:
```
r0mant, line 26: If this "Instance" object is not created, can K8s operator auto-discover it e.g. using predefined service name (`teleport.<namespace>.svc.cluster.local`) and secret name (`teleport-credentials`)? Would improve UX. Then maybe we can remove the need for Instance completely (or keep it for power users).

rfd/0001-kubernetes-manager.md
r0mant: Yeah ownerReference is mainly for controllers (daemon/replica sets, etc.) to inject their reference into their managed pods.

Why do resources need instanceName reference at all? K8s operator can just query the "instance" resource and then create a teleport resource with it?

rfd/0001-kubernetes-manager.md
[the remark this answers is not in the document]
marshall-lee: Yeah probably we can make it independent. Frankly, I don't have a strong preference here. It just follows a scheme we use in all the `teleport-plugins` stuff. Initially it was proposed by @klizhentas that all the plugin versions should match the current Teleport version no matter what. So the choice to start `instance.goteleport.com` versioning from `v7` follows this scheme.

rfd/0001-kubernetes-manager.md
marshall-lee: @r0mant Yes it can but why burden the Teleport with an extra responsibility? Any plugin can theoretically be a part of auth server but we make plugins anyway.
In programming it's better to divide than not to divide unless some serious concerns raised. Can you advance your reasoning?

marshall-lee: It was decided to run the controller as a part of Teleport deployment i.e. as a sidecar process. So neither instance nor secrets will be needed. Controller would be able to generate credentials for itself.

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Details
+
+### Implementation
+
+For the task, the [Kubebuilder](https://kubebuilder.io/) framework should be used. It has great code generation capabilities and gives a complete setup to develop & deploy the manager.
+
+### Teleport Auth server discovery and authentication
+
+To write something into Teleport, the manager needs to know what server to connect to and how. There're many options on how to discover Teleport instances in the cluster but for now, let's start with the simplest one - run manager as a sidecar container in the same pod as the Teleport auth server. Containers in the pod can share the volumes, so the manager container could have access to `/var/lib/teleport/proc` database, which contains administrator credentials. This is the same way how `tctl` utility works - it doesn't require any authentication when you run it on the same machine where Teleport is running.
```
r0mant, line 24: In HA scenario with multiple auth servers running, each one will have the operator sidecar I presume? How will the operator handle HA and how will multiple operators work together? Might be helpful to explore this a bit.

rfd/0001-kubernetes-manager.md
r0mant: Is this instance name field still needed?

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,105 @@
...
+kind: User
+metadata:
+  name: human-bean
+  namespace: teleport
+instanceName: some-instance
+spec:
+  roles: ["folks"]
+```
+
+Manager process watches for custom resource events. Once a custom resource is created/updated, the manager's watcher acknowledges it and upserts a resource into Teleport. Every custom resource must have the `instanceName` field useful to look up an instance resource in order to know what address to connect to. Instance resource and its associated secret must reside in the same namespace as the custom resource.
```
r0mant, line 52: I think this description may be out of date since we're running as a sidecar now?

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,105 @@
...
+
+```yaml
+apiVersion: control.goteleport.com/v1
+kind: Identity
+metadata:
+  name: access-slack
+spec:
+  username: access-slack
+  secretName: access-slack-secret
+```
```
r0mant, line 70: How is the Identity resource used by the operator? If it's running as a side-car, it should be able to talk to auth server directly?

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,105 @@
...
+
+Here, the `username` is a Teleport user name, and `secretName` is a name of a secret resource where the manager will write credentials. Once the `Identity` object is created in the Kubernetes, the secret object with credentials contents will be written in the same namespace where the identity object resides.
+
+### Limit the scope
+
+When running the manager as a sidecar container, the question arises: what objects is this particular instance of the manager responsible for and what objects not? There could be multiple Teleport deployments in the same cluster, and each deployment might have its manager sidecar container. So the question is how to limit the scope of objects observed.
```
r0mant, line 76: > There could be multiple Teleport deployments in the same cluster

Why would there be multiple Teleport clusters in a single Kubernetes cluster (if I understood correctly this sentence)?

r0mant: Oh this is probably for the Teleport Cloud use-case, correct?

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,105 @@
...
+
+The first option is to limit by the namespace, and the second is to limit by label selectors. Both of them must be present in the manager to be very specific on what Kubernetes objects to watch.
```
r0mant, line 78: I wonder, why not just have the operator monitor resources in the same namespace it's running in? Then there would no be any additional configuration. And it will work for Teleport Cloud use-case as well.

rfd/0001-kubernetes-manager.md
marshall-lee: It's useful to generate the credentials for plugins.
Plugins themselves could run as a sidecar but I don't find it a good idea for security reasons.

rfd/0001-kubernetes-manager.md
marshall-lee: @r0mant Yes but label selectors configuration is optional. If it's missing the selectors won't be applied.

I brought the label selectors to serve the helm chart use case — to explicitely limit the resources the manager handles only to those that were created by Helm. The helm chart uses label selectors everywhere so I just wanted to follow the practice.
Don't you think it's a right idea?

rfd/0001-kubernetes-manager.md
pierrebeaucamp: controllers written with the `kubebuilder` framework (like this one) utilise a locking mechanism. So if two operators claim to control the same resource, only the one holding the lock will receive events (afaik).

There are probably some edge-cases (for example, we don't want that second sidecar to crash-loop), but those can be sorted out I assume.

rfd/0001-kubernetes-manager.md
pierrebeaucamp: Limiting by namespace seems to be straight-forward enough. I don't know if we need the label selectors, to be honest.

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,103 @@
...
+To run some plugin in the cluster, the credentials are required. Normally, one can get the credentials credentials using `tctl auth sign` command or using `tsh login`. To automate the credentials generation, the `Identity` resource is introduced. The identity definition looks like this:
+
+```yaml
+apiVersion: control.goteleport.com/v1
+kind: Identity
+metadata:
+  name: access-slack
+spec:
+  username: access-slack
+  secretName: access-slack-secret
```
pierrebeaucamp, line 67: What I like about this RFD so far is that the CRDs are closely mapped to Teleport resources. That makes using the operator conceptually easy.

However, this resource breaks this pattern. "Identity" in this context is a new concept.

Both `tcl auth sign` and `tsh login` produce certificates under the hood, so I wonder if it might make more sense to call this a `Certificate` instead and replicate the flags of `tcl auth sign` in the `spec` section here.

Something like:

```yaml
apiVersion: resources.goteleport.com/v8
kind: Certificate
metadata:
  name: access-slack
spec:
  user: access-slack
  out:
    secretRef:
      name: access-slack-secret
```

rfd/0001-kubernetes-manager.md
marshall-lee: Yes `controller-runtime` utilizes a built-in Kubernetes leader election used by all the .

rfd/0001-kubernetes-manager.md
marshall-lee: Okay lets stay with just namespaces for now.

rfd/0001-kubernetes-manager.md
marshall-lee: @pierrebeaucamp nice idea to unify the names!

However I cannot agree that `Certificate` is better. It's called an "identity file" in Teleport:

```
tctl auth sign --help
usage: tctl auth sign --out=OUT [<flags>]

Create an identity file(s) for a given user

Flags:
  -d, --debug              Enable verbose logging to stderr
  -c, --config             Path to a configuration file [/etc/teleport.yaml]. Can also be set via the TELEPORT_CONFIG_FILE environment variable.
      --auth-server        Attempts to connect to specific auth/proxy address(es) instead of local auth [127.0.0.1:3025]
  -i, --identity           Path to an identity file. Must be provided to make remote connections to auth. An identity file can be exported with 'tctl auth sign'
      --insecure           When specifying a proxy address in --auth-server, do not verify its TLS certificate. Danger: any data you send can be intercepted or modified by an
                           attacker.
      --user               Teleport user name
      --host               Teleport host name
  -o, --out                identity output
      --format             identity format: "file" (default), "openssh", "tls", "kubernetes", "db" or "mongodb"
      --ttl                TTL (time to live) for the generated certificate
      --compat             OpenSSH compatibility flag
      --proxy              Address of the teleport proxy. When --format is set to "kubernetes", this address will be set as cluster address in the generated kubeconfig file
      --overwrite          Whether to overwrite existing destination files. When not set, user will be prompted before overwriting any existing file.
      --leaf-cluster       Leaf cluster to generate identity file for when --format is set to "kubernetes"
      --kube-cluster-name  Kubernetes cluster to generate identity file for when --format is set to "kubernetes"
```

And in terms of PKI stuff, the identity file it's not just a certificate

pierrebeaucamp: Let's keep calling it `Identity` then - thanks for pointing out that we already use that name in our CLI.

I don't mind keeping the `secretName` instead of `secretRef` either. I find the latter a bit easier to remember, but it's not a big deal really. Good to see that there is already some other CRDs calling it `secretName` as well. 👍

rfd/0001-kubernetes-manager.md
r0mant: Makes sense, let's mention this in the RFD please (add a separate section on HA).

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,103 @@
...
+
+The name of the Kubernetes custom resource is the name of the resource in Teleport. Also, every Kubernetes custom resource must have a `spec` field that has the same schema as a `spec` field of a resource in Teleport.
+
+Deletion of resources is being handled using [resource finalizers](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/). Every acknowledged resource gets a finalizer named `"resources.goteleport.com/delete"` so when the deletion happens, Kubernetes doesn't actually delete it but sets a deletion timestamp. While the finalizer list is non-empty, the manager can still access the resource's fields. Deletion of the Kubernetes object leads to deletion of the associated resource in Teleport. Once a deletion in Teleport succeeds, the manager removes a finalizer from the Kubernetes object so it can be garbage-collected.
+
+### Identity definitions
+
+To run some plugin in the cluster, the credentials are required. Normally, one can get the credentials credentials using `tctl auth sign` command or using `tsh login`. To automate the credentials generation, the `Identity` resource is introduced. The identity definition looks like this:
```
r0mant, line 58:
```suggestion
To run some plugin in the cluster, the credentials are required. Normally, one can get the credentials using `tctl auth sign` command or using `tsh login`. To automate the credentials generation, the `Identity` resource is introduced. The identity definition looks like this:
```

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,103 @@
...
+apiVersion: control.goteleport.com/v1
+kind: Identity
+metadata:
+  name: access-slack
+spec:
+  username: access-slack
+  secretName: access-slack-secret
+```
+
+Here, the `username` is a Teleport user name, and `secretName` is a name of a secret resource where the manager will write credentials. Once the `Identity` object is created in the Kubernetes, the secret object with credentials contents will be written in the same namespace where the identity object resides.
```
r0mant, line 70: Might be helpful to clarify what happens if:

- Identity resource is removed.
- The generated secret is removed (will be be re-created?)

rfd/0001-kubernetes-manager.md
r0mant: Yeah I agree with just using namespaces. Let's please clarify this in the RFD so it's explicit what's going to be implemented and what's out of scope.

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,103 @@
...
+    - key: app
+      operator: In
+      values: ["my-teleport"]
+```
+
+### API Groups and Versions
+
+`Identity` resource resides in an API group called `control.goteleport.com`, and all the Teleport resources reside in a group called `resources.goteleport.com`. Two different groups are used for a reason:
+
+- API Versions of `control.goteleport.com` start from `v1`.
```
r0mant, line 94: Nit: Just curious how you picked the name "control" as it's pretty broad. Wouldn't something like "identity" or "plugins" be a bit more suitable, or do you have other resources in mind that could potentially reside in the same API group?

r0mant (commented):
Don't have much else, just a couple more questions and the ask to bring the RFD up-to-date w.r.t. to all feedback.

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,116 @@
...
+
+Deletion of resources is being handled using [resource finalizers](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/). Every acknowledged resource gets a finalizer named `"resources.goteleport.com/delete"` so when the deletion happens, Kubernetes doesn't actually delete it but sets a deletion timestamp. While the finalizer list is non-empty, the manager can still access the resource's fields. Deletion of the Kubernetes object leads to deletion of the associated resource in Teleport. Once a deletion in Teleport succeeds, the manager removes a finalizer from the Kubernetes object so it can be garbage-collected.
+
+### Identity definitions
+
+To run some plugin in the cluster, the credentials are required. Normally, one can get the credentials using `tctl auth sign` command or using `tsh login`. To automate the credentials generation, the `Identity` resource is introduced. The identity definition looks like this:
+
+```yaml
+apiVersion: control.goteleport.com/v8
```
r0mant, line 65: Should this be updated to `instances.goteleport.com/v8`? Seems like you changed it below.

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,116 @@
...
+```yaml
+identities:
+  ownerRefEnabled: true
+```
+
+When `ownerRefEnabled` is true, the manager sets a [metadata.ownerReferences](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/) of a `Secret` object referenced by `secretName` of some `Identity`. Once an `Identity` resource is removed, Kubernetes will automatically clean up the referenced secret object.
+
+By default, `ownerRefEnabled` should be turned off because if the user suddenly deleted the `Identity` object it can break the deployment.
+
+If the secret referenced by `secretName` has been removed or its secret data has been overwritten, it's the manager's responsibility to re-generate the identity and to write it back.
```
r0mant, line 88: > it's the manager's responsibility to re-generate the identity and to write it back.

Who's the "manager" in this context? Human operator or this K8s operator? Can you clarify this in the doc.

rfd/0001-kubernetes-manager.md
```diff
@@ -0,0 +1,116 @@
...
+Example:
+
+```yaml
+scope:
+  namespace: some-namespace
+```
+
+### API Groups and Versions
+
+`Identity` resource resides in an API group called `control.goteleport.com`, and all the Teleport resources reside in a group called `resources.goteleport.com`. Two different groups are used for a reason:
```
r0mant, line 105:
```suggestion
`Identity` resource resides in an API group called `instances.goteleport.com`, and all the Teleport resources reside in a group called `resources.goteleport.com`. Two different groups are used for a reason:
```

?

r0mant (approved):

marshall-lee:
@pierrebeaucamp @r0mant I know you guys didn't have any objections except making the changes we were all agreed on. Now I did all the discussed changes so if nothing else comes to mind please approve the PR.

rfd/0001-kubernetes-manager.md
marshall-lee: it was a bad rebase :(

rfd/0001-kubernetes-manager.md
marshall-lee: The "manager" word is used everywhere in the text meaning the k8s controller. The RFD itself is called "Kubernetes manager" and the definition of it was given in a "What" section.

marshall-lee: The word "operator" is confusing too. The word "controller" is good but typically managers consist of multiple controllers. You've got `kube-controller-manager`, `cert-manager` etc in your cluster.

pierrebeaucamp (approved):

tcsc (approved):

PR #335, opened 2021-10-07
Adds timeout to event handler sendEvent

gzigzigzeo:
What's inside:
- [x] Event handler now retries several times if it fails to send an event to Fluentd
- [x] Most of the log messages are moved debug level

### DIFF
--- a/event-handler/app.go
+++ b/event-handler/app.go
@@ -75,6 +75,13 @@ const (
 	sessionBackoffMax = 2 * time.Minute
 	// sessionBackoffNumTries is the maximum number of backoff tries
 	sessionBackoffNumTries = 5
+
+	// sendBackoffBase is an initial (minimum) backoff value.
+	sendBackoffBase = 1 * time.Second
+	// sendBackoffMax is a backoff threshold
+	sendBackoffMax = 10 * time.Second
+	// sendBackoffNumTries is the maximum number of backoff tries
+	sendBackoffNumTries = 5
 )
 
 // NewApp creates new app instance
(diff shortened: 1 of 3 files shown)


tcsc (approved):

event-handler/app.go
```diff
@@ -75,6 +75,13 @@ const (
 	sessionBackoffMax = 2 * time.Minute
 	// sessionBackoffNumTries is the maximum number of backoff tries
 	sessionBackoffNumTries = 5
+
+	// sendBackoffBase is an initial (minimum) backoff value.
+	sendBackoffBase = 1 * time.Second
```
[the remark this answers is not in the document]
gzigzigzeo: These are for event send retries. If somehow fluentd is down, we try to re-send an event 5 times within approx one minute, and exit if it fails. Yes, that are just a defaults which seem reasonable for me. Possibly, you have a suggestion here?

lib/backoff/decorr.go
```diff
@@ -35,12 +35,12 @@ type decorr struct {
 	clock clockwork.Clock
 }
 
-// NewDecorr initializes an alogoritm.
+// NewDecorr initializes an alogorithm.
```
Joerger, line 38:
```suggestion
// NewDecorr initializes an algorithm.
```

lib/backoff/decorr.go
```diff
@@ -35,12 +35,12 @@ type decorr struct {
...
 func NewDecorr(base, cap time.Duration, clock clockwork.Clock) Backoff {
 	return NewDecorrWithMul(base, cap, 3, clock)
 }
 
-// NewDecorrWithMul initializes a backoff alogoritm with a give multiplier.
+// NewDecorrWithMul initializes a backoff alogorithm with a give multiplier.
```
Joerger, line 43:
```suggestion
// NewDecorrWithMul initializes a backoff algorithm with a given multiplier.
```

PR #342, opened 2021-10-13
Dump session events to stdout by default in fluentd

gzigzigzeo:
What's inside:
- [x] Default generated configuration now includes fluentd session endpoint
- [x] Print events are serialised including Data field

### DIFF
--- a/event-handler/teleport_event.go
+++ b/event-handler/teleport_event.go
@@ -36,7 +36,7 @@ const (
 // TeleportEvent represents helper struct around main audit log event
 type TeleportEvent struct {
 	// event is the event
-	Event events.AuditEvent
+	Event interface{}
 	// cursor is the event ID (real/generated when empty)
 	ID string
 	// cursor is the current cursor value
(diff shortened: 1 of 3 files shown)


klizhentas:
@Joerger @tcsc can you folks please review and merge with buddy PRs?

event-handler/teleport_event.go
```diff
@@ -27,40 +27,52 @@ import (
...
 	// IsSessionEnd is true when this event is session.end
 	IsSessionEnd bool
-
+	// IsPrint is true when this event is print
+	IsPrint bool
 	// SessionID is the session ID this event belongs to
 	SessionID string
 }
 
+// printEvent represents and artificial print event struct which adds json-serialisable data field
```
Joerger, line 58:
```suggestion
// printEvent represents an artificial print event struct which adds json-serialisable data field
```

event-handler/teleport_event.go
```diff
@@ -71,6 +83,7 @@ func NewTeleportEvent(e events.AuditEvent, cursor string) (TeleportEvent, error)
 
 		hash := sha256.Sum256(data)
 		id = hex.EncodeToString(hash[:])
+		e.SetID(id)
```
Joerger, line 86: If the struct underneath `e` isn't a pointer, then `SetID` wouldn't apply to `i`. Since this section of code can't be certain that `e` is a pointer, it would be safer to move the `var i interface{} = e` line below this line.

tcsc (approved):

PR #345, opened 2021-10-18
App and Database resource

gzigzigzeo:
It's meant to go out after 8.0.0 gets released. 7.* does not have respective methods backported in the API. Needs to be merged after `feature/new-terraform-resources` https://github.com/gravitational/teleport-plugins/pull/341.

PR #351, opened 2021-10-21
Terraform tests

gzigzigzeo:
Merge after feature/terraform-apps-databases https://github.com/gravitational/teleport-plugins/pull/345

Please note:
1. Reference to Terraform will require update.
2. Database and App tests are now disabled because they won't work on the CI until 8.0.0 goes live.

### DIFF
--- a/terraform/provider/resource_teleport_cluster_audit_config.go
+++ b/terraform/provider/resource_teleport_cluster_audit_config.go
@@ -28,6 +28,10 @@ import (
 	"github.com/gravitational/teleport/api/types"
 )
 
+// Set is not implemented in the API: https://github.com/gravitational/teleport/blob/1944e62cc55d74d26b337945000b192528674ef3/api/client/client.go#L1776
+//
+// https://github.com/gravitational/teleport/pull/7465
+
 // resourceTeleportClusterAuditConfig returns Teleport cluster audit resource definition
 func resourceTeleportClusterAuditConfig() *schema.Resource {
 	return &schema.Resource{
@@ -59,8 +63,9 @@ func resourceAuditConfigCreate(ctx context.Context, d *schema.ResourceData, m in
 		return diagFromErr(err)
 	}
 
-	err = c.SetClusterAuditConfig(ctx, &n)
-	if err != nil {
+	// Linter generates false positive here because the API always returns error (see comment above)
+	err = c.SetClusterAuditConfig(ctx, &n) //nolint
+	if err != nil {                        //nolint
 		return diagFromErr(describeErr(err, types.KindClusterAuditConfig))
 	}
 
@@ -126,8 +131,8 @@ func resourceAuditConfigUpdate(ctx context.Context, d *schema.ResourceData, m in
 		return diagFromErr(err)
 	}
 
-	err = c.SetClusterAuditConfig(ctx, n)
-	if err != nil {
+	err = c.SetClusterAuditConfig(ctx, n) //nolint
+	if err != nil {                       //nolint
 		return diagFromErr(describeErr(err, types.KindClusterAuditConfig))
 	}
 
@@ -141,8 +146,8 @@ func resourceAuditConfigDelete(ctx context.Context, d *schema.ResourceData, m in
 		return diagFromErr(err)
 	}
 
-	err = c.DeleteClusterAuditConfig(ctx)
-	if err != nil {
+	err = c.DeleteClusterAuditConfig(ctx) //nolint
+	if err != nil {                       //nolint
 		return diagFromErr(describeErr(err, types.KindClusterAuditConfig))
 	}
 


terraform/test/main_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	s.terraformProvider = provider.Provider()
+	s.terraformProviders = make(map[string]func() (*schema.Provider, error))
+	s.terraformProviders["teleport"] = func() (*schema.Provider, error) {
+		// Terraform configures provider on every test step, but does not clean up previous one, which produces
+		// to "too many open files" at some point.
+		//
+		// With this statement we try to forcefully close previously opened client, which stays cached in
+		// the provider variable.
+		s.closeClient()
```
Joerger, line 138: Is this needed when you already close the client at the end of each test?

Also, is this only an issue due to the way `resource.Test` works? Or can this also affect a running terraform provider in production?

gzigzigzeo: Yes, this is mandatory. And yes, this due to the way `resource.Test` works. Every test consists of a sequence of steps. Terraform test framework initialises the provider instance at the beginning of every step without cleaning up a previous one. For some reason, there is no cleanup handler in the Terraform test framework so this piece of code is the simple workaround.

`s.closeClient()` at the end of each test, in fact, closes the latest Terraform client created on the very last step of the whole suite. I think, I need to remove these calls and close the latest connection in the suite shutdown handler instead.

It does not affect production. In production Terraform runs from scratch every time.

terraform/test/auth_preference_test.go
```diff
@@ -0,0 +1,87 @@
...
+				),
+			},
+			{
+				Config:   update, // Check that there is no state drift
+				PlanOnly: true,
+			},
+		},
+	})
+
+	s.closeClient()
```
tcsc, line 86: You can add automatic cleanup to the `testify` suite by implementing the `suite.AfterTest` interface, if that cleans things up a bit

terraform/test/main_test.go
```diff
@@ -0,0 +1,152 @@
...
+	terraformProviders map[string]func() (*schema.Provider, error)
+}
+
+func TestTerraform(t *testing.T) { suite.Run(t, &TerraformSuite{}) }
+
+func (s *TerraformSuite) SetupSuite() {
+	var err error
+	t := s.T()
+
+	logger.Init()
```
tcsc, line 61: The logger setup will be clobbered by the calls to `s.AuthSetup.Setup...()` below. It _might_ be better to pull it out of both here and `AuthSetup`, and explicitly configure the logger in the `TestMain()` for each package.

terraform/test/main_test.go
```diff
@@ -0,0 +1,152 @@
...
+	logger.Setup(logger.Config{Severity: "debug"})
+
+	// We set such a big timeout because integration.NewFromEnv could start
+	// downloading a Teleport *-bin.tar.gz file which can take a long time.
+	ctx := s.SetContextTimeout(2 * time.Minute)
+
+	s.AuthSetup.SetupSuite()
+	s.AuthSetup.Setup()
```
tcsc, line 69: Is `AuthSetup.Setup()` expected to be called once for each test, rather than just once when the `suite` is being initialised?

terraform/test/main_test.go
tcsc: Possibly stupid question: how does closing the client here force Terraform to clean up its Provider set?

terraform/test/main_test.go
```diff
@@ -0,0 +1,152 @@
...
+	teleportConfig lib.TeleportConfig
+	// teleportFeatures represents enabled Teleport feature flags
+	teleportFeatures *proto.Features
+	// plugin represents plugin user name
+	plugin string
+	// terraformConfig represents Terraform provider configuration
+	terraformConfig string
+	// terraformProvider represents an instance of a Terraform provider
+	terraformProvider *schema.Provider
+	// terraformProviders represents an array of providers
```
tcsc, line 51:
```suggestion
	// terraformProviders represents an array of provider factories that Terraform will use to instantiate the provider(s) under test.
```

terraform/test/oidc_connector_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+import (
+	"github.com/gravitational/trace"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+func (s *TerraformSuite) TestOIDCConnector() {
+	if !s.teleportFeatures.AdvancedAccessWorkflows {
+		return
```
tcsc, line 27:
```suggestion
		s.T().Skip("AdvancedAccessWorkflows are disabled")
```

terraform/test/cluster_networking_config_test.go
```diff
@@ -0,0 +1,87 @@
...
+				Config: create,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(name, "kind", "cluster_networking_config"),
+					resource.TestCheckResourceAttr(name, "metadata.0.labels.example", "yes"),
+					resource.TestCheckResourceAttr(name, "spec.0.client_idle_timeout", "30m0s"),
+				),
+			},
+			{
+				Config:   create, // Check that there is no state drift
+				PlanOnly: true,
```
tcsc, line 69: Does `PlanOnly` cause the Terraform test runner to automatically fail the test on a nonzero-op plan? Or are there extra checks required?

terraform/test/main_test.go
gzigzigzeo: It does not. Unfortunately, there is no straightforward way to clean up a provider before it quits (https://github.com/hashicorp/terraform-provider-vsphere/issues/618). Let me explain this a bit more detailed to ensure that we are on the same page. 

Terraform provider is the struct which contains resources schema, methods and so on. The most important piece is `ConfigureContext` function, which opens the connection to Teleport and saves it to provider object metadata. It is called by Terraform when it initialises the provider set. The problem is that there is no `Cleanup` or `Close` counterpart. When Terraform finishes it's job in production, it just quits.

Every test consist of the following structures:

```go
Steps: []resource.TestStep{
	{
		Config: create,
		Check: resource.ComposeTestCheckFunc(
			resource.TestCheckResourceAttr(name, "kind", "cluster_auth_preference"),
			resource.TestCheckResourceAttr(name, "metadata.0.labels.example", "yes"),
			resource.TestCheckResourceAttr(name, "spec.0.disconnect_expired_cert", "true"),
		),
	},
	{
		Config:   create, // Check that there is no state drift
		PlanOnly: true,
	},
	{
		Config: update,
		Check: resource.ComposeTestCheckFunc(
			resource.TestCheckResourceAttr(name, "kind", "cluster_auth_preference"),
			resource.TestCheckResourceAttr(name, "metadata.0.labels.example", "yes"),
			resource.TestCheckResourceAttr(name, "spec.0.disconnect_expired_cert", "false"),
		),
	},
	{
		Config:   update, // Check that there is no state drift
		PlanOnly: true,
	},
},
```
At the beginning of every step (an array item above), the fresh new instance of the provider gets created and Terraform calls `ConfigureContext`. An old instance stays in memory with the old connection open. So, there would be 4 open connections to Teleport after this test is finished.

At some point, when there are too many open connections, it returns "too many open files" error. I guess, the reason for this error is go grpc client architecture, which is quite complicated and heavy, but I am not 100% sure. There might be other reasons for this. So, what I do here - I close an old provider connection every time Terraform test framework requests a new provider instance.

That would be much better if I could have shared the same connection between all provider instances, but this would require artificial modifications of the `ConfigureContext` function, which I do not want in production. Actually, to solve this issue, I simply need some kind of a connection pool, but I am not sure that provider should be responsible for this.

terraform/test/cluster_networking_config_test.go
gzigzigzeo: Yes, that is correct. It's the equivalent of `terraform plan`. It will show the state drift if any.

terraform/test/main_test.go
gzigzigzeo: Yes, these lines are artefacts of moving this setup to lib. I just missed their removal. I agree, that might be better to setup logger in TestMain(), but for now, this setup is the same for all plugins. Let me remove these lines and for now keep it like this for the sake of uniformity. We could get back to it in the separate PR.

terraform/test/cluster_audit_config_test.go
```diff
@@ -0,0 +1,53 @@
...
+			}			
+		}
+	`
+
+	resource.Test(s.T(), resource.TestCase{
+		ProviderFactories: s.terraformProviders,
+		Steps: []resource.TestStep{
+			{
+				Config:      create,
+				ExpectError: regexp.MustCompile("not implemented"),
```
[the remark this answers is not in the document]
gzigzigzeo: I not know much of the background here, but setting audit config via API was disabled at some point (https://github.com/gravitational/teleport/pull/7465). Added link to the PR here. I can remove this entity if there is no plan to use change it via API anymore.

terraform/test/main_test.go
gzigzigzeo: That's the naming issue. `Setup()` sets a Teleport service (or services) up. It should be used once per suite or separate test like it's done in 'lib/testing/integration' tests. Renamed to 'SetupService()'.

Joerger (approved):

terraform/test/main_test.go
tcsc: Thanks for the detailed reply.

terraform/test/cluster_audit_config_test.go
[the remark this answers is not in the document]
tcsc: I'd leave it for now - PR link helps  with context.

tcsc (approved):

terraform/provider/resource_teleport_cluster_audit_config.go
```diff
@@ -59,8 +63,9 @@ func resourceAuditConfigCreate(ctx context.Context, d *schema.ResourceData, m in
 		return diagFromErr(err)
 	}
 
-	err = c.SetClusterAuditConfig(ctx, &n)
-	if err != nil {
+	// Linter generates false positive here because the API always returns error (see comment above)
```
tcsc, line 66: If it always returns an error, why are we doing it?

tcsc (approved):
Looks good in general, but I would like to double check why we're doing things that are going to fail 100% of the time (the issues identified by the linter), and what effect those failures will have on the code that calls them.

I'm not yet convinced that disabling the lint is the right thing to do here.

tcsc (commented):
See above.

tcsc (changes requested):
See Above

terraform/provider/resource_teleport_cluster_audit_config.go
gzigzigzeo: That's the legacy. As I mentioned above, those API methods were disabled some time ago (Teleport PR#7465). They always return NotImplementedError, and that causes linter failure: it detects that the result of the method call is always error, hence, there's no reason to perform `if err != nil`. I can remove this entity from Terraform if we have no plans to revive it. And I think that API methods should have been removed, but for some reason they are still there - and that's why I chose to keep the resource on Terraform side as well.

Joerger: `AuditConfig` was not meant to be dynamically configurable, so they should be removed from the Terraform provider. - https://github.com/gravitational/teleport/pull/7465. 

These unimplemented methods are used for so that the client fits internal interfaces, but there's no reason they should be exposed in our public API. I'll make a PR to fix it.

tcsc (approved):

terraform/provider/resource_teleport_cluster_audit_config.go
tcsc: OK, so we will merge this as-is and remopve these calls in a subsequent PR. Everyone OK with that?

PR #356, opened 2021-11-03
Containerize the Slack plugin

marshall-lee:


access/slack/config.go
```diff
@@ -61,6 +64,13 @@ func LoadConfig(filepath string) (*Config, error) {
 	if err := t.Unmarshal(conf); err != nil {
 		return nil, trace.Wrap(err)
 	}
+	if strings.HasPrefix(conf.Slack.Token, "/") {
+		tokenBytes, err := ioutil.ReadFile(conf.Slack.Token)
```
tcsc, line 68: Is there a max token length or anything that we should be concerned about?

tcsc (approved):

access/slack/config.go
marshall-lee: The length of a token is considered further in `CheckAndSetDefaults`

tcsc: OK, so we check for token absence - but what if someone points this at a 10GB file?

Update: I just checked mainline Teleport, and we don't enforce a bound there so this is probably OK. (see $TELEPORT_REPO/lib/utils/token.go:29-40)

tcsc (approved):

access/slack/Makefile
```diff
@@ -16,6 +16,8 @@ RELEASE_MESSAGE := "Building with GOOS=$(OS) GOARCH=$(ARCH)."
 
 VERSRC = version.go gitref.go
 
+IMG ?= hashtable/teleport-access-slack:latest
```
Joerger, line 19: This should probably be `quay.io/gravitational/teleport-access-slack:$(Version)`, which is in line with the teleport docker image we publish. @russjones is this correct?

For testing you could still use your own docker hub account.

access/slack/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+COPY access/slack access/slack
+COPY vendor vendor
+COPY lib lib
+
+# Build
+RUN make -C access/slack
+
+# Use distroless as minimal base image to package the manager binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM gcr.io/distroless/base:latest
```
Joerger, line 20: I'm not very knowledgeable in the area, but distroless does seem nice for the plugins. However, the previous [dockerize slack PR](https://github.com/gravitational/teleport-plugins/pull/199) had some comments about copying aspects of the teleport dockerfile, such as using Ubuntu 18.04 or 20.04 as a base.  Has this already been internally discussed, and do we plan on making other plugin docker images using the base image?

access/slack/Makefile
marshall-lee: Yes

access/slack/config.go
marshall-lee: Sure. Don't think we need to check token length because it's something that admin provides, not an attacker.

access/slack/Dockerfile
marshall-lee: @Joerger 

Nope it wasn't discussed. This PR is a request for such a discussion :)

Initially, the idea of using distroless came up when I was playing with Kubebuilder framework while working on #319. The framework generated a `Dockerfile` that used `distroless:static` and I was really impressed by the resulting image size.

Then I decided to try it for Slack plugin and it failed. Then I tried `distroless:base` (it's the same as `distroless:static` but contains glibc) and it was okay. It turned out that `distroless:static` image is for static Go binaries built without cgo.

`CGO_ENABLED=1" is another thing that we copied from Teleport's `Makefile` and I don't understand why :) So maybe it's time to stop copying things from Teleport blindly? I don't see an argument why we might need Ubuntu for such a small things like plugins.

Want to invite @russjones and @webvictim for the discussion.

marshall-lee:
@tcsc @Joerger I also added a Helm chart that I wrote when tested the Kubernetes Manager #319

access/slack/Dockerfile
tcsc: We should probably bin the base image to a specific version (or hash) as well. We should opt-in to updates, rather than be surprised by them ;-)

tcsc:
I'm unsure how to proceed here. We have the requisite approvals, but the commentary suggests we awaiting the discussion re: `ubuntu` vs `distroless` as a base image for the slack plugin. I'm happy to merge as-is, but what are other people expecting?

access/slack/Dockerfile
marshall-lee: Yes but distroless images don't have a versioning scheme I suppose. So I just used the latest hash

marshall-lee:
@tcsc 

This PR also contains a Helm chart that doesn't work by now because Kubernetes manager is not yet released.
And then the manager is released also the Teleport helm chart should also be updated https://github.com/gravitational/teleport/pull/8789

access/slack/charts/teleport-access-slack/values.yaml
```diff
@@ -0,0 +1,54 @@
+##################################################
+# Values that must always be provided by the user.
+##################################################
+
+# The name of Kubernetes service of Teleport.
+# Must be in the same namespace as the plugin.
+teleportService: ""
```
Joerger, line 7: It's unclear what this value should be to me, is this supposed to be the teleport cluster name?
```suggestion
teleportClusterName: ""
```

Joerger (commented):
> This PR also contains a Helm chart that doesn't work by now because Kubernetes manager is not yet released.

What's the reason for adding the helm charts into this PR then? Can you move this to a separate PR and add a README for it?

access/slack/charts/teleport-access-slack/values.yaml
marshall-lee: @Joerger No, clusterName differs from service name. It's literally the name of Kubernetes service resource in the same namespace.

So if, for example, you release `teleport-cluster` Helm chart from Teleport repo then the `teleportService` will match the helm release name.

marshall-lee:
Moved helm charts to #366

marshall-lee:
> What's the reason for adding the helm charts into this PR then?

Oh initially I did it in the same branch because it was convenient for me to test the Kubernetes deployment. I was rebuilding my image repeatedly and testing how it works in Kubernetes.

PR #386, opened 2021-12-01
Remove broken access example

Joerger:
An OS user was having trouble getting started with the access plugin example so I took a look and realized it is out of date and can only be used from within the teleport-plugins go module.

Rather than fixing it, I've removed it and added a README with a short description and links to newer documentation and examples.


tcsc (approved):

Joerger:
@klizhentas Can you give this an approval if you agree with this change?

PR #392, opened 2021-12-13
Escape and truncate request, resolution and review reasons in access-slack and access-mattermost

espadolini:
Because of the way "markdown" (I use this term loosely here) works in Slack and Mattermost, escaping arbitrary strings is somewhat tricky:

- there's no way to escape backticks within single backticks, so we must use triple backticks; escaping backticks within triple backticks can be done by abusing `U+FEFF` as a zero-width no-break space
- triple backticks interact strangely with whitespace in Mattermost, to the extent that a double newline will result in the code block not actually being parsed
- we must add a newline right after the opening backticks, or Mattermost might interpret the first word as a syntax highlighting directive

In addition to the message escaping, with this we also conservatively truncate all the potentially user-provided text to stay well within the service limits; said service limits are now somewhat reflected in our mocks of Slack and Mattermost, which will return a 400 for overlong messages. The limits there are also more conservative than what the actual services allow.

This fixes TEL-Q321-1 and TEL-Q321-6.

### DIFF
--- a/access/mattermost/mattermost_test.go
+++ b/access/mattermost/mattermost_test.go
@@ -28,8 +28,8 @@ import (
 )
 
 var msgFieldRegexp = regexp.MustCompile(`(?im)^\*\*([a-zA-Z ]+)\*\*:\ +(.+)$`)
-var requestReasonRegexp = regexp.MustCompile(`(?im)^\*\*Reason\*\*:\ ` + "```\\n" + `(.*?)` + "```" + `(.*?)$`)
-var resolutionReasonRegexp = regexp.MustCompile(`(?im)^\*\*Resolution reason\*\*:\ ` + "```\\n" + `(.*?)` + "```" + `(.*?)$`)
+var requestReasonRegexp = regexp.MustCompile("(?im)^\\*\\*Reason\\*\\*:\\ ```\\n(.*?)```(.*?)$")
+var resolutionReasonRegexp = regexp.MustCompile("(?im)^\\*\\*Resolution reason\\*\\*:\\ ```\\n(.*?)```(.*?)$")
 
 type MattermostSuite struct {
 	Suite
--- a/access/mattermost/mattermost_test.go
+++ b/access/mattermost/mattermost_test.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"runtime"
 	"sort"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"testing"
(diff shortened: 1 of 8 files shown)


espadolini:
This doesn't deal with message size overflow resulting from an access request for 1000 or so roles, but that's likely not a sensible configuration to begin with, so I don't think we should worry much about that.

access/mattermost/mattermost_test.go
```diff
@@ -27,6 +28,8 @@ import (
 )
 
 var msgFieldRegexp = regexp.MustCompile(`(?im)^\*\*([a-zA-Z ]+)\*\*:\ +(.+)$`)
+var requestReasonRegexp = regexp.MustCompile(`(?im)^\*\*Reason\*\*:\ ` + "```\\n" + `(.*?)` + "```" + `(.*?)$`)
```
Joerger, line 31: nit: did you separate this into `` and "" parts to avoid using \\\\s? I'd prefer a single string with \\\\s personally.
```suggestion
var requestReasonRegexp = regexp.MustCompile("(?im)^\\*\\*Reason\\*\\*:\\ ```\\n(.*?)```(.*?)$")
```

access/mattermost/mattermost_test.go
```diff
@@ -27,6 +28,8 @@ import (
...
+var resolutionReasonRegexp = regexp.MustCompile(`(?im)^\*\*Resolution reason\*\*:\ ` + "```\\n" + `(.*?)` + "```" + `(.*?)$`)
```
Joerger, line 32: nit: same as above
```suggestion
var requestReasonRegexp = regexp.MustCompile("(?im)^\\*\\*Resolution Reason\\*\\*:\\ ```\\n(.*?)```(.*?)$")"
```

access/slack/bot.go
```diff
@@ -268,7 +273,11 @@ func (b Bot) msgSections(reqID string, reqData RequestData) []BlockItem {
 		msgFieldToBuilder(&builder, "Role(s)", strings.Join(reqData.Roles, ","))
 	}
 	if reqData.RequestReason != "" {
-		msgFieldToBuilder(&builder, "Reason", reqData.RequestReason)
+		// Slack has a 4000 character recommendation (with a more generous 40000
```
Joerger, line 276: Rather than adding duplicates and pointers to this comment, can you make 500 a constant and move the comment to that constant? Maybe make it a separate constant for slack and mattermost as well.

lib/escape.go
```diff
@@ -0,0 +1,32 @@
+package lib
+
+import "strings"
+
+// MarkdownEscape wraps some text `t` in triple backticks (escaping any backtick
+// inside the message), limiting the length of the message to `n` runes (inside
+// the single preformatted block). The text is trimmed before escaping.
+// Backticks are escaped and thus count as two runes for the purpose of the
+// truncation.
```
Joerger, line 9: Could you add an example to this comment?
```
e.g. x -> y
```

access/slack/bot.go
espadolini: Added commented constants for both.

lib/escape.go
espadolini: I went ahead and added an example test with a small handful of test cases.

Joerger (approved):

PR #393, opened 2021-12-13
Update golang image and teleport version for CI

Joerger:
Also replaces use of deprecated "admin" role with "editor" role.

Closes https://github.com/gravitational/teleport-plugins/issues/389

PR #394, opened 2021-12-14
Fix integration tests when migrating to the new Teleport version

marshall-lee:
Related to #393.


Joerger:
Thanks for this, I added the commit to https://github.com/gravitational/teleport-plugins/pull/393

PR #397, opened 2021-12-14
Check X-Gitlab-Token in constant time

espadolini:
This fixes TEL-Q321-12.

### DIFF
--- a/access/gitlab/webhook_server.go
+++ b/access/gitlab/webhook_server.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"crypto/subtle"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


access/gitlab/webhook_server.go
```diff
@@ -87,7 +88,8 @@ func (s *WebhookServer) processWebhook(rw http.ResponseWriter, r *http.Request,
 		http.Error(rw, "", http.StatusBadRequest)
 		return
 	}
-	if r.Header.Get("X-Gitlab-Token") != s.secret {
+	// the length of the secret token is not particularly confidential, so it's ok to leak it here
+	if subtle.ConstantTimeCompare([]byte(r.Header.Get("X-Gitlab-Token")), []byte(s.secret)) == 0 {
```
espadolini, line 92: This copies both strings into `[]byte`s; the only other alternative I could think of would've been to just copy the body of `subtle.ConstantTimeCompare` to have it work on `string`, but it didn't seem worth the subsequent maintenance effort.

tcsc: Seems reasonable.

tcsc (approved):

PR #398, opened 2021-12-14
Limit the amount we read from webhooks as per platform limits

espadolini:
Fixes TEL-Q321-11.

### DIFF
--- a/access/gitlab/webhook_server.go
+++ b/access/gitlab/webhook_server.go
@@ -21,6 +21,7 @@ import (
 	"crypto/subtle"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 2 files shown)


access/gitlab/webhook_server.go
```diff
@@ -93,12 +94,18 @@ func (s *WebhookServer) processWebhook(rw http.ResponseWriter, r *http.Request,
 		return
 	}
 
-	body, err := ioutil.ReadAll(r.Body)
+	// as per https://docs.gitlab.com/ee/user/gitlab_com/index.html#webhooks,
+	// the webhook payload size is limited to 25MB
+	body, err := ioutil.ReadAll(io.LimitReader(r.Body, 30*1024*1024))
```
tcsc, line 99: Mismatch between comment and code - we're giving them an extra 5MB over what the doc says. Are we allowing a little padding for some overhead?

Might also be an idea to pull it out into a constant as its used in a couple of places here.

access/jira/webhook_server.go
```diff
@@ -95,12 +96,19 @@ func (s *WebhookServer) processWebhook(rw http.ResponseWriter, r *http.Request,
 	ctx, log := logger.WithField(ctx, "jira_http_id", httpRequestID)
 
 	var webhook Webhook
-	body, err := ioutil.ReadAll(r.Body)
+	// as per
+	// https://developer.atlassian.com/cloud/jira/platform/webhooks/#known-issues,
+	// the webhook payload size is limited to 25MB
+	body, err := ioutil.ReadAll(io.LimitReader(r.Body, 30*1024*1024))
```
tcsc, line 102: Same question as above

access/gitlab/webhook_server.go
espadolini: I was leaving some overhead room but the webhook payloads shouldn't get anywhere near that size to begin with, I'll just make it adhere to the declared limits instead.

Joerger (approved):

Issue #246, opened 2021-05-26
Support tokens and other secrets from files for plugins


stevenGravy:
LIke the teleport join tokens allow for specifying files with tokens instead of just directly in the files.  
Ex:
```toml
[slack]
token = "xoxb-RdXVfx8Y"             # Slack Bot OAuth token
```
can have the token in a file `/var/lib/teleport-plugins/token.txt`
```toml
[slack]
token = "/var/lib/teleport-plugins/token.txt"             # Slack Bot OAuth token
```

This is also helpful in case the plugin is being run in a container such as a k8s so it can be a secret.

PR #399, opened 2021-12-14
Read api tokens from files for all of the access services

lxea:
Read tokens using the existing lib.ReadPassword and update access/slack to use ReadPassword instead as well.

#246


tcsc (approved):
Looks fine to me. Should we be concerned about someone pointing a gigabyte-sized file at any of these settings?

Joerger (approved):

PR #408, opened 2022-01-06
Remove vendor

Joerger:
Following the example of the main teleport repo, the vendor directory is no longer needed by teleport-plugins.

RFD - https://github.com/gravitational/teleport/pull/9545
Teleport PR - https://github.com/gravitational/teleport/pull/9637


Joerger:
@gzigzigzeo I noticed that `make -C terraform gen-schema` is broken on master, since terraform no longer has its own vendor like it did originally. This PR will make the current `gen-schema` unfixable. Does this need to be fixed now? We could try reading the packages directly from the local `$GOPATH/pkg/mod` cache, or try moving the schema generation to the teleport repo. Maybe there is a simpler solution. Any thoughts?

tcsc:
LGTM (once `make -C terraform gen-schema`  is resolved - either fixed or explicitly deferred)

Joerger:
> LGTM (once `make -C terraform gen-schema` is resolved - either fixed or explicitly deferred)

I made a band-aid fix for this, but it uses a symlink which is far from ideal, so I'll hold off on merging for now.

terraform/Makefile
```diff
@@ -60,11 +60,28 @@ reapply:
...
+# protoc expects to find .proto files needed by types.proto by its full path.
+# Due to the way go modules work, we don't have the exact path to the cached
+# module on file, so we make a temporary directory with a symlink to the cached 
+# module to get the desired file structure. This is a bandaid fix due to the
+# removal of /vendor, and should be replaced with a better solution. In the main
+# teleport repo, we ran into a similar issue with our grpc generation, and we used
+# the M flag to achieve a similar result, but I haven't found the way to do this here.
 gen-schema:
+	mkdir -p $(LOCALDIR)/vendor/github.com/gravitational/teleport/api
+	ln -s $(API_MOD_PATH)/* $(LOCALDIR)/vendor/github.com/gravitational/teleport/api
```
Joerger, line 77: @zmb3 This protoc issue is similar to the one we were working on the in main repo. I couldn't find any "nice" solution like the `M` flag supported by `gogofast_out`. I don't think it's worth spending too much time on this right now, but does any workaround come to mind? ( other than a symink :D )

Joerger:
@gzigzigzeo @tcsc I've updated the `gen-schema` target to make a tmp directory with `github.com/gravitational/teleport/api/types/wrappers/wrappers.proto` to get it working without the vendor directory. If you guys are ok with this solution, let's go ahead with it as is. 

Otherwise, we could update `protoc-gen-terraform` to handle this kind of issue with a cli flag, similar to the `M` option supported by https://github.com/gogo/protobuf which we use in the main Teleport repo for this issue.

tcsc:
Is either of these a "never have to bother with it again" solution? TBH I'm probably OK with this for now, as long as there is an issue capturing what needs doing with `protoc-gen-terraform`.

tcsc (approved):
Approved, assuming that an issue is created to capture the "ideal" solution

Issue #429, opened 2022-02-15
Add the `aws_role_arns` value to our Terraform provider roles 


pschisa:
Add the ability to set the `aws_role_arns` value in role via the Terraform provider and document this in the example (https://github.com/gravitational/teleport-plugins/blob/master/terraform/example/role.tf).

This is important for managing roles that are used for AWS console access feature: https://goteleport.com/docs/application-access/guides/aws-console/#step-4-configure-teleport-iam-role-mapping

Issue #428, opened 2022-02-15
Add the ability to set an OIDC auth_connector version in the Terraform provider


pschisa:
Add the ability to set the `version` value in an OIDC connector via the Terraform provider and document this in the example (https://github.com/gravitational/teleport-plugins/blob/master/terraform/example/oidc_connector.tf).

This is important now that version 3 has been released for the connector to allow for improved Google SSO experience: https://goteleport.com/docs/ver/9.0/enterprise/sso/google-workspace/#create-a-oidc-connector

Joerger:
https://github.com/gravitational/teleport-plugins/pull/431 updates the Terraform provider to use `OIDCConnectorV3`.

I don't believe there is currently a way for the Terraform Provider to make the OIDC Connector version configurable, as the `teleport/api` package only provides one or the other. See https://github.com/gravitational/teleport-plugins/pull/263#discussion_r662515098 for a similar issue.

pschisa:
That will resolve this issue yes, if the provider uses v3 .  Not sure if that will raise any new issues with people expecting to continue using v2 going forward (though that use case seems less likely)

Issue #427, opened 2022-02-11
New Error after upgrading to Terraform Provider 8.2.0 from 8.0.7


pschisa:
After upgrading the Terraform provider from 8.0.7 to 8.2.0, the Terraform code no longer works and gives the following error:

~~~
Error: field spec.0.allow not found in target struct
~~~

Attaching a santized version of the Terraform provider code.  Please let me know if any further information is required

~~~
resource "random_id" "token" {
  byte_length = 16
}

resource "teleport_provision_token" "token" {
  metadata {
    name    = random_id.token.hex
    expires = timeadd(timestamp(), "1h")
    labels = {
      test = "yes"
    }
  }
  spec {
    roles = ["Kube", "App", "Db"]
  }
}

terraform {
  required_version = ">= 1.0.0"

  required_providers {
    teleport = {
      version = "8.2.0"
      source  = "gravitational.com/teleport/teleport"
    }
  }
}

provider "teleport" {
  addr = ""
}

output "teleport_token" {
  value = random_id.token.hex
}
~~~

programmerq:
I also ran into this in my own lab.

I did some experimenting to try to make it work, but I wasn't able to find a solution.

This change in #408 does change `Allow` to `allow` which may be related: https://github.com/gravitational/teleport-plugins/pull/408/files#diff-160c8d9d2e45912ae5ad2964f3846088b7d9ef45fc606eff48937c4ab7dc9dd4

Joerger:
> This change in #408 does change `Allow` to `allow` which may be related: https://github.com/gravitational/teleport-plugins/pull/408/files#diff-160c8d9d2e45912ae5ad2964f3846088b7d9ef45fc606eff48937c4ab7dc9dd4

This is [likely the culprit](https://github.com/gravitational/teleport-plugins/pull/415#discussion_r796641178), thanks for spotting it.  That'd be my mistake, sorry about that, I'll PR a fix.

@gzigzigzeo Can we find a way to make `gen-schema` idempotent to avoid this issue going forward?

gzigzigzeo:
@Joerger Yes, it's already done in the latest version, which I am working on now. I guess, we need to fix `types.proto` as well.

benarent:
+1 I ran into this issue and had to build from master

PR #431, opened 2022-02-16
Update api dependency to v8.3.0

Joerger:
Changes:
 - Update api dependency to v8.3.0 
 - run `make -C terraform gen-schema`

teleport/api v8.3.0 introduces a fix to an inconsistency in `types.proto` where a proto field was not capitalized.

Closes https://github.com/gravitational/teleport-plugins/issues/427
Closes https://github.com/gravitational/teleport-plugins/issues/429
Closes https://github.com/gravitational/teleport-plugins/issues/428

### DIFF
--- a/terraform/tfschema/types_terraform.go
+++ b/terraform/tfschema/types_terraform.go
@@ -76,10 +76,10 @@ var (
 	SchemaUserV2 = GenSchemaUserV2()
 	// SchemaMetaUserV2 is schema metadata for UserV2 is version 2 resource spec of the user
 	SchemaMetaUserV2 = GenSchemaMetaUserV2()
-	// SchemaOIDCConnectorV2 is schema for OIDCConnectorV2 represents an OIDC connector.
-	SchemaOIDCConnectorV2 = GenSchemaOIDCConnectorV2()
-	// SchemaMetaOIDCConnectorV2 is schema metadata for OIDCConnectorV2 represents an OIDC connector.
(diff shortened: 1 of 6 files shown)


terraform/tfschema/types_terraform.go
```diff
@@ -1346,38 +1364,56 @@ func GenSchemaMetaProvisionTokenV2() map[string]*accessors.SchemaMeta {
 					IsTime:     false,
 					IsDuration: false,
 				},
-
+				// Allow is a list of TokenRules, nodes using this token must match one
+				// allow rule to use this token.
 				"allow": {
-					Name:       "allow",
```
Joerger, line 1351: This is the line which fixes #427

Joerger:
@russjones We will need to release a new v8.3 for teleport-plugins to fix the terraform issue linked above.

I think we should update the teleport-plugins release process to keep the `teleport/api` dependency up to date. If you agree, I can add some steps to `make version` to partially automate this, though we will probably need to manually update `terraform/gen_teleport.yaml`.

terraform/tfschema/types_terraform.go
Joerger: @pschisa @programmerq Can you confirm that this fixes the issue?

programmerq: I just built this from the branch and confirmed that it this change fixes the issue!

tcsc (approved):

r0mant (approved):

Issue #411, opened 2022-01-20
Have slackbot post to a channel for a requested role


klizhentas:
## Description

Have slack plugin post to a channel mapped to a role requested. Instead of DMs. Folks interested in review of the roles would join channel themselves.

In this case slack plugin does not DM, but posts to a channel with a name of the requested role. If multiple roles are requested, post to all of the channels matching.

Timeline is 1-2 months max. If we can implement it faster.

russjones:
Late March.

russjones:
@xinding33 @klizhentas What is the proposed syntax here?

```toml
[slack]
# Slack Bot OAuth token
token = "xoxb-11xx"

# Optional Slack Rooms
recipients = ["team","devops"]

# Channel map defines a mapping of roles to channel names to send requests to.
[channel_map]

# Send all requests for role to channel "dev-foo" .
"foo"= "dev-foo"

# Wildcard, send any non-matching requests to channel "dev".
"*"= "dev"
```

klizhentas:
@russjones I think it should be an error to specify both at the same time, to avoid confusion.

xinding33:
@russjones I'm probably just missing some context on the use cases we are trying to support but what purpose does `recipients` serve? Seems to me that the `[role_to_channel]` mapping, if used together with wildcard, covers all possible mapping cases.

russjones:
@xinding33 `recipients` is where all requests get sent, so it would be either `recipients` or `role_to_channel`.

Let's make that a requirement, you can only define one and the process should fail to start if you have not set at least one.

Also, `*` should always by required in `role_to_channel`.

Issue #279, opened 2021-08-09
Restrict slack plugin recipients to specific requested roles


pschisa:
The feature request is to allow for the slack plugin recipients to only receive notifications if a specific Teleport role is requested.  The matching of requested roles to recipients should be user-defined and could default to all recipients receive notifications for all requested roles unless otherwise configured.  This would allow users to restrict approval request notifications to the appropriate reviewer slack channels when multiple teams work on the same Teleport cluster.

PR #433, opened 2022-02-23
Slackbot role to recipients mapping

Joerger:
### Changes
 - Add support for `role_to_recipients` configuration option.
 - Deprecate `slack.recipients` since its functionality is equivalent to `role_to_recipients["*"]`.
   - `slack.recipients` is still supported, but it can't be used at the same time as `role_to_recipients`, and using it will result in a deprecation warning message. It will be removed in v11.0.0.
 - Update `README.md`
 - Update Teleport Docs - https://github.com/gravitational/teleport/pull/10607
   - Removed `INSTALL.md` which is unmaintained and a copy of the docs.
   - Brought API configuration docs up to date.
   - Replaced docs for `slack.recipients` with `role_to_recipients`.

### Usage

```toml
[teleport] 
addr = "proxy.example.com:443"
identity = "/var/lib/identity"

[slack]
token = "token"

[role_to_recipients]
"dev" = "devs-slack-channel"
"*" = ["admin@email.com", "admin-slack-channel"]
```

Note that `role_to_recipients` uses a custom marshaler to allow both `string` and `[]string` values.

Closes https://github.com/gravitational/teleport-plugins/issues/411
Closes https://github.com/gravitational/teleport-plugins/issues/279


access/slack/config.go
```diff
@@ -1,56 +1,62 @@
...
 	"github.com/gravitational/teleport-plugins/lib"
 	"github.com/gravitational/teleport-plugins/lib/logger"
+	"github.com/gravitational/teleport/api/types"
 	"github.com/gravitational/trace"
 	"github.com/pelletier/go-toml"
 )
 
 // Config stores the full configuration for the teleport-slack plugin to run.
 type Config struct {
-	Teleport lib.TeleportConfig `toml:"teleport"`
```
Joerger, line 14: providing the `toml:"..."` explicitly is only needed for non-lower-camel-case options. e.g. `role_to_recipients`.

access/slack/config.go
```diff
@@ -1,56 +1,62 @@
...
+	Log        logger.Config
 }
 
 // SlackConfig holds Slack-specific configuration options.
 type SlackConfig struct {
-	Token      string   `toml:"token"`
-	Recipients []string `toml:"recipients"`
+	Token string
+	// DELETE IN 10.0.0 (Joerger) - use "role_to_recipients["*"]" instead
+	Recipients []string
```
tcsc, line 26: Should we show  a deprecation message when this is configured?

tcsc (approved):

Joerger:
@xinding33 can you check the product decision section when you get a chance?

r0mant:
@Joerger FYI we discussed this on this week's product meeting and agreed to keeping the old `slack.recipients` for backwards compatibility but deprecating/hiding it from any user-facing material in favor of the new mapping syntax.

PR #435, opened 2022-03-02
Terraform provider (Framework version)

gzigzigzeo:
Review notes:

Most of the provider code is generated with `go:generate` from templates in `_gen` folder. There are four types of the generated files: singular/plural resource, singular/plural data source with a few exceptions. You can use the following resources as a source: `teleport_provision_token` (plural resource/data source with autogenerated id), `teleport_cluster_networking_config` (singular resource/data source). There are very few logical differences with other resources. Possibly, you might want to check `role` (it has default version) and `user` (it uses create/update instead of upsert).


terraform/install.mk
```diff
@@ -1,4 +1,4 @@
-VERSION=8.3.1
+VERSION=8.2.0
```
Joerger, line 1: Version is out of date

terraform/main.go
```diff
@@ -13,16 +13,20 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-
 package main
 
+//go:generate go run ./_gen/main.go
```
Joerger, line 18: Correct me if I'm wrong, but I'm not sure generating here makes sense. Don't the gen files need to present for the program to compile, and this generation will only trigger on runtime? The provider won't compile without the files already being generated.

terraform/protoc-gen-terraform-teleport.yaml
```diff
@@ -0,0 +1,338 @@
...
+    - "AuthPreferenceV2.Metadata.Name" # It's a signleton resource
+
+    # NetworkingConfig
+    - "ClusterNetworkingConfigV2.Metadata.Name" # It's a signleton resource
+
+    # Status
+    - "DatabaseV3.Status"
+
+    # Role
+    - "RoleV4.Spec.Allow.Namespaces" # These fields are not settable via API
```
Joerger, line 125: As we update the API and regenerate the schema, is there any clear criteria we can follow to add these special cases for new fields? I'm guessing we'll just have to handle them on a case by case basis?

Joerger (commented):
Can you update the readme? some of it is out of date with this change.

Additionally, the `Regenerating the schema` section should be updated to `Updating the provider` and include instructions on regenerating with `_gen/main.go`. In fact, would it make more sense to include the a make target for `_gen/main.go` generation and include it in `make gen-tfschema`?

Also I ran into some issues trying use the provider:
 - Running `make gen-tfschema` updates the timestamp proto import and makes `make apply` fail with error:
```
provider/data_source_teleport_app.go:28:2: package github.com/gravitational/teleport-plugins/terraform/tfschema imports github.com/golang/protobuf/ptypes/timestamp from implicitly required module; to add missing requirements, run:
	go get github.com/golang/protobuf@v1.4.3
```
 - Running `make apply` returns a host of errors. Is the example tf configuration not usable without adjustments? Are we able to make a bare bones tf config which doesn't require any editing of the `.tf` files?
 - `make test` is failing locally

gzigzigzeo:
@Joerger That's a weird timestamp behaviour. Never experienced it before, will look into it first thing in the morning. Have you had installed `protoc-gen-terraform` from the PR branch?

It makes sense to combine _gen and gen-tfschema, thanks for the suggestion.

Joerger:
> Have you had installed `protoc-gen-terraform` from the PR branch?

Yes

terraform/protoc-gen-terraform-teleport.yaml
gzigzigzeo: No, unfortunately there's no such criteria because, for instance, there's no way of knowing if a field is set to a default value on the API side after object creation and should be marked as computed on the Terraform side. However, I think that we could introduce custom protobuf flag for sensitive and required fields and parse them from .proto file.

gzigzigzeo:
Okay, I'll try to reproduce by starting with the fresh git clone then.

gzigzigzeo:
@Joerger Could you please clarify how make test is failing?

terraform/Makefile
```diff
@@ -30,40 +30,10 @@ clean: tfclean
...
-reapply:
-	terraform -chdir=$(TFDIR) apply -var-file="vars.tfvars"
-
 CUSTOM_IMPORTS_TMP_DIR ?= /tmp/protoc-gen-terraform/custom-imports
 
-# Regenerates types_terraform.go
-gen-schema:
-
+.PHONY: gen-tfschema
+gen-tfschema:
```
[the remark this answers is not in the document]
gzigzigzeo: I've added the `protoc-gen-terraform` presence and version validations here.

terraform/example/trusted_cluster.tf
```diff
@@ -1,20 +1,20 @@
 # Teleport trusted cluster
 
-resource "teleport_trusted_cluster" "cluster" {
-  metadata {
-    name = "primary"
-    labels = {
-      test = "yes"
-    }
-  }
+# resource "teleport_trusted_cluster" "cluster" {
```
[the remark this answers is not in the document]
gzigzigzeo: My original idea was to make `make apply` work without any additional setup. Setting up trusted cluster clearly requires additional setup setps, but we still need an example.

terraform/main.go
gzigzigzeo: Yes, they do. I think, you're correct: in case with `go:generate` we should not commit generated files, and need to generate them on the fly before build. This is not the case here: `go generate` is used as a shortcut to run _gen/main.go. For now, I moved this command to `make gen-tfschema`. It might be useful to have separate `make gen-resources` for this files.

Joerger (commented):
> However, I guess that we should do something with it. For instance, we could leave full examples of all resources in the examples folder and introduce minimal-config for make apply command leaving just a resources which do not require any additional/external setup like user, role, token and one of the configs).

I think that's a good idea, the example should be usable with minimal effort, and should only require following the steps in the README.  You could separate it into two separate examples, or you could also just comment out the resources which require setup and give instructions on how to un-comment and set them up. Up to you. 

I removed everything from the example except `user.tf`, `role.tf`, and `provision_token.tf` and got it working. Testing is working for me too.

Once the example and README is updated, LGTM.

terraform/README.md
```diff
@@ -1,56 +1,54 @@
...
 
-```
-go install github.com/gravitational/protoc-gen-terraform
-make gen-schema
-```
+6. Run tests:
 
-Please note that `ProvisionTokenV2.Allow` field is defined lowercase in `.proto` file (`repeated TokenRule allow = 2 [ (gogoproto.jsontag) = "allow,omitempty" ];`). You need to manually patch `types_tfschema.go`, set `Name: "Allow"` in provision token metadata struct (`GenSchemaMetaProvisionTokenV2()`) until it is fixed.
+    ```bash
+    make test
```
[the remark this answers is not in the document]
gzigzigzeo: That's weird. Testing framework supposed to be self sustained: it downloads required binaries and should work without license key. I'll try to do tests from scratch tomorrow. I possibly forgot to add feature flag for enterprise entities. `TESTING.md` exists in the root folder (testing framework is shared between plugins) - however, it might be a bit outdated, I'll take a look.

terraform/README.md
```diff
@@ -1,56 +1,54 @@
...
 
-2. Install the plugin to Teleport:
+1. Install [`protobuf`](https://grpc.io/docs/protoc-installation/).
+2. Install [`protoc-gen-terraform`](https://github.com/gravitational/protoc-gen-terraform) v1.0.0+.
 
-```bash
-cd teleport-plugins/terraform
-make install
-```
+    ```go install github.com/gravitational/protoc-gen-terraform@a63aa54956b6bbbcdac039d2f54261bae12d19e8```
```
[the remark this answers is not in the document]
gzigzigzeo: Yep, I hard coded the revision here to ensure it will work on your side.

gzigzigzeo:
> I think this is a good idea, can we do this for all example resources which require setup and add setup instructions to the README?

@Joerger I am not sure that adding setup instructions here is a good idea: for example, the setup flow for SAML connector is quite long, and I suppose that if a user needs it - they know what they're doing. I could add links to corresponding setup docs on the website, target .yaml files for tctl are not that different from .tf's (in fact, attribute names are the same, only format differs).

> the README. You could separate it into two separate examples, or you could also just comment out the resources which require setup and give instructions on how to un-comment and set them up. Up to you.

Let me separate examples then. It might be handy just to copy-paste files between two folders. 

Will get back with the updated examples tomorrow.

terraform/README.md
```diff
@@ -1,56 +1,91 @@
...
+
+    Please note that target identity file was exported to `/tmp/terraform-identity` on previous step. If you used another location, please change in in `main.tf`.
+
+1. Create sample resources:
+
+    ```
+    cp examples/user.tf.example examples/user.tf
+    cp examples/role.tf.example examples/role.tf
+    cp examples/provision_token.tf.example examples/provision_token.tf
+    ```
```
Joerger, line 76:
```suggestion
    ```
    cp example/user.tf.example example/user.tf
    cp example/role.tf.example example/role.tf
    cp example/provision_token.tf.example example/provision_token.tf
    ```
```

terraform/example/main.tf.example
```diff
@@ -1,19 +1,20 @@
-# Example Terraform Provider configuration
-
 variable "identity_file_path" {}
 variable "addr" {}
```
Joerger, line 2: These are getting overwritten below, I'd just remove these 2 lines

terraform/README.md
```diff
@@ -1,56 +1,91 @@
...
+
+    ```
+    make apply
+    ```
+
+1. Make changes to .tf files and run:
+
+    ```
+    make reapply
+    ```
```
Joerger, line 90: I think a clean up step would be nice here

```suggestion
1. Make changes to .tf files and run:

    ```
    make reapply
    ```

1. Clean up:

    ```
    make destroy
    ```
```

terraform/Makefile
```diff
@@ -84,6 +61,36 @@ gen-schema:
...
+endif
+	TF_ACC=true go test ./test -v
+
+.PHONY: apply
+apply: install
+	terraform -chdir=$(TFDIR) init && terraform -chdir=$(TFDIR) apply -auto-approve
+
+.PHONY: reapply
+reapply:
+	terraform -chdir=$(TFDIR) apply
```
Joerger, line 96:
```suggestion
.PHONY: reapply
reapply:
	terraform -chdir=$(TFDIR) apply
	
.PHONY: destroy
destroy:
	terraform -chdir=$(TFDIR) destroy -auto-approve
```

terraform/README.md
```diff
@@ -1,56 +1,91 @@
 # Terraform Provider Plugin
 
-# Installation
+## Usage
 
-1. Clone the plugin:
+Please, refer to [GETTING_STARTED guide](GETTING_STARTED.md) and [official documentation](https://goteleport.com/docs/setup/guides/terraform-provider/).
```
Joerger, line 5: I would be in favor of dropping the getting started guide and just referencing the official documentation instead. They share the same information, but keeping both is just additional upkeep. Up to you.

terraform/README.md
```diff
@@ -1,56 +1,91 @@
...
+2. Install [`protoc-gen-terraform`](https://github.com/gravitational/protoc-gen-terraform) v1.0.0+.
 
-```bash
-cd teleport-plugins/terraform
-make install
-```
+    ```go install github.com/gravitational/protoc-gen-terraform@latest```
 
-3. Configure teleport:
+    _NOTE_: Once PR is merged, we'll replace SHA with v1.0.0
```
Joerger, line 14: `@latest` is sufficient for now
```suggestion
```

Joerger (commented):
`gen-tfschema` is generating a broken schema because the wrong dependency versions are being used. Can you update them to the following and re-run go mod tidy?
 - github.com/golang/protobuf v1.5.0
 - google.golang.org/protobuf v1.27.1

Once that's fixed the PR should be good to merge!

terraform/README.md
gzigzigzeo: We have not merged protoc-gen-terraform yet (

gzigzigzeo:
@Joerger Did the upgrade, could you please check on your side? That's not clear for me why on my local machine it worked and working okay.

klizhentas:
@Joerger @r0mant @russjones folks, can we merge this now? What are our release plans?

PR #440, opened 2022-03-04
teleport plugins release process

Joerger:
- Create an RFD to track the teleport-plugins release process and add steps which are currently missed.
- Add a new `update-api-version` make target to update the Teleport API dependency
- Update `update-version` make target to call `update-api-version`


rfd/0002-teleport-plugins-versioning.md
```diff
@@ -0,0 +1,83 @@
...
+Teleport Plugins release to ensure compatibility with its corresponding Teleport
+version. This can be done with `go get github.com/gravitational/teleport/api@vX.Y.Z`,
+which can be added to the `make update-version` target.
+
+However, this may lead to errors if significant changes have been made to the API,
+and will require some manual work to resolve. 
+
+To avoid delaying releases, a PR should be made as soon as `teleport@vX.Y.Z-beta.1`
+is released to update the Plugins version to `vX.Y.Z-beta.1`. Any issues can then 
+be resolved well before `teleport@vX.Y.Z-beta.1` is released, and updated afterwards.
```
r0mant, line 49: Did you mean to write "before `teleport@vX.Y.Z` is released"?

Makefile
```diff
@@ -80,8 +80,21 @@ releases: release/access-slack release/access-jira release/access-mattermost rel
 .PHONY: build-all
 build-all: access-slack access-jira access-mattermost access-pagerduty access-gitlab access-email terraform event-handler
 
+.PHONY: release-version
+release-version: update-api-version update-version terraform/gen-schema update-tag
```
r0mant, line 84: I'm not sure `update-tag` should be a part of this target though? I usually do it like this:
- Call `make update-version VERSION=vX.Y.Z`  to update versions, push to Github, open a PR.
- Once the PR gets merged, update local master and only then call `make update-tag VERSION=vX.Y.Z`.

If we bundle them together, it will tag the branch and start building it before the actual PR with the changes gets merged.

Makefile
```diff
@@ -80,8 +80,25 @@ releases: release/access-slack release/access-jira release/access-mattermost rel
...
+	# go get github.com/gravitational/telepeport/api@v0.0.0-timestamp-6_byte_hex_commit_hash.
+	go get github.com/gravitational/telepeport/api@v$(VERSION)
+	go mod tidy
+
+	# Once the API version is updated, the terraform schema
+	# must be regenerated with the up to date .proto files.
+	# If significant changes have been made to the grpc API
+	# then the terraform/gen_teleport.yaml file will need to
+	# be updated manually.
+	$(MAKE) -C terraform gen-schema
```
r0mant, line 98: Does this only make sense in the context of `update-api-version` target, or would it make sense to have a separate target for it which `update-version` would depend on? Isn't this how you had it here before?

r0mant (commented):
lgtm to me in general but let's get a 2nd pair of eyes from e.g. @zmb3

Makefile
Joerger: I think it makes more sense in the context of `update-api-version`, since that's when the `.proto` file dependencies are updated. I don't think you'd ever want to do one without the other, which is why I moved it here.

tcsc: Would this be invoked like
```shell
$ make update-api-version VERSION=8.3.0
```
... or would we expect `VERSION` to be  set in an other makefile, or build environment?

tcsc (approved):

Makefile
Joerger: The former, because it will primarily be called when you run `make update-version VERSION=8.3.0`. That's how it's being done currently, though we could add `VERSION` to the makefile.

rfd/0003-teleport-plugins-versioning.md
```diff
@@ -0,0 +1,83 @@
...
+the same compatibility guidelines above.
+
+### Release process
+
+The full release process is contained in `make release-version VERSION=vX.Y.Z`.
+This will update the version files of each plugin, update the `github/gravitational/teleport/api`
+dependency, regenerated the Terraform Schema, and update the git tags.
+
+Some basic manual changes may need to be made to handle significant changes
+made to the Teleport API, as explained below.
```
Joerger, line 35: We also need to update `TELEPORT_GET_VERSION` in integration tests, as well as `lib/testing/integration/download.downloadVersions`, at least for major versions.

rfd/0003-teleport-plugins-versioning.md
```diff
@@ -0,0 +1,83 @@
...
+Teleport Plugins release to ensure compatibility with its corresponding Teleport
+version. This can be done with `go get github.com/gravitational/teleport/api@vX.Y.Z`,
+which can be added to the `make update-version` target.
+
+However, this may lead to errors if significant changes have been made to the API,
+and will require some manual work to resolve. 
+
+To avoid delaying releases, a PR should be made as soon as `teleport@vX.Y.Z-beta.1`
+is released to update the Plugins version to `vX.Y.Z-beta.1`. Any issues can then 
+be resolved well before `teleport@vX.Y.Z` is released, and updated afterwards.
```
marcoandredinis, line 49: Currently, a new Plugin's release brings a lot of changes directly into master
Those changes were not tested in the the Plugin's context
So, when we add them, things can (and do) break

As an example, there's a terraform test for the SAML connector
The teleport's validation was changed and now there's a [required field](https://github.com/gravitational/teleport/commit/734362909bd28717433ebbb5f1c8571262b6e898#diff-9ca35f158737d192d68963280080325252f5321cba3adb494a7c2b2fc61c6fa7)
(It's not my goal to discuss the change itself, but the fact that we now require a certain field)
When we upgrade the teleport version used in tests (currently [9.0.4](https://github.com/gravitational/teleport-plugins/blob/69b532777478cbb2ad3356be21e3bf2337a8bb0f/lib/testing/integration/download.go#L50)) this test will break

The fact that we had a breaking change means we:
- need manual work when running the release script
- need to have context on the failing change, in order to fix it

We can solve this in a more holistic way, by requiring PRs in teleport to have the Plugin's state into consideration - using a monorepo
Continuing the example, the PR above would only be accepted if the author changes the terraform test as well
The downside is that this will add extra burden when developing a new Teleport's feature
As a side effect, we would only have a single release point in time, and backports would be "free"

I'm fine with your suggested approach, because I understand the negative impacts the monorepo approach will have in the long run.
Nonetheless, have we considered merging the `teleport-plugins` repo into the `teleport` one?

Joerger: > Currently, a new Plugin's release brings a lot of changes directly into master Those changes were not tested in the the Plugin's context So, when we add them, things can (and do) break

On the flipside, we've had things break from not updating the API module, terraform schema, etc. See https://github.com/gravitational/teleport-plugins/pull/431 for example, which is what led me to start this RFD.

> The fact that we had a breaking change means we:
> 
> * need manual work when running the release script
> * need to have context on the failing change, in order to fix it
> 
> We can solve this in a more holistic way, by requiring PRs in teleport to have the Plugin's state into consideration - using a monorepo Continuing the example, the PR above would only be accepted if the author changes the terraform test as well The downside is that this will add extra burden when developing a new Teleport's feature As a side effect, we would only have a single release point in time, and backports would be "free"
>
> I'm fine with your suggested approach, because I understand the negative impacts the monorepo approach will have in the long run. Nonetheless, have we considered merging the `teleport-plugins` repo into the `teleport` one?

I think that expanding the Teleport repo to include more submodules similar to `teleport/api` could be useful and has not yet been explored. I agree that this would be a much more holistic and effective approach for us at Teleport. However it diminishes part of the idea behind the plugins, which is to make it an "independent" project which serves as an example to customers who want to develop their own plugins. But if we value the teleport-plugins as a product more than the api at large, then this may be a worthwhile approach.

I think another way to approach this issue holistically would be to ask how we can make it easier for third-party developers to use the Teleport API more easily, including ourselves. Adding tests to the Teleport Repo which checks for compatibility with teleport-plugins would solve the problems listed above, but would not solve all of the possible issues for users attempting to create and maintain their own custom Teleport plugin. Instead we could make the API more durable to breaking changes in Teleport, and add tests to verify this durability. 

These proposed durability changes could include things like:

#### Refactoring our gRPC service / protobuf files.

Right now our proto files are hard to import directly and are reliant on an abandoned tool [gogo-proto](https://github.com/gogo/protobuf) to compile. This leads to difficulties with external tools like our terraform provider which rely directly on the protobuf schema.

We may be able to  use [Buf](https://docs.buf.build/introduction) to alleviate this issue, though I don't know much about it ( cc @codingllama if you have anything to add, or correct, about Buf). Buf would make it easy for users to directly import our protobuf schema, check for breaking changes when updating, etc. We could also use this to expand the API to other languages and develop an SDK.

#### Reorganizing the API module into separate modules

Currently we export types, constants, gRPC types, other functions, etc., but the only thing which we protect with our versioning guidelines is the gRPC service itself, and the client methods in `teleport/api/client.go`. As such, it would make more sense to move the client and gRPC service into their own module(s) (`teleport/client`), and move the types, constants, defaults, credential logic, etc. in to a separate module (`teleport/api` or `teleport/external`?) which the new client module would depend on. This way, when a user upgrades `teleport/client` they won't have to worry about any changes to types etc that may have occurred (ex: types.RoleV4 -> types.RoleV5).

#### Deprecation warnings

We could provide the user with descriptive deprecation warnings so that they aren't surprised when something is no longer working. For example, rather than making that sso field a required field without any explanation, we could have done it with a deprecation step for one major version saying `Starting in Teleport v11.0.0, this field will be required, please add it in your resource spec`.

#### Improving test coverage

Right now we don't have any automatic tests in place to guarantee that the API follows the version compatibility guidelines. For example, our recent v9.3.4 release actually did not have an httpfallback in place for `GetAccessRequests` which underwent some changes. We could set up tests to ensure that every API request performs properly with the previous, current, and next major version of Teleport. In this case we could have a test for v8.0.0, v9.0.0, and v10.0.0. Once we have this setup I imagine it'd be very easy to add additional tests for new/updated endpoints. We also just need this for out Teleport automatic tests in general.

#### Conclusion

In the short term I think we should at least go forward with formalizing this release process, which is essentially the current one which often requires manual work. This will at least put the onus of breaking changes on us at Teleport rather than users of teleport-plugins.

In the long term, I think we should invest more resources into developing an API which is widely usable, easy to upgrade, and extensible to more use cases. 

@zmb3 Any thoughts on this?

Joerger: Yeah you're right that the major issues we're encountering are only with the Terraform provider, and after looking at the K8s Operator, I do agree that we could/should move the Terraform provider into the main repo. Thanks for taking the time to read my ramblings and return it to a concrete solution :+1:

Joerger:
Closing since this is stale

Issue #235, opened 2021-05-12
Publish Terraform Provider


benarent:
To make setup easier, we should publish our Terraform provider. https://www.terraform.io/docs/registry/providers/publishing.html 

This will require use to move it to a new repo. `terraform-provider-teleport` 

Other steps: 
- Making a release
- Signing release.

benarent:
I chatted with @klizhentas about this on Monday. We wont' publish to Terrafrom, but should compile and publish the module. Terraform provides a few options for private modules https://www.terraform.io/docs/language/modules/sources.html

webvictim:
I think those docs relate to modules rather than providers - I don't think there's a way to use a provider other than:
1. compile (or extract) the binary for the provider locally to `~/.terraform.d/plugins/gravitational.com/teleport/teleport/6.2.0/linux_amd64` or similar
2. pull it from Terraform registry

What's the incentive for not publishing it to the registry?

ari-becker:
@klizhentas @webvictim any updates?
I see that despite being added to the `Teleport plugins 7.2` milestone that this was not completed in time for https://github.com/gravitational/teleport-plugins/releases/tag/v7.2.0
Publishing the provider to registry.terraform.io or a hypothetical `terraform-registry.goteleport.com` that obeys the [Terraform Provider Registry protocol](https://www.terraform.io/docs/internals/provider-registry-protocol.html) would make my life much, much easier and make plugin upgrades much easier to handle.

sover02:
This came up today internally over here in elastic-land as well :)

dmsergeevN26:
A new major version is out and the provider is still not published :(

pzduniak:
What is the reasoning? Why do we, the community, have to spend time scripting cross-platform hooks that imitate the proper process, just because someone won't pull the trigger? This project is 90% completed, someone probably already called it done - yet it's still just missing the last, critical part.

tcsc:
A quick update on our progress.

As part of an internal supply-chain hardening effort we have decided to host our own custom Terraform Provider Registry and publish there, rather than distribute it via `registry.terraform.io`. We're currently building it out and integrating it into our release pipeline. 

The horizon for this is currently in the order of weeks (as opposed to days, months, etc), and we'll let you know here if anything changes.

benarent:
@tcsc , we have an expected timeline? It looks like most people are happy with the RFD now https://github.com/gravitational/teleport-plugins/pull/444

[...]

PR #444, opened 2022-03-07
Adds RFD-0002 - Custom Terraform Registry

tcsc:
The task of supporting the creation of a custom Terraform registry has snowballed somewhat, and so (at @r0mant's suggestion) I've put together a small RFD to make sure that everyone is on the same page.

This is very definitely a first pass, and the _discussion_ part of the RFD process is highly welcomed.

This contributes to https://github.com/gravitational/teleport-plugins/issues/235.


rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,185 @@
...
+
+1. Read access to a signing key, to sign the package (see note below)
+2. Write access to the AWS bucket storing the signed packages for distribution
+
+_**Note:**_ We currently expect to use the RPM signing key as the Terraform
+signing key as it already exists and is already used for a similar purpose.
+That said, there is no technical reason why it _must_ be the RPM signing
+key. As long as the private key used to sign the package agrees with the a
+public key exposed via the registry, the signature should be deemed valid
+by Terraform.
```
wadells, line 157: I'd like to move away from using a monolithic gpg key for signing artifiacts.  Instead, I propose we used gpg subkeys.

The setup:

1) we have a root-of-trust key signing gpg key, preferably with the private key offline, or in our highest security online storage.  This private key never touches build infra
2) We use the root-of-trust key to sign subkeys for each different artifact type (RPM, APT, TF).  These subkeys are independently exported, and used in the online build process, only for that artifact type.
3) In the even of build infra compromise, we can revoke and reissue subkeys, without having to rotatate the entire root of trust.

The sticky situation we're in currently is:

* The current root of trust isn't all that trustworthy, as it was _probably not but maybe we should treat it like it was_ compromised in https://goteleport.com/blog/hack-via-pull-request/.  There is no evidence that anyone but me attacked this, but nor can I prove the key wasn't exfiltrated via an advanced persistent threat type actor.
* Furthermore, that private key has been on every builder, and a half a dozen employees workstations/secret stores.
* Rotating keys will require customer action.  Either in a:
  * We changed keys on the official, one repo, and broke your current setup.  => "Why did things stop working?" type break.
  * We published a new repo, and the old one will stop getting updates -- better update your config.  => "Why did things fall out of date?" type break.

Anyhow, this is another topic where we probably need an RFD for the plan going forward.  Not something to change yet in this RFD, but something to be aware may change in the future.

cc @reedloden

rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,185 @@
...
+are produced.
+
+_This_ is the major open question in this RFD - _how will our promotion
+process communicate to our custom registry that a new release exists?_
+
+#### Options
+
+1. Fully manual process
+2. Promotion process generates an index file that can be used by the
+   registry in `gravitational/next`, either by
```
wadells, line 181: I like this solution as a "good enough" solution.

Upsides:
* No need to host extra logic in next/somewhere else.
* Consistent with @fheinecke's plans in https://github.com/gravitational/teleport/pull/10746 and our other repos

Downsides:

* I haven't looked at the TF index format yet.  If it explicitly has a "latest" or other floating tags, we'll end up in a https://github.com/gravitational/cloud/issues/958 situation
* Concurrency concerns -- we don't want two builds to be rebuilding/uploading an index at the same time.  We need some sort of lease/mutex to do this properly.  That said, we've gotten pretty far with only drone's concurrency controls.

wadells (commented):
Thanks for putting this together!

Overall, I'd lean towards "lets not run an extra service or code in next". Can we just generate the metadata at promotion time? However I also recognize this path has been paved with tears for apt.  It really comes down to "What does the hashicorp TF plugin index file look like, and how do clients (`terraform`) consume that?".  If the file is a dumb index (e.g. no "latest", an unordered list), and clients do the version selection, we're probably ok.  If clients implicitly rely on e.g. the first item in a list or some sort of latest metadata, we may have more trouble.

rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,185 @@
...
+ 2. We want to maintain close control of the distribution of our software.
+
+With these two goals in mind, the most practical solution for distributing
+the Teleport Terraform provider appears to be hosting a custom Terraform
+registry.
+
+### Why not use the public Hashicorp registry?
+
+This conflicts with goal #2 (i.e. maintaining close control of the
+distribution of our software).
```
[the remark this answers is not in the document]
benarent: I've this open ticket for Docker https://github.com/gravitational/teleport/issues/4159

tcsc:
> Can we just generate the metadata at promotion time? 

Initially I thought the registry needed to be smart, but on reflection there is no actual _reason_ why this should be so. It should be possible to generate an index and all of the version entries at promotion time, arrange them in an appropriate directpory tree and serve them statically.

This is nice as it means deployment reverts to being `s3 sync`  at promotion time.

Will work this idea up and see if it comes to anything.

rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,185 @@
...
+    3. generate the signature file by signing the checksum file
+2. Upload all generated files to the AWS bucket that backs the registry
+3. Update the registry index to include the newly released providers (see
+   _Populating the registry_, below)
+
+I expect this to be implemented as a small Go program, rather than a shell
+script. Using small Go programs for build tasks has shown itself to be a
+useful, flexible and above all _legible_ tool for scripting build tasks in
+the mainline Teleport repository, and we should continue that pattern in
+`teleport-plugins`.
```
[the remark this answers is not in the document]
logand22: Does the mainline teleport repository use any tools to run these small go programs, see [Mage](https://magefile.org/), or does it run it by simply running `go run task.go`

benarent:
As part of the registry can I request that we also provide beta builds? I automate a lot of my cluster creation, and it would be awesome to have the latest teleport terraform provider for setup.

rfd/0002-custom-terraform-registry.md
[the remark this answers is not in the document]
tcsc: `go run`.

tcsc:
> Can we just generate the metadata at promotion time?

@wadells - I'm now 98% sure we can do this. I have an proof-of-concept where I can generate the metadata in a batch and serve it statically from an S3 bucket via cloudfront.

That last 2% uncertainty is only because I haven't got terraform to 100% install from the registry yet, but that looks like a signing key problem and not a file hosting problem: 

Example terraform
```terraform
terraform {
    required_providers {
      teleport = {
        source  = "terraform.clarke.mobi/gravitational/teleport"
        version = "~> 8.3"
      }
  }
}

```

Example `terraform` output:
```
> TF_LOG=DEBUG terraform init
 # ... skipped irrelevant debug logging... 
Initializing provider plugins...
- Finding terraform.clarke.mobi/gravitational/teleport versions matching "~> 8.3"...
2022-03-09T17:02:30.852+1100 [DEBUG] Service discovery for terraform.clarke.mobi at https://terraform.clarke.mobi/.well-known/terraform.json
2022-03-09T17:02:31.446+1100 [DEBUG] GET https://terraform.clarke.mobi/index/gravitational/teleport/versions
2022-03-09T17:02:32.708+1100 [DEBUG] GET https://terraform.clarke.mobi/index/gravitational/teleport/8.3.4/download/linux/amd64
2022-03-09T17:02:33.918+1100 [DEBUG] GET https://terraform.clarke.mobi/store/terraform-provider-teleport-v8.3.4-linux-amd64-bin.zip.sums
2022-03-09T17:02:34.773+1100 [DEBUG] GET https://terraform.clarke.mobi/store/terraform-provider-teleport-v8.3.4-linux-amd64-bin.zip.sums.sig
- Installing terraform.clarke.mobi/gravitational/teleport v8.3.4...
╷
│ Error: Failed to install provider
│ 
│ Error while installing terraform.clarke.mobi/gravitational/teleport v8.3.4: error decoding signing key: openpgp: invalid data: entity without any identities
╵
```

reedloden:
Are we sure that private terraform registries will work for our customers?

I see this on https://www.terraform.io/cloud-docs/registry/using

> Note: Terraform Enterprise does not support private providers.

I would hate for us to do a bunch of work and then find out that our customers can't even use it.

tcsc:
> Note: Terraform Enterprise does not support private providers.

It's unfortunate that _private_ providers is an overloaded term in the Terraform docs. It could mean:

1. A provider that is distributed anywhere other than the public Hashicorp registry
2. A provider that is distributed from a Terraform Cloud Private Registry
3. A provider that is only available on a private network via a Custom Registry

I _think_, from reading the [Private Registry Docs](https://www.terraform.io/registry/private), that the note above refers to option (2), as it requires client authentication that the Terraform Enterprise service will not have access to.

That said; this is probably cheap to confirm and expensive to get wrong, so it definitely needs checking out.

rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,185 @@
...
+1. A web service that exposes the registry index,
+2. The binary packages referenced by the registry must be packaged in zipfiles,
+   as opposed to the compressed tarballs our current build process produces, and
+3. The checksum files must be signed with a GPG key
+
+### Serving the registry and packages
+
+#### Registry Hosting
+
+The prototype registry (under development) is currently set up to be served as
```
r0mant, line 79: Any reason we can't just host the registry index file off of S3 as well and update it as a part of the Drone publishing pipeline?

rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,185 @@
...
+That said, there is no technical reason why it _must_ be the RPM signing
+key. As long as the private key used to sign the package agrees with the a
+public key exposed via the registry, the signature should be deemed valid
+by Terraform.
+
+### Populating the registry
+
+Populating the registry is a two-part problem:
+
+First, we must create a new, back-filled registry containing all of the
```
r0mant, line 163: Do we need all previously published providers? Maybe publish latest v8, latest v9 and then keep publishing new versions as we're releasing them?

rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,185 @@
...
+the mainline Teleport repository, and we should continue that pattern in
+`teleport-plugins`.
+
+#### Build Security issues
+
+This process will require that the `teleport-plugins` promotion process will
+need new secrets:
+
+1. Read access to a signing key, to sign the package (see note below)
+2. Write access to the AWS bucket storing the signed packages for distribution
```
r0mant, line 150: Talking with @russjones, one of invariants we want to maintain is no secrets in CI. Does our Drone instance run in AWS? Can we use IAM roles instead of secrets?

rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,185 @@
...
+3. Update the registry index to include the newly released providers (see
+   _Populating the registry_, below)
+
+I expect this to be implemented as a small Go program, rather than a shell
+script. Using small Go programs for build tasks has shown itself to be a
+useful, flexible and above all _legible_ tool for scripting build tasks in
+the mainline Teleport repository, and we should continue that pattern in
+`teleport-plugins`.
+
+#### Build Security issues
```
r0mant, line 144: Can you explicitly call out permissions the pipeline will need on the buckets. For example, it should be able to write assets but not delete/modify anything. Also, enabling some sort of monitoring on S3 buckets (CloudWatch) may be useful too.

rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,185 @@
...
+
+This appears to be pretty straightforward, as I already have a proof-of-
+concept program that scans the release plugins AWS bucket and repacks all
+of the available releases, generating a TypeScript dictionary that can be
+embedded directly in the front-end registry code.
+
+Secondly, we need a mechanism to _maintain_ the registry as new releases
+are produced.
+
+_This_ is the major open question in this RFD - _how will our promotion
```
r0mant, line 174: Can the build/promote pipeline in teleport-plugins' [drone.yaml](https://github.com/gravitational/teleport-plugins/blob/master/.drone.yml) do it? So the process would look like:

- A new teleport-plugins tag is created.
- In addition to building all plugins like now, Drone builds terraform plugin artifacts in required format and uploads them to S3.
- When terraform-plugin artifact build is promoted, Drone regenerates index file and uploads it to S3 also.

Any reason something like this wouldn't work?

rfd/0002-custom-terraform-registry.md
wadells: > invariants we want to maintain (or maybe rather, introduce and maintain) is no secrets in CI

We don't use Drone for teleport-plugins Continuous Integration anymore. We use GCB for CI and Drone for Release/CD. We do use Drone (with AWS secrets) for release processes, most of which this RFD is concerned with.

This invariant will hold -- no secrets in CI, because the secrets needed will be for release/CD.

> Does our Drone instance run in AWS?

It runs in GCP GKE -- in the `drone-gravitational` project.

> Can we use IAM roles instead of secrets?

We're still a way away from this.  We've got the right infra in AWS -- with well defined principle-of-least-privilege policies that can bet attached to roles/service users.  However the bootstrap to get a short lived secret/token ahead of any user controllable processes isn't available yet.

I'd be cautious about coupling the reslease secrets work with this terraform RFD. There is at least a RFD and weeks to months of work (e.g. machine ID for CD) to improve the state of the art for release secrets. IMO we don't want to slip on Q1 time to value work (this) for that security project that may make more sense to look at after machine ID is available and we look at using one of our dogfooding teleport clusters as an aws intermediary for CD.

rfd/0002-custom-terraform-registry.md
wadells: https://github.com/gravitational/cloud/blob/master/rfd/0017-artifact-storage-standards.md#gold-1 should cover most of this.  For examples, you can grab some policy/terraform from https://github.com/gravitational/cloud/blob/master/rfd/assets/0017-artifact-storage-standards/artifact-bucket.tf

rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,185 @@
...
+    Terraform, we want to supply them with a Terraform provider, and a method
+    for installing it that is natural to the Terraform ecosystem.
+
+ 2. We want to maintain close control of the distribution of our software.
+
+With these two goals in mind, the most practical solution for distributing
+the Teleport Terraform provider appears to be hosting a custom Terraform
+registry.
+
+### Why not use the public Hashicorp registry?
```
knisbet, line 35: You may want / need to check in with potential customers who are looking for this on whether they would accept this or not. I got from @FrankWeissmann on a side discussion for something else today, and this came up, that a customer is really pushing for this because they have a policy of not using any third party plugins, so they might see this in a similar light.

tcsc: Having some discussions with @FrankWeissmann and the client involved now. Hopefully we can come to a resolution that works for everyone.

rfd/0002-custom-terraform-registry.md
```diff
@@ -107,7 +161,149 @@ We _could_ modify Houston to serve the required files, but as the service is
...
+Once created, each `download` metadata does not need modification, unless
+a provider object is re-published.
+
+##### 3. The object store (key prefix `/store`)
+
+The object store is where all the release package files are stored. Each
+provider artefact, together with its associated hash & signatures files.
+
+This section of the bucket should be write-only. No objects in this section
+should be modifiable or delete-able.
```
wadells, line 254: Pedantic clarification: `/dev/null` is write only. I think you mean "write once, read many" or the unfortunate acronym WORM.  A humorous site to illustrate the difference: http://www.supersimplestorageservice.com/

Furthermore object versioning on buckets is an all or nothing thing, I don't think we can do WORM on just part of a bucket.

Fortunately, we don't particularly care if artifacts are overwritten when object versioning is on because:

1) we can easily revert to old versions
2) it is logging/SIEM's responsibility to yell at us if this happens.

In some cases (e.g. a release mistake) being able to overwrite (while preserving history) is quite valuable.

rfd/0002-custom-terraform-registry.md
```diff
@@ -107,7 +161,149 @@ We _could_ modify Houston to serve the required files, but as the service is
...
+The release tarball(s) will be injected into the build by being downloaded
+from the staging bucket, based on the version tag supplied during the
+promotion process.
+
+##### Terraform Plugin API Version
+
+In order to correctly index the providers, the registry need to know what
+version of the Terraform Plugin API a given provider supports. This is not
+something we can deduce from the provider itself, and must be "just known"
+in advance.
```
wadells, line 278: What do these versions look like and where do they come from?  Is it possible that info is encoded in the go.mod or makefile somehow, and our release automation could derive it from there, such that it is always correct with no update required (outside of bumping the go mod or whatever).

rfd/0002-custom-terraform-registry.md
```diff
@@ -107,7 +161,149 @@ We _could_ modify Houston to serve the required files, but as the service is
...
+For example (from my PoC):
+
+```shell
+Initializing provider plugins...
+- Finding terraform.clarke.mobi/gravitational/teleport versions matching "~> 8.3"...
+- Installing terraform.clarke.mobi/gravitational/teleport v8.3.4...
+- Installed terraform.clarke.mobi/gravitational/teleport v8.3.4 (self-signed, key ID 7DA6C64E1701F9E4)
+```
+
+My proof-of-concept cose expects the signing key in ASCII-armor format, but I
```
wadells, line 303: "cose"?

rfd/0002-custom-terraform-registry.md
```diff
@@ -118,33 +314,60 @@ provider consists of:
...
+
+* There is no way to push an object to a bucket only if an object with the same
+  key does not already exist, which means we can't really implement a race free
+  lockfile-like protocol
+* There are no compare-and-swap style put primitives
+* While versioning _is_ supported, there is no way to say "only overwrite object
+  _X_ if its current version is _Y_".
+
+Our initial approach may have to be relying on the build serialisation tools
+provided by Drone.
```
wadells, line 365: This is fine.  We'll get to chubby at some point in the future :laugh:.  AFAIK we haven't had any races during promotion yet.  They've all been earlier in the process.

wadells (approved):
LGTM!

rfd/0002-custom-terraform-registry.md
tcsc: Nope. You apparantley just gotta know it.

rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,485 @@
...
+Although the Provider Registry Specification is written in terms of a web
+service that will dynamically generate the appropriate responses, it is
+_entirely possible_ to generate a registry index as a simple directory tree and
+then statically serve those files to a client. This approach dramatically
+simplifies the process of applying updates to the registry, and is therefore
+the approach we will consider here.
+
+#### Why can't we just use Houston?
+
+The existing Houston distribution tool at [get.gravitational.com](https://get.gravitational.com/)
```
klizhentas, line 154: @tcsc @pierrebeaucamp Houston does not support that, but the design below looks very similar to our CDN project, any reason we can't serve this asset from our CDN and support additional URL format to support terraform requirements?

klizhentas: @tcsc I also did not realize that you've mostly implemented this RFD here:

https://github.com/gravitational/teleport-plugins/pull/457/files

I would not block us on consolidating it right now, but may be merge this code at some point in the CDN in the future version for unification?

rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,485 @@
...
+
+In selecting the name for the production registry, there is a tension between
+having (arguably) the most memorable and obvious name for the registry (but
+having it's configuration split of several hosts), versus and having a registry
+with all of its configuration and data in one place.
+
+How we name the production registry ultimately comes down to deciding which
+side of this tension we care more about.
+
+If we want people to reference our provider as `goteleport.com/gravitational/teleport`,
```
logand22, line 204: On that note though, how would we disambiguate a docker image and a terraform provider that would presumably share the same naming scheme.

Or potentially helm charts. I personally think `terraform.releases.teleport.dev/gravitational/teleport` is a very lengthy name, but `goteleport.com/gravitational/teleport` could potentially have multiple artifacts vying for that spot. 

Thoughts on `terraform.goteleport.com/gravitational/teleport`?

tcsc: I'd be happy with that, certainly. I've only suggested the longer `terraform.releases.teleport.dev` as it matches the existing naming scheme. I'll add it to the list.

tcsc: Updated

rfd/0002-custom-terraform-registry.md
tcsc: Wording updated

rfd/0002-custom-terraform-registry.md
tcsc: Longer version: Ultimately the API version numbers come from some constants buried deep inside the the transitive dependencies of the Terraform plugin framework. Terraform uses GRPC as the basis of it's plugin API, so they are  probably originally specified in a protobuf file somewhere.

All of the the types and code that Terraform uses to load and validate plugins are inside `internal` packages used by the Terraform command-line app, and so are not easily available for us to just pick up and use. There _may_ be a way to do it by subverting the testing hashicorp-supplied Provider testing framework, but that's getting really dodgy. 

At the moment no automated solution is looking less bad than manually making sure the dronefile injects the right versions into the promotion code.

gzigzigzeo: I'd add that those API version upgrades are really rare and always backwards-compatible, as far as I understood. We even might want to stick with an older version for a while when this upgrade happens, to make things work with older Terraform versions.

rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,512 @@
...
+
+Upload the repackaged `zip` files, the updated `versions` index and the new
+`download` files to the S3 bucket.
+
+#### Synchronization Hazard
+
+The Download-Update-Publish cycle for updating the registry index is vulnerable
+to corruption if multiple independent updates occur simultaneously.
+
+Unfortunately, AWS S3 does not provide any built-in methods for preventing this
```
gzigzigzeo, line 396: I might be wrong, but as far as I remember, S3 has simple object locks with/without timeouts. For sure, they can't replace CAS, but might be helpful to fail fast and retry if an another conflicting build is in progress.

rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,512 @@
...
+include the Teleport provider in their terraform scripts:
+
+1) the registry host; current options are:
+   * `goteleport.com`
+   * `terraform.releases.teleport.dev` (i.e. matching existing distribution site naming scheme)
+   * `terraform.goteleport.com` (something in-between)
+   * ... something else entirely?
+2) the provider namespace; options are `gravitational` or `teleport`,
+3) the provider name; the obvious choice is `teleport`, but mentioned for
+   completeness
```
gzigzigzeo, line 468: I vote for the shorter domain (easier to remember), and "gravitational" namespace in case we decide to add a provider for something other than Teleport itself.

rfd/0002-custom-terraform-registry.md
tcsc: I looked in to AWS locks, and I don't think they're suitable for our purposes. Will take a deeper look and see.

rfd/0002-custom-terraform-registry.md
klizhentas: We have helm charts and other assets are already on teleport.dev (.dev is operated by google).
https://charts.releases.teleport.dev/[](https://charts.releases.teleport.dev/)

Why not [teleport.dev/terraform](http://teleport.dev/terraform)
Either way, it can be goteleport.com, but in this case we should move helm charts and other assets to goteleport.com.I think all our assets should be distributed from one domain.
Do you agree?

klizhentas (changes requested):

rfd/0002-custom-terraform-registry.md
tcsc: Unfortunately `teleport.dev/terraform` wouldn't match the requirements of the Terraform discovery protocol, but we could certainly do something like `terraform.teleport.dev/gravitational/teleport`

rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,512 @@
...
+For more info, see [the Terraform docs](https://www.terraform.io/cloud-docs/registry/using)
+on using custom registries.
+
+### What about name squatting on the Hashicorp registry?
+
+What to do about people publishing homemade Teleport providers to the
+Hashicorp registry is beyond the scope of this RFD.
+
+It's worth noting, though, that _not_ being on the main Hashicorp registry _does_
+make our official solution harder to find.
```
webvictim, line 77: This is one of my biggest issues with us looking to use a custom registry instead of the one that basically everyone else does. People are going to go there, search for Teleport and not find us. If they Google, they probably will find us, but some will definitely think "why on earth aren't they using the main registry?"

rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,512 @@
...
+
+ 1. We want to meet out customers where they are, e.g. if they're using
+    Terraform, we want to supply them with a Terraform provider, and a method
+    for installing it that is natural to the Terraform ecosystem. This is a
+    good practice in general and, in the specific, is part of our Q1 focus on
+    improving our Time-to_first-Value.
+
+ 2. We want to maintain close control of the distribution of our software,
+    including supply chain management and monitoring of publishing, changes,
+    downloads and host of other activities.
```
webvictim, line 33: This doesn't really answer _why_ - it's an answer, of course, but it doesn't actually offer much detail explaining why this solution is explicitly better than the public Terraform registry. Could you maybe add a few examples of situations where this would benefit us? Explain the limitations of the public Terraform registry?

rfd/0002-custom-terraform-registry.md
webvictim: I'd encourage us to stick with the same `terraform.releases.teleport.dev` style as we have for the current repos for the sake of consistency. The full name `terraform.releases.teleport.dev/gravitational/teleport` is long and a little redundant, but I think that for something that users are going to copy/paste once and use for a longer period, it's not going to be a huge issue.

rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,512 @@
...
+1. adding an alias record to our DNS config that points
+   `terraform.staging.teleport.dev` back to the CloudFront CCN, and
+2. configuring CloudFront to use a custom TLS certificate when serving
+  requests for `terraform.staging.teleport.dev`.
+
+##### Production
+
+In selecting the name for the production registry, there is a tension between
+having (arguably) the most memorable and obvious name for the registry (but
+having it's configuration split of several hosts), versus and having a registry
```
webvictim, line 198:
```suggestion
having its configuration split over several hosts), versus and having a registry
```

rfd/0002-custom-terraform-registry.md
```diff
@@ -0,0 +1,512 @@
...
+
+To update the registry, we take the artefacts and data generated in the
+previous step and
+
+1. Download the existing `versions` index file (or create fresh, if
+    not present)
+2. Add entries for the new providers in to the index
+3. Create the `download` entries for each file
+
+I expect this process to be implemented a sa small Go program (or a small suite
```
webvictim, line 380:
```suggestion
I expect this process to be implemented as a small Go program (or a small suite
```

rfd/0002-custom-terraform-registry.md
reedloden: 👍 -- I'm not seeing any significant value yet where having a private registry would be better, especially for the overhead involved.

rfd/0002-custom-terraform-registry.md
tcsc: Biggest issue I can see is that they require OAuth access to our GitHub organization, and want write access to manage webhooks in all our repositories.

reedloden: We can create a separate GitHub organization for this purpose easily enough.

I'll reach out to HashiCorp about this to see why they need such permissions.

webvictim: It seems like a vast overreach in permission terms as well.

klizhentas (approved):

r0mant (approved):

rfd/0002-custom-terraform-registry.md
klizhentas: @webvictim @reedloden @tcsc Trent wrote this RFD following my ask to have all assets that we distribute to be under our control, including Terraform registry. That's primary reason for us hosting private registry, similarly to our parallel efforts of hosting our all assets. The primary reason for that is to protect us from supply chain attacks. Having all assets distributed by teleport CDN is a costly initial investment, but a year ago we made this trade off to make sure we will not be vulnerable to these class of attacks.

I'm open to revisiting this discussion and providing other type of guidance to developers, but let's do this a bit later and outside of this PR, as this is a larger discussion.

logand22 (approved):

rfd/0002-custom-terraform-registry.md
reedloden: Good to know.

Side note, are we tracking https://github.com/gravitational/teleport/issues/4870 in that same vein? We're already offering Homebrew as an option on our [install guide](https://goteleport.com/docs/getting-started/linux-server/#install-a-teleport-client-locally), and [analytics](https://formulae.brew.sh/formula/teleport#default) show it is in wide use.

rfd/0002-custom-terraform-registry.md
```diff
@@ -367,20 +418,102 @@ provided by Drone.
...
+   * `terraform.releases.teleport.dev` (i.e. matching existing distribution site naming scheme)
+   * `terraform.goteleport.com` (something in-between)
+   * ... something else entirely?
+2) the provider namespace; options are `gravitational` or `teleport`,
+3) the provider name; the obvious choice is `teleport`, but mentioned for
+   completeness
+
+In short: which one of these do we want people to use?
+
+1. `terraform.releases.teleport.dev/gravitational/teleport`
```
wadells, line 479: This one.

wadells (approved):

r0mant:
@tcsc Can you bring this up to date and merge please?

Issue #453, opened 2022-03-16
teleport access plugins report an unhelpful version


wadells:
## Summary
Teleport access plugins and teleport-event-handler have different version info:

```
walt@work:~/git/teleport-plugins/access/gitlab/build$ ./teleport-gitlab version
teleport-gitlab v0.2.0 go1.17.5

walt@work:~/git/teleport-plugins/event-handler/build$ ./teleport-event-handler version
Teleport event handler v8.3.1 git:1d26d96de584ebe07b65f891b5811c0520f4e1e5 go1.17.5
```

`v0.2.0` and no hash is next to useless.  We should update these to be consistent.

wadells:
tsh/tctl versions for comparison (these are what we should be consistent with)

```
$ tsh version
Teleport v8.0.6 git:v8.0.6-0-g807c6d662 go1.17.3
$ tctl version
Teleport Enterprise v8.3.3 git:v8.3.3-0-g2c125b019 go1.17.3
```

wadells:
CC @tcsc 

I think it'd be nice to add a version target to the terraform plugin too, using the same format.

I don't know that many folks will use this, since the usual way to check the plugin version is:

```
$ terraform version
Terraform v1.0.5
on linux_amd64
+ provider gravitational.com/teleport/teleport v8.0.7
```

However if we ever have a stray terraform provider binary it'd be nice tobe able to quickly ID it. I haven't found another way to do that via the built in sdk.

PR #454, opened 2022-03-17
Improve access version reporting

logand22:
This PR fixes https://github.com/gravitational/teleport-plugins/issues/453

This change updates the access plugins and event-handlers to have consistent version reporting that matches the format of the `tsh, tctl` commands.

The original method was to generate the `version.go` and `gitref.go` files on release and testing prior to building the binary. This change instead replaces the variables at build time using `ldflags`. The only potential issue would be if `go test` looked for specifically set version variables however from searching this doesn't seem to be the case. 

There doesn't seem to be a semantic difference between the new method of driving the version from `ldflags` and the old generation method. The separate make targets should all behave the same. 

I also add an additional step to the docker build gcb configuration to fetch tags

## Versions

**Access Plugin Email**
```bash
$ docker run -it 146628656107.dkr.ecr.us-west-2.amazonaws.com/teleport/access-plugin-email:9.0.0 version 
teleport-email v9.0.0 git:teleport-email-v9.0.0-1-g0d4c060b-dirty go1.17.8
```

**Access Plugin Gitlab - See [gitlab tag](#gitlab-tag)**
```bash
$ docker run -it 146628656107.dkr.ecr.us-west-2.amazonaws.com/teleport/access-plugin-gitlab:9.0.0 version
teleport-gitlab v9.0.0 git:teleport-gitlab-v0.2.0-177-g0d4c060b-dirty go1.17.8
```

**Access Plugin Jira**
```bash
$ docker run -it 146628656107.dkr.ecr.us-west-2.amazonaws.com/teleport/access-plugin-jira:9.0.0 version
teleport-jira v9.0.0 git:teleport-jira-v9.0.0-1-g0d4c060b-dirty go1.17.8
```

**Access Plugin Mattermost**
```bash
$ docker run -it 146628656107.dkr.ecr.us-west-2.amazonaws.com/teleport/access-plugin-mattermost:9.0.0 version
teleport-mattermost v9.0.0 git:teleport-mattermost-v9.0.0-1-g0d4c060b-dirty go1.17.8
```

**Access Plugin Pagerduty**
```bash
$ docker run -it 146628656107.dkr.ecr.us-west-2.amazonaws.com/teleport/access-plugin-pagerduty:9.0.0 version
teleport-pagerduty v9.0.0 git:teleport-pagerduty-v9.0.0-1-g0d4c060b-dirty go1.17.8
```

**Access Plugin Slack**
```bash
$ docker run -it 146628656107.dkr.ecr.us-west-2.amazonaws.com/teleport/access-plugin-slack:9.0.0 version
teleport-slack v9.0.0 git:teleport-slack-v9.0.0-1-g0d4c060b-dirty go1.17.8
```

**Event Handler Plugin**
```bash
$ docker run -it 146628656107.dkr.ecr.us-west-2.amazonaws.com/teleport/event-handler-plugin:9.0.0 version
Teleport event handler v9.0.0 git:teleport-event-handler-v9.0.0-7-g91dd050e-dirty go1.17.8
```

## Gitlab Tag
Before merging I'd want to confirm with someone why the gitlab plugin has been pinned to 0.2.0 and if it should stay that way. I will update this section when I have further information. 

## Testing
I confirmed that each built image referred correctly to the write Gitref as was originally specified and also made sure all tests passed.


wadells (commented):
I think the root Makefile may need to be upated too -- see the `update-version` target

Also, was gitlab just an unfortunate choice when I was looking at https://github.com/gravitational/teleport-plugins/issues/453 ? It seems more busted than the others. Even after your fix the version looks wrong compared to peers.

I'm fine with the -ldflags technique, but this is different (simpler imo) than what core uses in the main teleport repo.  We should check with @tcsc and @r0mant to see if they'd rather keep the consistent "generated version.go` technique.

wadells:
> It's my understanding that the `update-version` target is a _helper_ target for setting the versions across the various plugins and terraform provider. The actual versioning of these plugins is driven from their local Makefile. What kind of changes are you thinking of for the root Makefile?

I don't think `make version.go` is a target anymore:

https://github.com/gravitational/teleport-plugins/blob/74e6a76bd523704c8fecb6585d7a079dddadd872/Makefile#L107

logand22:
> I don't think `make version.go` is a target after your changes:

Wow, apologies, I didn't see that. Fixed in https://github.com/gravitational/teleport-plugins/pull/454/commits/9449515b03f4ca2fce8fa7356f0be0609b137be9

wadells (approved):

access/Dockerfile
```diff
@@ -7,6 +7,9 @@ ARG ACCESS_PLUGIN
 
 WORKDIR /workspace
 
+# Copy .git directory needed for go build
```
tcsc, line 10: Am I right in thinking that this image will be rebuilt on every push in Drone? Otherwise, I'm not sure I understand adding the `.git` dir

tcsc (approved):

access/Dockerfile
kbence: I believe it's needed for the `git describe ...` commands in the `Makefile`s

kbence (approved):
Fantastic, I love clean (== non-dirty) builds!

access/Dockerfile
logand22: @kbence is correct. The binaries are built with their Gitref variable set to the output of the `git describe` command. Also the goal is for these images to be built uniquely on every push to the master branch, however this PR just fixes some of the processes with ensuring the versioning is correctly reported (matching the tsh, tctl commands) across the different plugins.

Issue #438, opened 2022-03-03
Publish OCI images for teleport plugins


wadells:
## Summary
This is a sub-issue of making our teleport plugins more user friendly (https://github.com/gravitational/teleport-plugins/issues/160). This ticket tracks the following dev tasks:

- [x] develop OCI images for all teleport plugins
- [x] make sure these images are validated via CI
- [x] automatically publishing images to a staging location during tag build and master push builds
  - [x] we may need to ensure unique build ids for all commits -- to accommodate pushing untagged commits without overwrite
- [x] copy images from the staging location to the customer facing distribution during promotion
- [x] write dev facing documentation about the images/publishing in this repo
- [x] write customer facing documentation about using images in the teleport docs
- [x] backport all the above to the appropriate branches (probably v9, but worth double checking v8 and prior can be excluded)

logand22:
For the CI portion of this. I noticed both **drone** and **cloudbuild** being used in this repository. Should I lean towards one or the other?

wadells:
> For the CI portion of this. I noticed both **drone** and **cloudbuild** being used in this repository. Should I lean towards one or the other?

Use cloudbuild for CI.  I think we still prefer drone for publishing, as the point of this project isn't to refactor the publishing/promotion pipleline.
 
> > Make sure these images are validated via CI
> 
> Can you clarify what you mean by this?

Ideally, we'd build the docker images before merge, such that if there was a syntax error in the dockerfiles, we'd catch it.

My general thought is:

* PR -> build image, do not publish anywhere, throw it away after completion
* Merge to `master` / `branch/*` or push a tag -> build an image and push to an internal private repo
* Promote a build -> copy the bits from the internal private repo to a customer facing repo

My concern here would be: can we safely build an image at the CI stage.  DinD hasn't been sufficient isolation previously, and I'm not sure what options GCB offers for building images at the CI phase.  Build after approval (for external contributors) is fine.

tcsc:
> Use cloudbuild for CI. I think we still prefer Drone for publishing

Yeah. The _eventual_ goal is GCB for everything, but we're a long, long way from there.

> Can we safely build an image at the CI stage
 
GCB can definitely build/push images to a docker repo, so the question becomes can we do it _safely_. Reasonably sure that it's done by mounting the docker socket into the build container (i.e. Docker-out-of-Docker), with whatever implications that has (and I don't know enough about that to comment on off the top of my head).

logand22:
An update: 

[...]

PR #456, opened 2022-03-17
Push Access Plugins and Event Handler Docker Images to ECR

logand22:
This PR is a part of https://github.com/gravitational/teleport-plugins/issues/438

This adds `docker push` make targets.

Additionally it adds additional steps in the tag event pipelines in the drone file. 

## Questions
Will the MacOS build target have docker installed in it in order to run builds. Additionally, do I actually need to run these builds? Docker for Desktop, MacOS images on intel (which I believe the builder here is that) get tagged as linux + amd64, which would be the same as the kubernetes builds. 

## Testing
Tested that the new Makefile targets work properly by running them locally. 

The drone changes were not tested locally due to not being able to run `drone exec` with kubernetes pipelines. I am open to suggestions on testing drone pipeline changes before merging to master.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -201,6 +201,36 @@ steps:
       target: teleport-plugins/tag/${DRONE_TAG}
       strip_prefix: /go/src/github.com/gravitational/teleport-plugins/build
 
+  - name: Build and push to ECR
+    image: docker:git
+    environment:
+      AWS_ACCESS_KEY_ID:
+        from_secret: PLUGIN_DRONE_ECR_KEY
+      AWS_SECRET_ACCESS_KEY:
+        from_secret: PLUGIN_DRONE_ECR_SECRET
+      AWS_DEFAULT_REGION: us-west-2
+      DOCKER_BUILDKIT: 1
+    volumes:
+      - name: dockersock
+        path: /var/run
+    commands:
+      - apk add --no-cache make aws-cli
+      - export PLUGIN_TYPE=$(echo ${DRONE_TAG} | cut -d- -f2)
+      - aws ecr get-login-password | docker login -u="AWS" --password-stdin 146628656107.dkr.ecr.us-west-2.amazonaws.com
+      - make docker-push-access-${PLUGIN_TYPE}
+
+services:
+  - name: start docker
+    image: docker:dind
+    privileged: true
+    volumes:
+      - name: dockersock
+        path: /var/run
+
+volumes:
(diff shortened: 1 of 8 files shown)


wadells (changes requested):
Consider moving these  to their own pipeline, such that we have `tag-build-plugins-linux` and `tag-build-plugins-docker` for instance.  This should speed up execution via parallelization, and also allow re-running only a single pipeline if it fails.


> ## Questions
> 
> Will the MacOS build target have docker installed in it in order to run builds. Additionally, do I actually need to run these builds? Docker for Desktop, MacOS images on intel (which I believe the builder here is that) get tagged as linux + amd64, which would be the same as the kubernetes builds.

I am under the impression docker on mac ran everything inside a linux vm... meaning those images would be the same as the linux images.  See the pipeline idea above for how we can publish images once.

> ## Testing
> 
> Tested that the new Makefile targets work properly by running them locally.
> 
> The drone changes were not tested locally due to not being able to run `drone exec` with kubernetes pipelines. I am open to suggestions on testing drone pipeline changes before merging to master.

I'd recommend adding a trigger to the new pipeline that will trigger off a push to your branch, and running it there.  Do that in a test branch.  See https://github.com/gravitational/teleport/pull/10118 for an example.

Because we're publishing to the staging repo here, we don't particularly care if there are some extra images in it.  At least, we wouldn't if we had distinct build versioning a la what you set up for houston.  If you're going to be pushing e.g a `9.0.2` and we have tag immutability on... you may need to be more careful.

logand22:
> Consider moving these to their own pipeline, such that we have tag-build-plugins-linux and tag-build-plugins-docker for instance. This should speed up execution via parallelization, and also allow re-running only a single pipeline if it fails.

Changed in https://github.com/gravitational/teleport-plugins/pull/456/commits/6d86bed300f8175fddad2b3b8c941ee8cef7dcd7

## Testing
Drone changes tested here: https://drone.platform.teleport.sh/gravitational/teleport-plugins/960/2/3

The push to ECR for the access plugin failed only due to missing the `DRONE_TAG` env variable due to it not being a tag triggered event. This won't be a problem with the changes made in the above drone file.

logand22:
This will work for tagged builds. I will followup with a PR that continuously deploys images on push to master. The tracking issue is here: https://github.com/gravitational/teleport-plugins/issues/460

wadells (approved):

tcsc (approved):

PR #457, opened 2022-03-18
Terrform custom registry generator & promotion script

tcsc:
This script is intended to maintain a custom terraform registry that will be used to distribute the Teleport terraform provider.

The basic promotion steps are:
1. retrieve he pre-built terraform provider(s) from the staging bucket
2. repack them into a terraform-registry compatible format
3. retrieve the current repository `versions` index file (if present)
4. update (or create, if necessary) a local copy of the registry index, end
5. push all of the generated & updated files up the S3 bucket that backs the registry

The promotion code runs in response to two events:
1. A `terraform` tag build. The promotion is dairy-chained onto the terraform provider builds using Drone's `depends_on` pipeline config, and the resulting package is uploaded to a _staging_ repository, and 
3. A release promotion, where the resulting package is uploaded to the _production_ repository

The Drone YAML is configured to execute the same code in both cases, with different values supplied via the Drone environment.


tooling/internal/terraform/registry/protocol.go
```diff
@@ -0,0 +1,160 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"os"
+	"path"
+	"path/filepath"
+
+	"github.com/ProtonMail/go-crypto/openpgp"
+	"github.com/gravitational/trace"
+	"golang.org/x/crypto/openpgp/armor"
```
tcsc, line 12: Should be the ProtonMail version

tooling/internal/filename/parse_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseFilename(t *testing.T) {
+	t.Run("WithLeadingPath", func(t *testing.T) {
+		info, err := Parse("/some/bath/to/file/terraform-provider-teleport-v7.0.0-darwin-amd64-bin.tar.gz")
```
wadells, line 11:
```suggestion
		info, err := Parse("/some/path/to/file/terraform-provider-teleport-v7.0.0-darwin-amd64-bin.tar.gz")
```

tooling/internal/filename/parse_test.go
```diff
@@ -0,0 +1,39 @@
+package filename
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseFilename(t *testing.T) {
```
wadells, line 9: Maybe add a semver prerelease test case?  We do get the odd ask to publish prereleases:

- https://github.com/gravitational/teleport/issues/7691
- https://github.com/gravitational/teleport-plugins/pull/444#issuecomment-1062346211

To be clear, I don't expect that we'll publish pre-releases in this swing -- I am thinking this may be an edge that someone would trip over in the future.

wadells (commented):
Go code LGTM.

I may have missed this in the RFD, but I think it might be good to "shift left" the zip repacking and signing from promote to tag. For instance, this is what the current workflow looks like:

* tag -> publish to `dronestorage`
* promote -> copy tar from `dronestorage` to `clientbuilds`, repack from tar to zip and sign, generate new index, write to release bucket

Ignoring the fact that `dronestorage` and `clientbuilds` need to die, and be replaced by the release service, I'd propose:

* tag -> publish tar to `dronestorage`, repack from tar to zip and sign, publish to internal staging repo/bucket, generate new staging index
* promote -> copy tar from `dronestorage` to `clientbuilds`, copy signed zip from internal staging repo to release bucket, generate new release index

The benefit here is we can test changes to the zip -> tar conversion, signing, and index generation without actually publishing to customers.  This is pretty valuable, especially during the next month where we'll be iterating on this quite a bit.

An alternative is to have a mirror infra (e.g. buckets in `teleport-dev` where publishing can be tested safely), but my gut says that wiring up to Drone will be a pain.

tcsc:
>  it might be good to "shift left" the zip repacking

I _think_ it's possible in Drone to daisy-chain a new pipeline onto the existing `tag build` using prerequisites or something similar. Will have to do some more research. If so, that should be pretty easy to plug this into the back of the existing `tag`-triggered build.

 If not, my alternative was (similarly to yous, I think) to use Drone's notion of promotion environments to:

1. set up separate testing and production environments, and
2.  publish to "staging" during development. 
 
Although, this _may_ be a bit more dangerous, as we'd have to make sure to pick the right env every time we promote.

> my gut says that wiring up to Drone will be a pain.

I'll listen to your gut for now, at least in matters of Drone.

tooling/cmd/promote-terraform/main.go
```diff
@@ -0,0 +1,252 @@
...
+		return nil, trace.Wrap(err)
+	}
+
+	return s3.NewFromConfig(cfg), nil
+}
+
+func loadSigningKey(keyText string) (*openpgp.Entity, error) {
+	log.Info("Decoding signing key")
+
+	strings.NewReader(keyText)
```
logand22, line 239: Is this needed?

tcsc: Nope. Cut & Paste error. Gone.

.drone.yml
```diff
@@ -625,7 +625,102 @@ volumes:
...
+    environment:
+      STAGING_REGION: us-east-2
+      STAGING_BUCKET:
+        from_secret: AWS_S3_BUCKET
+      STAGING_ACCESS_KEY_ID:
+        from_secret: AWS_ACCESS_KEY_ID
+      STAGING_SECRET_ACCESS_KEY:
+        from_secret: AWS_SECRET_ACCESS_KEY
+      
+      PRODUCTION_REGION: ap-southeast-2
```
tcsc, line 712: Should probably parameterize the regions as well.

.drone.yml
```diff
@@ -625,7 +625,102 @@ volumes:
...
+depends_on:
+  - tag-build-terraform-linux
+  - tag-build-terraform-darwin
+
+concurrency:
+  limit: 1
+
+steps:
+  - name: Upload terraform provider to staging registry
+    image: golang:1.17.5
```
gzigzigzeo, line 648: I think, we could parameterise the required Go version as it seems equal for the whole suite (if Drone allows it).

Makefile
```diff
@@ -73,9 +73,14 @@ event-handler:
 
 # Run all tests
 .PHONY: test
-test:
+test: test-tooling
 	@echo Testing plugins against Teleport $(TELEPORT_GET_VERSION)
-	go test -race -count 1 ./...
+	go test -race -count 1 $(shell go list ./...)
```
gzigzigzeo, line 78: Not sure if my comment is applicable here, but looks like. Terraform provider tests need env var `TF_ENV=true` to be specified (Terraform framework requirement). Terraform must be available as well.

tooling/cmd/promote-terraform/main.go
```diff
@@ -0,0 +1,357 @@
...
+		if err != nil {
+			return trace.Wrap(err, "failed to extract key")
+		}
+
+		log.Tracef("... to %s", key)
+
+		doUpload := func() error {
+			f, err := os.Open(f)
+			if err != nil {
+				return err
```
gzigzigzeo, line 153: I think, `trace.Wrap()` is missing here

tooling/cmd/promote-terraform/main.go
```diff
@@ -0,0 +1,357 @@
...
+			}
+			defer f.Close()
+
+			_, err = uploader.Upload(ctx, &s3.PutObjectInput{
+				Bucket: aws.String(bucketName),
+				Key:    aws.String(key),
+				Body:   f,
+			})
+
+			return err
```
gzigzigzeo, line 163: `trace.Wrap()`?

gzigzigzeo (approved):

tooling/cmd/promote-terraform/main.go
```diff
@@ -0,0 +1,357 @@
...
+	versionIndex[newVersion.Version] = newVersion
+	versions.Versions = flattenVersionIndex(versionIndex)
+
+	if err = versions.Save(versionsFilePath); err != nil {
+		return trace.Wrap(err, "failed saving index file")
+	}
+	files = append(files, versionsFilePath)
+
+	// Finally, push the new index files to the production bucket
+	err = uploadRegistry(ctx, s3client, prodBucket.bucketName, workspace.productionDir, files)
```
r0mant, line 125: Is there anything we can do on the tool side to make sure that the index file hasn't changed between when we downloaded it above and saving it here? In case we get simultaneous builds running somehow. I don't suppose S3 has something like compare-and-swap?

tooling/internal/filename/parse.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/coreos/go-semver/semver"
+	"github.com/gravitational/trace"
+)
+
+var (
+	filenamePattern *regexp.Regexp = regexp.MustCompile(`^(?P<plugin>.*)-teleport-v(?P<version>.*)-(?P<os>linux|darwin|windows)-(?P<arch>amd64|arm|aarch)-bin.tar.gz$`)
+)
+
+type Info struct {
```
r0mant, line 16: NIt: Add godocs on exported methods and objects.

tooling/internal/terraform/registry/archives.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"archive/tar"
+	"archive/zip"
+	"compress/gzip"
+	"io"
+
+	"github.com/gravitational/trace"
+)
+
+func repack(dst io.Writer, src io.Reader) error {
```
r0mant, line 12: Can you add a godoc explaining what repack does? I.e. what are the expected src and dst formats. Maybe include a link to the terraform docs explaining what the format for the provider is expected to be.

tooling/internal/terraform/registry/protocol.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+func VersionsFilePath(prefix, namespace, name string) string {
+	return path.Join(prefix, namespace, name, versionsFilename)
+}
+
+// Versions is a Go representation of the Terraform Registry Protocol
+// `versions` file
+type Versions struct {
+	Versions []*Version `json:"versions"`
```
r0mant, line 27: Nit: Does this need to be a pointer?

tooling/internal/terraform/registry/protocol.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func NewVersions() Versions {
+	return Versions{Versions: []*Version{}}
+}
+
+func (idx *Versions) Save(filename string) error {
```
r0mant, line 34: Nit: Could you please add godocs on all exported methods and types here?

tooling/internal/terraform/registry/sha.go
```diff
@@ -0,0 +1,19 @@
...
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"io"
+	"path/filepath"
+
+	"github.com/gravitational/trace"
+)
+
+// Sha256Sum generates the equivalent contents of running `sha256sum` om some
```
r0mant, line 13:
```suggestion
// sha256Sum generates the equivalent contents of running `sha256sum` on some
```

tooling/internal/terraform/registry/sign.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// RepackProvider takes a provider tarball and repacks it as a zipfile compatible
+// with a terraform provider registry, generating all the required sidecar files
+// as well. Returns a `RepackResult` instance containing the location of the
+// generated files and information about the packed plugin
+func RepackProvider(dstDir string, srcFileName string, signingEntity *openpgp.Entity) (*RepackResult, error) {
+	info, err := filename.Parse(srcFileName)
+	if err != nil {
+		return nil, trace.Wrap(err, "Bad filename")
```
r0mant, line 60: Include filename in the error message?

tooling/internal/terraform/registry/sign.go
```diff
@@ -0,0 +1,144 @@
...
+
+	result := &RepackResult{
+		Info:          info,
+		FileNames:     makeFileNames(dstDir, info),
+		SigningEntity: signingEntity,
+	}
+
+	// compute sha256 and format the sha file as per sha256sum
+	var sums bytes.Buffer
+	result.Sha256, err = sha256Sum(&sums, result.Zip, zipArchive.Bytes())
```
r0mant, line 88: Is there a way to do this without reading the entire zip archive in memory? Using a reader for example.

tooling/internal/terraform/registry/sign.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func writeOutput(entry *RepackResult, zip, sums, sig []byte) error {
+	zipFile, err := os.Create(entry.Zip)
+	if err != nil {
+		return trace.Wrap(err, "opening zipfile failed")
+	}
+	defer zipFile.Close()
+
+	_, err = zipFile.Write(zip)
```
r0mant, line 116: Instead of os.Create+file.Write, would ioutil.WriteFile work?

r0mant (commented):
lgtm mostly, just a few suggestions

tooling/internal/terraform/registry/sign.go
tcsc: One option would be to compress to a temp file, and then move it into place at the end rather than write it out.

tooling/internal/terraform/registry/protocol.go
tcsc: Nope

tooling/internal/terraform/registry/sign.go
tcsc: Done

tooling/internal/terraform/registry/sign.go
tcsc: Yep. And much neater, too.

tooling/internal/terraform/registry/sign.go
tcsc: Done

Makefile
tcsc: This is change is to avoid `go test` picking up the `tooling` module here. For some reason, in my testing, `./...` included the `tooling` package and failed, where `$(shell go list ./...)` correctly treated it as a separate module.

Just tested it again now an all seems fine.

wadells (approved):

.drone.yml
```diff
@@ -625,7 +625,104 @@ volumes:
...
+---
+kind: pipeline
+type: kubernetes
+name: promote-terraform-provider
+
+trigger:
+  event:
+  - promote
+  target:
+  - production
```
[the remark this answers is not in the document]
tcsc: Excellent idea. Added an explicit staging build as well for this same reason

r0mant (approved):

Issue #439, opened 2022-03-03
Write helm charts for teleport plugins


wadells:
## Summary
This is a sub-issue of makging our teleport plugins more user friendly (https://github.com/gravitational/teleport-plugins/issues/160). This ticket tracks the following dev tasks:

 - [x] Using the images from https://github.com/gravitational/teleport-plugins/issues/438, write a helm chart for each teleport plugin that needs to be hosted
 - [x] upon promotion publish the charts to https://charts.releases.teleport.dev/
 - [ ] write developer facing documentation on how to maintain/publish these charts in this repo
 - [ ] write end user facing documentation on how to use these charts in the teleport docs
 - [ ] backport all the above to the appropriate branches (probably v9, but worth double checking v8 and prior can be excluded)

PR #459, opened 2022-03-18
Helm chart for acces-plugin-email

kbence:
This change adds a Helm chart for `access-plugin-email` and it'll serve as a base for the rest of the plugins. 

Part of #439

TODO:
- [x] Clean unused/unnecessary values from `values.yaml` added by the default template
- [ ] ~Make it possible to use secrets and configmaps where it's applicable (but also keep string input where it makes sense)~
- [x] Write tests using `helm unittest`
- [x] Create `values.schema.json` to help validating values.

# Testing done

I've tested the chart in a local minikube environment, I'll also test it under EKS once the images are available on ECR.

I've also added some unit tests using the third party plugin [vbehar/helm3-unittest](https://github.com/vbehar/helm3-unittest) (we are already using it in [gravitational/teleport](https://github.com/gravitational/teleport/blob/master/Makefile#L481)).

# Deployment

Deployment of the helm chart is going to be in a separate pull request.


charts/access/email/templates/configmap.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "email.fullname" . }}
+  labels:
+    {{- include "email.labels" . | nindent 4 }}
+data:
+  teleport-email.toml: |
+    [teleport]
```
logand22, line 9: Seems to be some missing config options in this section. See https://github.com/gravitational/teleport-plugins/blob/9e149895f54edd4870f760be08a4c97dae49c151/lib/config.go#L18

charts/access/email/templates/configmap.yaml
```diff
@@ -0,0 +1,35 @@
...
+    private_key = "{{ .Values.mailgun.privateKey }}"
+    {{ else if .Values.smtp.enabled }}
+    [smtp]
+    host          = "{{ .Values.smtp.host }}"
+    port          = {{ .Values.smtp.port }}
+    username      = "{{ .Values.smtp.username }}"
+    {{ if eq .Values.smtp.passwordFile "" -}}
+    password      = "{{ .Values.smtp.password }}"
+    {{ else -}}
+    password_file = "{{ .Values.smtp.passwordFile }}"
```
logand22, line 25: Is this possible to be set within the deployment.yaml?

charts/access/email/values.yaml
```diff
@@ -0,0 +1,72 @@
...
+nodeSelector: {}
+
+tolerations: []
+
+affinity: {}
+
+#
+# Plugin specific options
+#
+teleport:
```
logand22, line 47: Possible to document the different options and potentially there precedence? For instance passwordFile takes precedence over password for `smtp` if they are both set.

logand22 (commented):
Looking good so far.

charts/access/email/templates/configmap.yaml
kbence: `AuthServer` deprecates `Addr` and the fields `ClientKey`, `ClientCrt` and `RootCAs` are just another way to pass the same data that's supposed to be in the file pointed by `Identity`. @walt and I agreed that we should only support "best practices" and avoid multiple ways to configure the same thing. That's why I've left out those fields after investigating what they are actually doing.

charts/access/email/templates/configmap.yaml
kbence: Oh, you are right, there's no way we can pass volumes to be mounted. Thanks for spotting this!

.cloudbuild/ci/helm-unittest.yaml
```diff
@@ -0,0 +1,9 @@
+steps:
+  - name: 'alpine/helm'
+    entrypoint: /bin/sh
+    args:
+      - -c
+      - |
+        apk add openssl curl bash git &&
+        helm plugin install https://github.com/vbehar/helm3-unittest &&
```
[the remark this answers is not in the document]
kbence: I based my change on https://github.com/gravitational/teleport/pull/11062/files, I'm not sure what was the reasoning behind choosing one over the other. Pinning it makes sense, thanks for catching it!

Joerger (approved):
LGTM

PR #464, opened 2022-03-21
Update versioning for build systems

logand22:
This PR updates teleport-plugin versioning in order to _shift left_ the build and push of development images for the teleport plugins oci images. 

**NOTE**: This PR is currently in draft mode while I await the merging of https://github.com/gravitational/teleport-plugins/pull/456 and https://github.com/gravitational/teleport-plugins/pull/454

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -106,6 +106,66 @@ steps:
     image: golang:1.17.5
     commands:
       - make build-all
+---
+kind: pipeline
+type: kubernetes
+name: build-and-push-images-on-push-linux
+
+trigger:
+  branch:
+    - master
+  event:
+    include:
+      - push
+
+depends_on:
+  - test-linux
+
+steps:
+  - name: fetch tags
+    image: docker:git
+    commands:
+      - git fetch --tags
+
+  - name: build images
+    image: docker:git
+    environment:
+      DOCKER_BUILDKIT: 1
+    volumes:
+      - name: dockersock
+        path: /var/run
+    commands:
+      - make docker-build-access-plugins
+      - make docker-build-event-handler
+
+  - name: push to ECR
+    image: docker:git
+    environment:
+      AWS_ACCESS_KEY_ID:
+        from_secret: PLUGIN_DRONE_ECR_KEY
+      AWS_SECRET_ACCESS_KEY:
+        from_secret: PLUGIN_DRONE_ECR_SECRET
+      AWS_DEFAULT_REGION: us-west-2
+    volumes:
+      - name: dockersock
+        path: /var/run
+    commands:
+      - apk add --no-cache make aws-cli
+      - aws ecr get-login-password | docker login -u="AWS" --password-stdin 146628656107.dkr.ecr.us-west-2.amazonaws.com
+      - make docker-push-event-handler
+      - make docker-push-access-plugins
+
+services:
+  - name: start docker
+    image: docker:dind
+    privileged: true
+    volumes:
+      - name: dockersock
+        path: /var/run
+
+volumes:
+  - name: dockersock
+    temp: {}
 
 ---
 kind: pipeline
@@ -509,6 +569,6 @@ steps:
 
 ---
 kind: signature
-hmac: 7381b82e23954f22b26d90324d9219d7b1f2c0d81a4b9e2b27f868674beec197
+hmac: ede0e0312d72f7cb7f00e4278b27c1d4668d9399ded92ae6957fffde092ed8c5
 
 ...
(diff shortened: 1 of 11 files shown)


.drone.yml
```diff
@@ -106,6 +106,66 @@ steps:
...
     commands:
       - make build-all
+---
+kind: pipeline
+type: kubernetes
+name: build-and-push-images-on-push-linux
+
+trigger:
+  branch:
+    - master
```
tcsc, line 116: Will we need to build releases for `branch/v*` as well?

version.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/sh
```
tcsc, line 1: We're trying to shift away from using bash for build scripts & tooling, and using small go programs wherever possible.

If you can, please update it to use go. @zmb3 found [this library](https://github.com/bitfield/script) the other day, which may help, and [go-git](https://github.com/go-git/go-git) is useful for manipulating repositories.

Also: Beware that `go-git` is not a fully-featured `git` implementation and may not have what you need. If you go down this path and get hit by one of these omissions, my advice is not try & force `go-git`to handle it - just shell out to the command line `git`. I've lost _days_ fighting with go-git, and while it's very good when it works, when it doesn't work it just. doesn't. work.

(My 2c)

.drone.yml
logand22: I noticed that `branch/v*` is referenced in the drone branch triggers but there isn't any `branch/v*` branches. This PR https://github.com/gravitational/teleport-plugins/pull/116 added them but makes no mention as to their purpose within the repository. 

However, if desired I can add them in.

wadells (commented):
I agree with @tcsc about go.  The version.sh script, while tested, is prone to failures like https://github.com/gravitational/teleport/issues/9494

More holistically, I'm not sure I want to push using git tags as the version source of truth into teleport (recognizing we've done this on the gravity side for a while).

Why? Determinism.

I assert, as a long term target: When you check out a particular commit and build teleport for it, we want the output artifacts to be the same every time. This is something @fspmarshall is working towards as an ancillary part of https://github.com/gravitational/teleport/pull/10690.

To achieve this, we want the contents of the build to only depend on information present in the commit.  Because git implements tags as a layer above commits (tags have commits, not vice versa) then the version info (part of the binary) shouldn't depend on the git tag.

Instead, what I think we should do is:

1. Use a version in a well-known file. (e.g. `9.0.0`, or `10.0.0-dev`).  This will be part of the tree, and thus encoded in the the git commit hash.
2. Have the version script take that, and calculate how many changes since the version file was changed. Use this as the version pre-release modifier.  Because commits encode parent info, the delta since the last versionfile update is also part of the commit hash.
3. Change the version string to use this deterministic info.  Some details on how:

```
Key
S: static: these chars never change for a product
d: deterministic: these chars are derived directly from the tree/commit and
   cannot change without also changing the commit hash
g: generated deterministic: these fields are calculated using information present
   in the commit, but we may need to do some math or string manipulation.
n: nondeterminsic: these fields can change without any change in the commit hash
```

Here is our current version output:
```
product v8.0.6 git:v8.0.6-1-g807c6d662 go1.17.3
        ^          ^      ^ ^          ^
SSSSSSSdddddddSSSSSnnnnnnSnSSdddddddddSdddddddd
```
Notably, the git field uses information outside the commit, and is not semver compliant
(v8.0.6-1 sorts before v8.0.6 according to the spec, though git uses -1 to communicate
1 commit after v8.0.6).

I propose instead:
```
             n+1 for correct sorting
             |
product v8.0.7-dev.1+f37e383d git:f37e383d23cdd6d712ea718ea8804c130b2a500d go1.17.3
        ^    ^                    ^                                        ^
SSSSSSSSdddddggggggggggggggggSSSSSddddddddddddddddddddddddddddddddddddddddSdddddddd
```
Patch, prerelease and build metadata are all calcuated based on info included in the commit hash
We  choose to add build metadata to the version string that artifacts containing the build id will have distinct URIs.  There are many commits that are one after 8.0.6, but only one f37e383d and we don't want `v8.0.7-dev.1+f37e383d` artifacts to overwrite bits from `v8.0.7-dev.1+adb60841`.

Should all of this be implemented herein: maybe.  I was hoping to put together a RFD for this idea and solicit feedback, but I haven't found the time to do so yet.

I'd suggest either:

1) We wait on the shift left and focus on other publishing tasks. This is probably the right answer from the customer perspective.
2) We get out a proper versioning RFD and get buy in from core for the change.  If you really want to get into versioning, and get a 2nd rfd out this quarter, you can pursue this.

PR #466, opened 2022-03-24
Add promotion step to drone pipeline for pushing to quay

logand22:
As a part of https://github.com/gravitational/teleport-plugins/issues/438, this PR adds additional pipelines to `.drone.yml` file in order to promote internally staged docker images to their public registry with quay.io. 

The current promotion workflow occurs on a per tag basis. This means that for every tag created by the `make update-tag` target, a teleport employee triggers a promotion step for this tag. 

This workflow is reflected in the additional promotion steps added in this PR. Therefore, no change to the procedure in which a tag is promoted will occur in order for the images to be promoted to Quay.io.

## Testing
The promotion step was confirmed through local manual testing.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -508,7 +508,116 @@ steps:
       strip_prefix: /go/src/github.com/gravitational/teleport-plugins/
 
 ---
+kind: pipeline
+type: kubernetes
+name: publish-access-images
+
+trigger:
+  event:
+    - promote
+  target:
+    - production
+    - publish-images
+    - publish-access-images
+  ref:
+    include:
+      - refs/tags/teleport-jira-v*
+      - refs/tags/teleport-mattermost-v*
+      - refs/tags/teleport-pagerduty-v*
+      - refs/tags/teleport-slack-v*
+      - refs/tags/teleport-email-v*
+
+clone:
+  disable: true
+
+steps:
+  - name: Promote image
+    image: docker:git
+    environment:
+      AWS_ACCESS_KEY_ID:
+        from_secret: PLUGIN_DRONE_ECR_KEY
+      AWS_SECRET_ACCESS_KEY:
+        from_secret: PLUGIN_DRONE_ECR_SECRET
+      AWS_DEFAULT_REGION: us-west-2
+      DOCKER_BUILDKIT: 1
+    volumes:
+      - name: dockersock
+        path: /var/run
+    commands:
+      - apk add --no-cache make aws-cli
+      - export PLUGIN_TYPE=$(echo ${DRONE_TAG} | cut -d- -f2)
+      - aws ecr get-login-password | docker login -u="AWS" --password-stdin 146628656107.dkr.ecr.us-west-2.amazonaws.com
+      - docker login -u="$QUAYIO_DOCKER_USERNAME" -p="$QUAYIO_DOCKER_PASSWORD" quay.io
+      - make docker-promote-access-${PLUGIN_TYPE}
+
+services:
+  - name: start docker
+    image: docker:dind
+    privileged: true
+    volumes:
+      - name: dockersock
+        path: /var/run
+
+volumes:
+  - name: dockersock
+    temp: {}
+
+---
+kind: pipeline
+type: kubernetes
+name: publish-event-handler-image
+
+trigger:
+  event:
+    - promote
+  target:
+    - production
+    - publish-images
+    - publish-event-handler-image
+  ref:
+    include:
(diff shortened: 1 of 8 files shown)


wadells:
> ## Testing
> 
> The promotion step was confirmed through local manual testing.

How do you plan to test this after merge?  I'd like to make sure we've done end-to-end testing before the next 9.0.x release, when these would go live.

wadells (approved):
Looks reasonable.

wadells:
> @wadells
> 
> My current plan is to walk through an entire release process with a dummy tag and dummy promote. I will only be promoting the publish-images target and plan on cleaning up the tags afterwards.
> 
> For example:
> 
> ```
> tag 0.0.193 -> triggers drone push to ecr -> test the images -> run promotion on those tags -> check and test quay images. 
> ```

This sounds great!  The one addendum I'd add is:

Use a tag like `9.0.2-dev.1` and don't delete the tag afterwards. The artifacts can probably be cleaned up from the prod repo -- staging doesn't need to be cleaned up.

Why? Russell asks folks not to delete any tags (including testing ones) from core product repos.  A long time ago, an engineer re-pushed a git tag our customers had all ready pulled, and it spooked a customer.  After that incident then we've treated tags as immutable for teleport -- even testing tags. I wish we could enforce that immutability it via github, but it isn't possible currently.

We can play a bit more fast and loose with tags in non-customer-facing or non-product repos.

logand22:
> Use a tag like 9.0.2-dev.1 and don't delete the tag afterwards. The artifacts can probably be cleaned up from the prod repo -- staging doesn't need to be cleaned up.

That sounds good. Just to clarify though, leave the git tag alone but clean up the production images in quay?

wadells:
> > Use a tag like 9.0.2-dev.1 and don't delete the tag afterwards. The artifacts can probably be cleaned up from the prod repo -- staging doesn't need to be cleaned up.
> 
> That sounds good. Just to clarify though, leave the git tag alone but clean up the production images in quay?

Yeah -- thats probably what I'd do. I don't think leaving test images around is harmful, but nor is it particularly helpful to our customers.

.drone.yml
```diff
@@ -507,8 +507,121 @@ steps:
...
+      AWS_SECRET_ACCESS_KEY:
+        from_secret: PLUGIN_DRONE_ECR_SECRET
+      AWS_DEFAULT_REGION: us-west-2
+      DOCKER_BUILDKIT: 1
+    volumes:
+      - name: dockersock
+        path: /var/run
+    commands:
+      - apk add --no-cache make aws-cli
+      - aws ecr get-login-password | docker login -u="AWS" --password-stdin 146628656107.dkr.ecr.us-west-2.amazonaws.com
```
r0mant, line 607: Should `146628656107.dkr.ecr.us-west-2.amazonaws.com` come from some environment variable? This way it also won't need to be repeated twice.

access/email/Makefile
```diff
@@ -58,3 +59,9 @@ docker-build: ## Build docker image with the plugin.
 .PHONY: docker-push
 docker-push: 
 	docker push ${DOCKER_IMAGE}
+
+.PHONY: docker-promote
+docker-promote:
+	docker pull ${DOCKER_IMAGE}
+	docker tag ${DOCKER_IMAGE} ${DOCKER_IMAGE_QUAY}
+	docker push ${DOCKER_IMAGE_QUAY}
```
r0mant, line 67: Same for other promote targets below.
```suggestion
	docker pull ${DOCKER_IMAGE} && \
	docker tag ${DOCKER_IMAGE} ${DOCKER_IMAGE_QUAY} && \
	docker push ${DOCKER_IMAGE_QUAY}
```

r0mant (approved):
With a couple of comments.

.drone.yml
logand22: `146628656107.dkr.ecr.us-west-2.amazonaws.com` is referenced in two separate pipelines. I'm not aware of a way to share environment variables between multiple pipelines. 

However, I could move the login step into a Makefile target, say `make ecr-login`, in order to make it not be repeated in multiple places.

access/email/Makefile
logand22: Resolved with https://github.com/gravitational/teleport-plugins/pull/466/commits/2bec230f19840dfcfa1d72fd5740ac736bfb38ab

.drone.yml
logand22: I'm going to leave it as is for now, however, if we find ourselves needing to update this often then I will work on a solution to reduce the repetition.

PR #467, opened 2022-03-25
TF SAML Connector: require any of (EntityDescriptor, EntityDescriptorURL)

### LINKED ISSUE #448: entity_descriptor should not be required if entity_descriptor_url is set
In teleport in the upstream, an error is only thrown if both are not present. See: https://github.com/gravitational/teleport/blob/v8.3.4/api/types/saml.go#L358-L361

In the terraform provider, if you set only `entity_descriptor_url`, but not `entity_descriptor`, you get the following error:

```
│ Error: Missing required argument
│
│   on auth.tf line 9, in resource "teleport_saml_connector" "okta":
│    9:   spec {
│
│ The argument "entity_descriptor" is required, but no definition was found.
```

marcoandredinis:
For Terraform SAML Connector, at least one of the following Spec attributes must exist:
- EntityDescriptor
- EntityDescriptorURL

However, we were always requiring the EntityDescriptor
We removed this field as required

Things would work out because the server would validate that either one of them is present.
The server returns an error if both are empty.

However, to improve the UX we added a `AnyOf` validation to the terraform side
It will validate before hitting the server

Fixes #448

### DIFF
--- a/terraform/tfschema/validators.go
+++ b/terraform/tfschema/validators.go
@@ -165,30 +165,30 @@ OUTER:
 
 }
 
-// OneOfValidator validates that at least one of the attributes is set (known and not null)
-type OneOfValidator struct {
+// AnyOfValidator validates that at least one of the attributes is set (known and not null)
+type AnyOfValidator struct {
 	Keys []string
 }
 
-// UseOneOfValidator creates OneOfValidator
-func UseOneOfValidator(keys ...string) tfsdk.AttributeValidator {
-	return OneOfValidator{
+// UseAnyOfValidator creates AnyOfValidator
+func UseAnyOfValidator(keys ...string) tfsdk.AttributeValidator {
+	return AnyOfValidator{
 		Keys: keys,
 	}
 }
 
 // Description returns validator description
-func (v OneOfValidator) Description(_ context.Context) string {
-	return fmt.Sprintf("OneOf '%s' attributes must be present", strings.Join(v.Keys, ", "))
+func (v AnyOfValidator) Description(_ context.Context) string {
+	return fmt.Sprintf("AnyOf '%s' attributes must be present", strings.Join(v.Keys, ", "))
 }
 
 // MarkdownDescription returns validator markdown description
-func (v OneOfValidator) MarkdownDescription(_ context.Context) string {
-	return fmt.Sprintf("OneOf `%s` attributes must be present", strings.Join(v.Keys, ", "))
+func (v AnyOfValidator) MarkdownDescription(_ context.Context) string {
+	return fmt.Sprintf("AnyOf `%s` attributes must be present", strings.Join(v.Keys, ", "))
 }
 
 // Validate performs the validation.
-func (v OneOfValidator) Validate(ctx context.Context, req tfsdk.ValidateAttributeRequest, resp *tfsdk.ValidateAttributeResponse) {
+func (v AnyOfValidator) Validate(ctx context.Context, req tfsdk.ValidateAttributeRequest, resp *tfsdk.ValidateAttributeResponse) {
 	if req.AttributeConfig == nil {
 		return
 	}
@@ -196,7 +196,7 @@ func (v OneOfValidator) Validate(ctx context.Context, req tfsdk.ValidateAttribut
 	value, ok := req.AttributeConfig.(types.Object)
 	if !ok {
 		resp.Diagnostics.AddError(
-			"OneOf Object validation error",
+			"AnyOf Object validation error",
 			fmt.Sprintf("Attribute %v can not be converted to Object", req.AttributePath.String()),
 		)
 		return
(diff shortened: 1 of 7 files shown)


r0mant (commented):
@marcoandredinis This should only get merged after you fix the issue not requiring entity descriptor if entity descriptor URL is set, correct? Otherwise, TestSAMLConnectorOnlyEntityDescriptorURL test will be failing?

marcoandredinis:
> @marcoandredinis This should only get merged after you fix the issue not requiring entity descriptor if entity descriptor URL is set, correct? Otherwise, TestSAMLConnectorOnlyEntityDescriptorURL test will be failing?

@r0mant 
It already works like that: either one of them is necessary, otherwise we'll get an error
If the description is empty, it will try to fetch it using the URL:
https://github.com/gravitational/teleport/blob/062c92b81e524a89dd2b34e3cd2b8fb93ff34a83/lib/services/saml.go#L48

Maybe I'm not following, sorry

r0mant:
@marcoandredinis Isn't the issue described in https://github.com/gravitational/teleport-plugins/issues/448 that terraform provider doesn't allow setting `entity_descriptor_url`? Have you tried creating a SAML connector with terraform provider to repro the scenario described in the ticket?

Joerger:
The issue is [this line](https://github.com/gravitational/teleport-plugins/blob/9e149895f54edd4870f760be08a4c97dae49c151/terraform/protoc-gen-terraform-teleport.yaml#L254) which makes `entity_descriptor` a required field.

terraform/tfschema/validators_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/stretchr/testify/require"
+)
+
+func getObjectWithTwoKeys() types.Object {
+	return types.Object{
+		Attrs: map[string]attr.Value{
```
gzigzigzeo, line 32: It would be safer if AttrTypes would be defined here as well:
```
AttrTypes: map[string]attr.Type{
    "key1": types.String,
    "key2": types.String,
},
```

gzigzigzeo (approved):
LGTM

Joerger (approved):

terraform/tfschema/validators.go
```diff
@@ -163,3 +164,67 @@ OUTER:
...
+			fmt.Sprintf("Attribute %v can not be converted to Object", req.AttributePath.String()),
+		)
+		return
+	}
+
+	if value.Null || value.Unknown {
+		return
+	}
+
+	for _, key := range v.Keys {
```
r0mant, line 209: I may be missing something but I don't see where in this function we set the error if both fields are set? Shouldn't they be mutually exclusive?

terraform/tfschema/validators_test.go
```diff
@@ -0,0 +1,90 @@
...
+		"one-of-two": {
+			input:       getObjectWithTwoKeys(),
+			oneOfValues: []string{"key1"},
+			expectError: false,
+		},
+		"two-of-two": {
+			input:       getObjectWithTwoKeys(),
+			oneOfValues: []string{"key1", "key2"},
+			expectError: false,
+		},
```
r0mant, line 67: Shouldn't this test case expect error? In case of SAML connector for example, you should set either entity_descriptor or entity_descriptor_url, not both, correct?

r0mant (changes requested):
Unless I'm missing something, I think "either-or" validation isn't implemented.

Joerger:
> Unless I'm missing something, I think "either-or" validation isn't implemented.

They don't need to be mutually exclusive, internally we accept both, but `entityDescriptorURL` takes precedence (It's used to fill out `entityDescriptor`).

The naming `one-of` is definitely misleading though. This is more verbose but `requireAtleastOne` would be descriptive.

r0mant:
Yeah, `AtLeastOneOf` would probably be a bit more suitable name cause `OneOf` sort of implies exclusivity (which may be another useful validator we might need to add in future).

marcoandredinis:
@r0mant @Joerger Thank you for your reviews
You are correct, `OneOf` indicates exclusivity

I was thinking about using `AnyOf` to use a similar terminology that exists in OpenAPI/Swagger
https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/#anyof
> Use the anyOf keyword to validate the data against any amount of the given subschemas. That is, the data may be valid against **one or more** subschemas at the same time.

I'm fine with `AtLeastOneOf` if you prefer it :+1:

r0mant:
@marcoandredinis AnyOf sounds good to me also.

r0mant (approved):

Joerger (approved):

PR #473, opened 2022-03-30
Update teleport-plugin docs with OCI information

logand22:
This PR contains general documentation improvements as well as information on how to find and run teleport plugins docker images. This is a part of https://github.com/gravitational/teleport-plugins/issues/438

There are now three documented methods of installing the teleport-plugins:
1. Install from release
2. Docker images
3. Build from source

All three of these methods are documented in each plugin repository. 

Documentation has been added also on running the plugin both with local binary as well as with the docker images. 

## Testing
I tested that the docker image links provided throughout the documentation worked such that someone could copy and paste the image name from the documentation. 

However, I did not test that the plugin still works.

### DIFF
--- a/access/email/README.md
+++ b/access/email/README.md
@@ -2,11 +2,17 @@
 
 The plugin allows teams to receive email notifications about new access requests.
 
-## Setup
+## Install the plugin
 
-### Install the plugin
+There are several methods to installing and using the Teleport Email Plugin:
 
-Get the plugin distribution.
+1. Use a [precompiled binary](#precompiled-binary)
+
+2. Use a [docker image](#docker-image)
+
+3. Install from [source](#building-from-source)
+
+### Precompiled Binary
 
 ```bash
 $ curl -L https://get.gravitational.com/teleport-access-email-v7.1.0-linux-amd64-bin.tar.gz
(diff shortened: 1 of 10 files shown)


access/email/README.md
```diff
@@ -15,11 +21,34 @@ $ cd teleport-access-email
...
+```bash
+$ docker run quay.io/gravitational/access-plugin-email:9.0.2 version
+teleport-email v9.0.2 git:teleport-email-v9.0.2-0-g9e149895 go1.17.8
+```
+
+For a list of available tags, visit [https://quay.io/](https://quay.io/repository/gravitational/access-plugin-email?tab=tags)
+
+### Building from source
+
+To build the plugin from source you need Go >= 1.16 and `make`.
```
wadells, line 38: I'd avoid duplicating the go version in the readme. This feels like it will fall out of date.  Instead, maybe link to where the go version is set?

access/slack/README.md
```diff
@@ -1,19 +1,163 @@
...
+$ teleport-slack configure > /etc/teleport-slack.yml
+```
+
+Then, edit the config as needed.
+
+```TOML
+# Example slack plugin configuration TOML file
+
+[teleport]
+# This section
```
wadells, line 111: "This section" seems like a fragment. Admittedly, this may be the default from configure?  If it is verbatim, don't worry about it.  If it isn't verbatim, finish the thought?

wadells (approved):

access/pagerduty/README.md
```diff
@@ -14,90 +14,98 @@ This guide assumes you have
...
-    review_requests:
-      roles: ...
-      where: ... 
-version: v3
-EOF
-
-# ...
-$ tctl create -f rscs.yaml
+        verbs: ['list', 'read', 'update']
+version: v4
```
Joerger, line 76: Role version should be v5 here and in the other readmes as of `v9`
```suggestion
version: v5
```

Joerger (approved):

access/pagerduty/README.md
logand22: Fixed in https://github.com/gravitational/teleport-plugins/pull/473/commits/368cc434b1d661e0a63d1f1c66429a1cd0e32b1a

access/email/README.md
logand22: Fixed in https://github.com/gravitational/teleport-plugins/pull/473/commits/1124b8c56e4e0056573473df3a084931c75744d1

access/slack/README.md
logand22: It was verbatim, however I updated the messaging in the example configuration so that future versions don't include that line when printing the configuration. 

See https://github.com/gravitational/teleport-plugins/pull/473/commits/43d7e12cc82a73db1028d6525c4f8b89137075d2

Issue #291, opened 2021-08-18
Add support for loading secrets from environment


klizhentas:
Terraform enterprise does not support loading secrets from files, only from environment. Add support for loading secrets from environment and test with Terraform enterprise.

bpdohall:
Concrete example of a customer using Terraform Enterprise - https://goteleport.slack.com/archives/CEZH6UL64/p1637188987324600 - they are blocked from using the Teleport provider because a file on the Terraform agent's filesystem must be referenced by the provider config parameter `identity_file_path`.

If we were to offer an alternative which takes base64 content, the user would be able to leverage the [Terraform Enterprise API `vars` endpoint](https://www.terraform.io/docs/cloud/api/workspace-variables.html) to inject the identity file.

PR #476, opened 2022-03-31
improvement: allow setting terraform auth secrets through base64 encoded env vars

kazimierzbudzyk:
Fixes: https://github.com/gravitational/teleport-plugins/issues/291


r0mant:
@gzigzigzeo @marcoandredinis Could you guys take a look?

kazimierzbudzyk:
Sorry for the back and forth with commits - this is now ready for review.

terraform/provider/provider.go
```diff
@@ -279,6 +316,43 @@ func (p *Provider) validateAddr(addr string, resp *tfsdk.ConfigureProviderRespon
 	return true
 }
 
+// getCredentialsFromBase64 returns client.Credentials built from base64 encoded keys
+func (p *Provider) getCredentialsFromBase64(certBase64, keyBase64, caBase64 string, resp *tfsdk.ConfigureProviderResponse) (client.Credentials, bool) {
+	cert, err := base64.StdEncoding.DecodeString(certBase64)
+	if err != nil {
+		resp.Diagnostics.AddError(
```
marcoandredinis, line 323: Can we please add the env var name and config name?
I'm sure it will help with debugging

marcoandredinis, line 323: Can we please add the env var name and config name?
I'm sure it will help with debugging

terraform/provider/provider.go
```diff
@@ -171,6 +199,15 @@ func (p *Provider) Configure(ctx context.Context, req tfsdk.ConfigureProviderReq
 		creds = append(creds, cred)
 	}
 
+	if certBase64 != "" && keyBase64 != "" {
```
marcoandredinis, line 202: What if we want to get the cert+key from Base64 env vars but the CA from a file?

I'm not sure this is a requirement but seems to be reasonable

kazimierzbudzyk: I am mostly following the logic applied for the path based approach, where we only check cert and key. Here my assumption is that if we don't have anything passed  for `rootCaBase64` it will end up being decoded to empty `[]byte{}` and optional - like loglines are suggesting `Using auth with certificate, private key and (optionally) CA`.

kazimierzbudzyk:
> Can we please add some tests?
> 
> There was a bug in the `stringFromConfigOrEnv` 🙈

Do you want it in this PR? As there's no tests for the existing logic I had no good patterns for what you would like to test. If we want to add them here would really appreciate some guidance.

kazimierzbudzyk:
Additionally, I didn't want to jam it together with this PR but env var and default lookup could potentially be done in more idiomatic way using [DefaultFunc](https://www.terraform.io/plugin/sdkv2/schemas/schema-behaviors#defaultfunc) and `schema.EnvDefaultFunc`. It would look something like:
```
"cert_base64": {
    Type:              types.StringType,
    Optional:        true,
    Description:   "Base64 encoded TLS auth certificate.",
    DefaultFunc: schema.EnvDefaultFunc("TF_TELEPORT_CERT_BASE64", ""),
},
```
But again I would rather have it PRed separately as it's more of a refactor than change I'm trying to make here.

terraform/provider/provider.go
kazimierzbudzyk: Done.

terraform/provider/provider.go
marcoandredinis: Sounds good :+1:

marcoandredinis:
> > Can we please add some tests?
> > There was a bug in the `stringFromConfigOrEnv` see_no_evil
> 
> Do you want it in this PR? As there's no tests for the existing logic I had no good patterns for what you would like to test. If we want to add them here would really appreciate some guidance.

I think we can move forward with the current implementation :+1:

marcoandredinis (approved):
LGTM
Thank you @kazimierzbudzyk

marcoandredinis:
> Additionally, I didn't want to jam it together with this PR but env var and default lookup could potentially be done in more idiomatic way using [DefaultFunc](https://www.terraform.io/plugin/sdkv2/schemas/schema-behaviors#defaultfunc) and `schema.EnvDefaultFunc`

I'm not sure we can do this :thinking: 

We are using Terraform Plugin Framework and it seems there's no way to do this (yet?)
It was possible when using SDKv2 but can't find the docs/code on how to do this now

kazimierzbudzyk:
Ah yes you are right @marcoandredinis, need to update my tf knowledge. That's unfortunate, old way of doing it was pretty neat.

marcoandredinis:
@kazimierzbudzyk 
I'll create some tests around this feature
Can I take over your PR?

kazimierzbudzyk:
> @kazimierzbudzyk I'll create some tests around this feature Can I take over your PR?

Go for it @marcoandredinis. Appreciate that!

r0mant:
@marcoandredinis Yeah, let's just buddy-merge @kazimierzbudzyk's PR and add the tests ourselves.

r0mant:
@marcoandredinis Also, see above - looks like unit tests are failing, could you take a look and see if it's something caused by these changes?

marcoandredinis:
Checks are now passing
Just did a local end-to-end test and everything seems to be working

Will wait for additional approvals now

r0mant (approved):

PR #480, opened 2022-04-01
Update Teleport API dep

smallinsky:
## What

Update teleport API dep to fix incompatibility between terraform auth client creation  when tsh profile directory is created by  `tsh`>= v9 where CAS dir was introduced. 

Fix:
https://github.com/gravitational/teleport/pull/11567


r0mant (approved):

go.sum
```diff
@@ -256,6 +256,8 @@ github.com/gravitational/protobuf v1.3.2-0.20201123192827-2b9fcfaffcbf h1:MQ4e8X
 github.com/gravitational/protobuf v1.3.2-0.20201123192827-2b9fcfaffcbf/go.mod h1:SlYgWuQ5SjCEi6WLHjHCa1yvBfUnHcTbrrZtXPKa29o=
 github.com/gravitational/teleport/api v0.0.0-20220317023058-7bbe6f15c5c8 h1:DiTkx0vquZMgap4FpUCZ/jbxtDXiWNbh3NLU7KX62Tk=
```
Joerger, line 257: Can you `go mod tidy`? Looks like it's duplicating in the go.sum.

smallinsky: Indeed, Nice catch.

Joerger (approved):

PR #491, opened 2022-04-13
Disable CGO for terraform plugin

### LINKED ISSUE #447: terraform: support Alpine Linux
The makefile for the terraform provider explicitly sets `CGO_ENABLED=1`, which results in a build artifact that won't run on Alpine Linux containers (they lack glibc).

This prevents the use of GitLab CI with our terraform provider, as GitLab provides terraform [images based on Alpine](https://gitlab.com/gitlab-org/terraform-images).

If we don't actually require CGo in the provider, setting `CGO_ENABLED=0` might be all we need to do in order to support Alpine images.

Investigate whether the provider still functions with `CGO_ENABLED=0` and if an artifact compiled in this way will run in an Alpine container.

marcoandredinis:
We don't have a hard requirement to have CGO enabled for the terraform
plugin

Disabling it will generate a static binary, wich doesn't need glibc.

This will the usage of the plugin without depending on glibc
As an example: alpine and nixos distros

Fixes #447

### DIFF
--- a/terraform/Makefile
+++ b/terraform/Makefile
@@ -8,7 +8,7 @@ TFDIR ?= example
 
 ADDFLAGS ?=
 BUILDFLAGS ?= $(ADDFLAGS) -ldflags '-w -s'
-CGOFLAG ?= CGO_ENABLED=1
+CGOFLAG ?= CGO_ENABLED=0
 
 RELEASE = terraform-provider-teleport-v$(VERSION)-$(OS)-$(ARCH)-bin
 


Joerger (approved):

terraform/Makefile
```diff
@@ -8,7 +8,7 @@ TFDIR ?= example
 
 ADDFLAGS ?=
 BUILDFLAGS ?= $(ADDFLAGS) -ldflags '-w -s'
-CGOFLAG ?= CGO_ENABLED=1
+CGOFLAG ?= CGO_ENABLED=0
```
Joerger, line 11: You could just remove the `CGOFLAG`, unless you want it to be explicit.

marcoandredinis: I can remove the line entirely and be explicit in the build command, like so:

```
GOOS=$(OS) GOARCH=$(ARCH) CGO_ENABLED=0 go build -o $(BUILDDIR)/terraform-provider-teleport $(BUILDFLAGS)
```

Was this your suggestion?

zmb3 (approved):

r0mant (approved):

tcsc (approved):

PR #506, opened 2022-04-20
Fix sed command for BSD version

kbence:
The BSD version of the sed command is slightly different from the GNU version, so let's decide at the beginning which format to use based on the fact that BSD sed doesn't have a --version option.

This might sound a bit hacky, but it's actually more stable than testing for `uname -s` (which will only work for the systems we test for).

Tested with GNU sed on Linux and BSD sed on MacOS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,12 @@
+# Set up a system-agnostic in-place sed command
+IS_GNU_SED = $(shell sed --version 1>/dev/null 2>&1 && echo true || echo false)
+
+ifeq ($(IS_GNU_SED),true)
+	SED = sed -i
+else
+	SED = sed -i ''
+endif
+
 .PHONY: access-slack
 access-slack:
 	make -C access/slack
@@ -43,7 +52,7 @@ docker-push-access-%: docker-build-access-%
 # Pulls and pushes image from ECR to quay.
 .PHONY: docker-promote-access-%
 docker-promote-access-%:
-	$(MAKE) -C access/$* docker-promote 
+	$(MAKE) -C access/$* docker-promote
 
 # Build event-handler plugin with docker
 .PHONY: docker-build-event-handler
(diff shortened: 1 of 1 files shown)


kbence:
Execution looks like the following:

<details>
<summary>when executed on macOS</summary>
<pre>
$ make update-version VERSION=9.0.4
# Make sure VERSION is set on the command line "make update-version VERSION=x.y.z".
sed -i '' '1s/.*/VERSION=9.0.4/' event-handler/Makefile
sed -i '' '1s/.*/VERSION=9.0.4/' access/jira/Makefile
sed -i '' '1s/.*/VERSION=9.0.4/' access/mattermost/Makefile
sed -i '' '1s/.*/VERSION=9.0.4/' access/slack/Makefile
sed -i '' '1s/.*/VERSION=9.0.4/' access/pagerduty/Makefile
sed -i '' '1s/.*/VERSION=9.0.4/' access/email/Makefile
sed -i '' '1s/.*/VERSION=9.0.4/' terraform/install.mk
</pre>
</details>

<details>
<summary>when executed on Linux</summary>
<pre>
$  make update-version VERSION=9.0.4
# Make sure VERSION is set on the command line "make update-version VERSION=x.y.z".
sed -i '1s/.*/VERSION=9.0.4/' event-handler/Makefile
sed -i '1s/.*/VERSION=9.0.4/' access/jira/Makefile
sed -i '1s/.*/VERSION=9.0.4/' access/mattermost/Makefile
sed -i '1s/.*/VERSION=9.0.4/' access/slack/Makefile
sed -i '1s/.*/VERSION=9.0.4/' access/pagerduty/Makefile
sed -i '1s/.*/VERSION=9.0.4/' access/email/Makefile
sed -i '1s/.*/VERSION=9.0.4/' terraform/install.mk
</pre>
</details>

kbence:
One change that I can think of would be to rename the variable to `SED_I`, but that looked weird. @logand22, @wadells: what do you think?

wadells (approved):

wadells:
> One change that I can think of would be to rename the variable to `SED_I`, but that looked weird. @logand22, @wadells: what do you think?

`SED` is fine IMO. The real make patttern would be `SEDFLAGS` in addition to `SED` but that is absolute overkill for our purposes.

logand22 (approved):
LGTM! Thanks for this!

tcsc (approved):

PR #508, opened 2022-04-20
Helm chart for the Jira plugin

### LINKED ISSUE #497: Implement Helm chart for `teleport-plugin-jira`
- [x] Write tests (using `helm-unittest`)
- [x] Add it to the Drone build (build + upload to the staging and the production buckets)
- [x] Test the Helm chart by installing it and connecting it somewhere (eg. `access-dev.teleportinfra.dev`)
- [x] Document settings
- [x] Document installation

kbence:
This change adds a basic Helm chart for the Jira plugin

Closes #497


kbence:
What's still missing:
- Tested the chart with a Jira instance in Atlassian Cloud (it's possible to create a free account). There's a TLS handshake issue when Jira is calling back the plugin, but I wasn't able to figure out what's happening. The traffic have reached the pod but it the remote server (Jira) shut down the connection with sending a single byte through the TLS connection. It might be some configuration issue with the service or the load balancer it creates.
- The service should be more generic, I used the type `LoadBalancer` only for simplicity. `type`, `metadata.annotations` and `ports` should be configurable (and probably `metadata.labels` to help using the chart in Terraform and other scripts).
- We should test letting a request expire besides denying and approving the access request (it requires a fourth column called `Expired` as far as I remember).

wadells (approved):
LGTM. :ship:

charts/access/jira/templates/deployment.yaml
```diff
@@ -0,0 +1,80 @@
...
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        # - name: debug
+        #   image: kbence/auto-install
+        #   args: [tini, sleep, 1h]
```
wadells, line 30: This looks like leftover debugging code.

charts/access/jira/templates/service.yaml
```diff
@@ -0,0 +1,15 @@
+kind: Service
+apiVersion: v1
+metadata:
+  name: {{ include "jira.fullname" . }}
+  annotations:
+    service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: "true"
+    service.beta.kubernetes.io/aws-load-balancer-type: nlb
```
wadells, line 7: Can we abstract these cloud specific annotations and allow custom ones?

I'd look to https://github.com/gravitational/teleport/blob/78e2155268a94100b7763e729a7a70dbe3574e2e/examples/chart/teleport-cluster/templates/service.yaml#L10-L19 as an example here.

charts/access/jira/templates/service.yaml
```diff
@@ -0,0 +1,15 @@
...
+spec:
+  selector:
+    {{- include "jira.selectorLabels" . | nindent 4 }}
+  type: LoadBalancer
+  ports:
+  - name: https
+    port: 443
+    targetPort: 8443
```
wadells, line 15: This should be based off of listenAddress, right?  If that is updated, this will break.

More importantly, do we need to allow the service <-> pod port to be configured at all?  This seems like a knob folks are unlikely to tweak.  Instead, I'd guess customers would want to tweak the inbound port that Jira will hit.

charts/access/jira/values.yaml
```diff
@@ -0,0 +1,72 @@
...
+jira:
+  url: ""
+  username: ""
+  apiToken: ""
+  project: ""
+  issueType: "Task"
+
+http:
+  listenAddress: ":8443"
+  publicAddress: ""
```
wadells, line 61: See my earlier remark about which port customers are likely to change.  Users probably won't want to tweak the port between the service and the pod, but instead the port that the service is publicly available on.

wadells (approved):

wadells (changes requested):
See above feedback.

kbence:
Pulled this back to draft since it needs a bit more work.

charts/access/jira/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: "9.1.3"
```
wadells, line 18: We're up to 9.3.4 as of https://github.com/gravitational/teleport-plugins/commit/8894eb7556983be7cc1432b20939589131e7fe5d

charts/access/jira/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "9.1.3"
```
wadells, line 24: We're up to 9.3.4 as of https://github.com/gravitational/teleport-plugins/commit/8894eb7556983be7cc1432b20939589131e7fe5d

wadells (approved):

gozer (approved):

logand22 (approved):

Joerger (approved):

PR #512, opened 2022-04-22
Access/Email: add StartTLS Policy to SMTP config

### LINKED ISSUE #484: Support email plugin without TLS (to SMTP server)
### What
Email plugin currently can only use TLS communication with SMTP server. This request is to have option to start email plugin without TLS (to SMTP server).

### How
Provide option to disable TLS

### Why
This is good for testing this plugin with SMTP server which dont have TLS enabled.

### Workaround
N/A

marcoandredinis:
When sending an email over SMTP we can define how we communicate to the server.
It starts with plain text, but we can then upgrade the connection to using TLS,
before sending the actual email data

We can use:
- disabled/NoStartTLS: the communication will be in plain text between client
and server
- opportunistic: we'll try to use TLS, however if the server refuses to use an
encrypted channel, we fallback to plain text
- mandatory: we'll only send the email using TLS, if the server refuses, we
won't be sending anything

This PR aims to introduce such configuration
Previously we were always using the 'mandatory' option, which will continue to
be the default from now on.

Using the port 465 will override everything we said above and will always start
the connection using TLS
There's no plain text, neither a negotation to upgrade the connection

Fixes #484


access/email/config.go
```diff
@@ -175,9 +180,26 @@ func (c *SMTPConfig) CheckAndSetDefaults() error {
...
+func mailStartTLSPolicy(p string) (mail.StartTLSPolicy, error) {
+	switch p {
+	case "mandatory", "":
+		return mail.MandatoryStartTLS, nil
+	case "opportunistic":
+		return mail.OpportunisticStartTLS, nil
+	case "disabled":
+		return mail.NoStartTLS, nil
+	default:
+		return mail.MandatoryStartTLS, fmt.Errorf("wrong value '%s' - provide one of mandatory, opportunistic or disabled", p)
```
r0mant, line 199: Nit:
```suggestion
		return mail.MandatoryStartTLS, fmt.Errorf("unsupported starttls_policy %q - provide one of mandatory, opportunistic, disabled or leave empty to default to mandatory", p)
```

r0mant (approved):

Joerger (approved):

Issue #524, opened 2022-04-28
Update drone build process to use correct Go version on Darwin


tcsc:
The current plugins build process for Darwin uses the system-installed Go version, which is currently 1.16.9. 

This means that the plugins are being build with a version of go that is _unexpected_ by the developers, and potentially unpatched for security vulnerabilities.

This is currently causing the Darwin build of the plugins to fail

PR #525, opened 2022-04-28
Makes Darwin builder use the release-appropriate go version

tcsc:
The Darwin build and test scripts now download and install a release-appropriate version of Go before building and testing the code.

See-Also: #524

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -59,26 +59,41 @@ trigger:
 workspace:
   path: /tmp/teleport-plugins/test-darwin
 
+environment:
+  GOLANG_VERSION: go1.17.9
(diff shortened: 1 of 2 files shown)


r0mant (approved):

.drone.yml
```diff
@@ -59,27 +59,42 @@ trigger:
 workspace:
   path: /tmp/teleport-plugins/test-darwin
 
+environment:
+  GOLANG_VERSION: go1.17.9
```
wadells, line 63: This is called `GO_VERSION` in most/all the plugin Makefiles.  It'd be nice to be consistent with that for ripgrep.

wadells (approved):

.drone.yml
tcsc: Definitely. Changing.

PR #526, opened 2022-04-29
Move Darwin build env from pipeline to build step

tcsc:
It turns out that pipeline-level environment settings are only valid for
docker pipelines, but drone will still happily accept the syntax in the
YAML without complaint.

See-Also: #524
See-Also: #525


r0mant (approved):

zmb3 (approved):
This has bit me so many times.

wadells (approved):

tcsc:
> This has bit me so many times.

It makes everything so much cleaner, but no: Drone says we can't have the nice things.

PR #530, opened 2022-04-29
Create temporary TF provider zipfile in destination dir

tcsc:
Repacking the zipfile is failing on Drone due to the `/tmp`  and
`workspace` directories being on separate devices.

This patch makes makes the repacking routine create the temporary
zipfile in the destination directory to minimize the chance that the
files are on different devices.

### DIFF
--- a/tooling/internal/terraform/registry/sign.go
+++ b/tooling/internal/terraform/registry/sign.go
@@ -77,7 +77,14 @@ func RepackProvider(dstDir string, srcFileName string, signingEntity *openpgp.En
 	}
 	defer src.Close()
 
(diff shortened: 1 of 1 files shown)


smallinsky (approved):
Alternative approach will be to use own file Copy implementation instead of `os.Rename` call but the`RepackProvider` function is used internally  `tooling/cmd`  so the fix LGTM.

espadolini (approved):
I would argue that this is the correct way; copying the file instead of renaming it would be slower to boot, and if we don't have permission or there's not enough space in the destination directory then the operation would fail at the end anyway.

zmb3 (approved):

PR #547, opened 2022-05-11
Add rolemap support to teleport-plugin-email chart

kbence:
Extends #459 with the new configuration value added in #488

## Testing

Tested with a MailGun sandbox instance using both the legacy (`delivery.recipients`) and the new setting, works as expected.

Related to #439

### DIFF
--- a/charts/access/email/tests/__snapshot__/configmap_test.yaml.snap
+++ b/charts/access/email/tests/__snapshot__/configmap_test.yaml.snap
@@ -7,10 +7,11 @@ should match the snapshot (mailgun on):
         addr = "teleport.example.com:1234"
         identity = "/var/lib/teleport/plugins/email/auth_id"
 
-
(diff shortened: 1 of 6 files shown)


charts/access/email/tests/__snapshot__/configmap_test.yaml.snap
```diff
@@ -135,6 +136,41 @@ should match the snapshot (smtp on, starttls disabled):
         addr = "teleport.example.com:1234"
         identity = "/var/lib/teleport/plugins/email/auth_id"
 
+        [smtp]
+        host          = "smtp.someservice.com"
```
wadells, line 140: Lets use example.com per https://datatracker.ietf.org/doc/html/rfc2606

wadells (approved):

Joerger (approved):

PR #550, opened 2022-05-12
terraform: provider fails to create resource with cache on

### LINKED ISSUE #513: Terraform "Error reading App"
I tried the "teleport_app" example ([app.tf.example](https://github.com/gravitational/teleport-plugins/blob/master/terraform/example/app.tf.example)) and get the following error:

```
Terraform will perform the following actions:

  # teleport_app.example will be created
  + resource "teleport_app" "example" {
      + id       = (known after apply)
      + kind     = (known after apply)
      + metadata = {
          + description = "Test app"
          + labels      = {
              + "teleport.dev/origin" = "dynamic"
            }
          + name        = "example"
          + namespace   = (known after apply)
        }
      + spec     = {
          + uri = "localhost:3000"
        }
      + version  = (known after apply)
    }

Plan: 1 to add, 0 to change, 0 to destroy.


Do you want to perform these actions in workspace "terraform-fsn1"?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

teleport_app.example: Creating...
╷
│ Error: Error reading App
│
│   with teleport_app.example,
│   on teleport.tf line 1, in resource "teleport_app" "example":
│    1: resource "teleport_app" "example" {
│
│ Terraform user has no rights to perform this action. Check that Terraform
│ user role has ['list','create','read','update','delete'] verbs for 'app'
│ resource.
│ 	application "example" doesn't exist
╵
```

The terraform role certainly has the correct verbs (also took that straight from the example):

```yaml
kind: role
metadata:
  id: 1650840248224075652
  name: terraform
spec:
  allow:
    app_labels:
      '*': '*'
    db_labels:
      '*': '*'
    db_names:
    - '*'
    db_users:
    - '*'
    rules:
    - resources:
      - user
      - role
      - token
      - trusted_cluster
      - github
      - oidc
      - saml
      - cluster_auth_preference
      - cluster_networking_config
      - session_recording_config
      - app
      - db
      verbs:
      - list
      - create
      - read
      - update
      - delete
  deny: {}
  options:
    cert_format: standard
    desktop_clipboard: true
    enhanced_recording:
    - command
    - network
    forward_agent: false
    max_session_ttl: 30h0m0s
    port_forwarding: true
    record_session:
      desktop: true
version: v5
```

Also, the app seems to be created nonetheless. But it isn't "visible" to the terraform provider on subsequent runs (it always wants to re-create the app, but fails because it exists)

```
$ tctl get app
kind: app
metadata:
  description: Test app
  id: 1650841107014304941
  labels:
    teleport.dev/origin: dynamic
  name: example
spec:
  insecure_skip_verify: false
  uri: localhost:3000
version: v3
```

Terraform Version: 1.1.9 (Terraform Cloud / linux_amd64)
Provider Version: 9.1.0 

gz#5009

### LINKED ISSUE #540: Terraform provider reports resource failed to create even when it was created successfully
Given the following resource:

```yaml
resource "teleport_app" "app" {
  metadata = {
    name        = "example"
    description = "simple example"
    labels = {
      "teleport.dev/origin" = "dynamic"
      "test" = "true"
    }
  }

  spec = {
    insecure_skip_verify = true
    uri         = "http://example.com"
  }
}
```

This type of error can happen:

```
│ Error: Error reading App
│
│   with teleport_app.app,
│   on main.tf line 84, in resource "teleport_app" "app":
│   84: resource "teleport_app" "app" {
│
│ Terraform user has no rights to perform this action. Check that Terraform user role has ['list','create','read','update','delete'] verbs for 'app' resource.
│ 	application "example" doesn't exist
╵
```

After this error, the app/example object definitely exists in the teleport backend.

I believe this section of the code is what throws the error: https://github.com/gravitational/teleport-plugins/blob/master/terraform/provider/resource_teleport_app.go#L104-L109

After eliminating the teleport rbac system and any user/role issues out, I figured there must be a race condition happening. The original environment where this was reported is using a dynamodb backend, which can have a very small window after a write before subsequent reads are successful.

I was able to force this to happen reliably in a lab. I set up a teleport cluster with a single replica etcd, and multiple teleport auth+proxy replicas. I then degraded the performance of etcd by running `tc qdisc add dev eth0 root netem delay 4000ms` inside the container where etcd is running, which adds four seconds of latency. The cluster still works, but every operation to etcd takes much longer. Once I did this, I was able to reliably reproduce the error above.

I tried to get a similar error from creating a role or user, but wasn't able to. I did, however, get a similar error when I marked a teleport_role resource as tainted and reran terraform. The delete succeeded, but the subsequent create failed because the `Client.GetRole` call returned the resource that had just been deleted.

```
╷
│ Error: Role exists in Teleport
│
│   with teleport_role.terraform-test,
│   on main.tf line 20, in resource "teleport_role" "terraform-test":
│   20: resource "teleport_role" "terraform-test" {
│
│ Role exists in Teleport. Either remove it (tctl rm role/terraform-test) or import it to the existing state (terraform import teleport_app.terraform-test terraform-test)
╵
```

I figure the terraform provider should have some retries built in to help the backend settle, especially since teleport can use backends that are eventually consistent. A very simple retry+exponential backoff on the step that tries to get the newly created resource could help ensure that new resource is properly settled. A similar get attempt may be appropriate after the resource delete is called to help a case where terraform is replacing a resource.

gz#5009

marcoandredinis:
When creating an app, if the Teleport Auth service has cache enabled,
the provider will fail but the app is created anyways.

This happens because when the provider receives the CreateApp
instruction from terraform, it will:
- check if app exists, if it does: error
- create the app
- get the app state to store it locally (in tf state)

Between the second and third step, we will most likely get a `not found`
error, because the app is not yet visible due to caching
eviction/refresh constraints.

However, given the second step (app creation) was executed succefully,
the app already exists in Teleport

This should happen for other resources as well, however, we are only
adding a test to the app as an example (the fix will be applied to every
resource)

To fix this, we are going to add a retry handler that retries when:
a) we just created the resource
and b) teleport's client returns a `not found` error

We didn't set any limit on the number of retries because the terraform sdk
already provides a deadline for completing the operation.

Fixes #540
Fixes #513


lib/testing/integration/integration.go
```diff
@@ -264,7 +264,7 @@ func (integration *Integration) Version() Version {
 }
 
 // NewAuthService creates a new auth server instance.
-func (integration *Integration) NewAuthService() (*AuthService, error) {
+func (integration *Integration) NewAuthService(cacheEnabled bool) (*AuthService, error) {
```
r0mant, line 267: Instead of passing bool flag, can you make this into a functional "option" argument? Something like

```go
type AuthServiceOption func(yaml string) string

func WithCache() AuthServiceOption {
  return func(yaml string) string {
    return strings.Replace(yaml, "{{TELEPORT_CACHE_ENABLED}}, "true")
  }
}

func (integration *Integration) NewAuthService(opts ...AuthServiceOption) (*AuthService, error) { ... }
```

Then you won't need to modify existing tests also, or explain what passing "true" or "false" means.

terraform/provider/provider.go
```diff
@@ -181,6 +204,8 @@ func (p *Provider) Configure(ctx context.Context, req tfsdk.ConfigureProviderReq
 	profileDir := p.stringFromConfigOrEnv(config.ProfileDir, "TF_TELEPORT_PROFILE_PATH", "")
 	identityFilePath := p.stringFromConfigOrEnv(config.IdentityFilePath, "TF_TELEPORT_IDENTITY_FILE_PATH", "")
 	identityFile := p.stringFromConfigOrEnv(config.IdentityFile, "TF_TELEPORT_IDENTITY_FILE", "")
+	retryBaseDurationStr := p.stringFromConfigOrEnv(config.IdentityFile, "TF_TELEPORT_RETRY_BASE_DURATION", "1ms")
+	retryCapDurationStr := p.stringFromConfigOrEnv(config.IdentityFile, "TF_TELEPORT_RETRY_CAP_DURATION", "1s")
```
r0mant, line 208: Nit: 1ms/1s seems a little aggressive TBH, I'd probably start with 1s and have a cap of 5s or something like this.

terraform/provider/resource_teleport_app.go
```diff
@@ -102,7 +103,21 @@ func (r resourceTeleportApp) Create(ctx context.Context, req tfsdk.CreateResourc
 	}
 
 	id := app.Metadata.Name
-	appI, err := r.p.Client.GetApp(ctx, id)
+	var appI apitypes.Application
+
+	backoff := backoff.NewDecorr(r.p.RetryConfig.Base, r.p.RetryConfig.Cap, clockwork.NewRealClock())
```
r0mant, line 108: What is the maximum retry duration? I.e. after how long will we give up retrying?

terraform/_gen/plural_resource.go.tpl
```diff
@@ -20,16 +20,18 @@ package provider
...
 	"crypto/rand"
 	"encoding/hex"
-	{{end}}
+{{- end}}
 
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
+	"github.com/jonboulle/clockwork"
```
r0mant, line 32: Nit: Move this import to the 3rd group (third-party imports).

terraform/_gen/singular_resource.go.tpl
```diff
@@ -23,7 +23,9 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/jonboulle/clockwork"
```
r0mant, line 26: Nit: Move this import to the 3rd group (third-party imports).

Joerger (commented):
LGTM with Roman's suggestions

lib/testing/integration/integration.go
marcoandredinis: That's awesome, thank you

terraform/provider/resource_teleport_app.go
marcoandredinis: I didn't set any limit here
This delegates the actual maximum to the one coming from Terraform (trough the `context.Done()`

The sdkv2 has a timeout of 20 minutes
https://www.terraform.io/plugin/sdkv2/resources/retries-and-customizable-timeouts#default-timeouts-and-deadline-exceeded-errors

However, we are using plugin framework and not sdkv2, which seems to have a higher timeout
image

I guess we have two options here:
- let it run anyway and wait for the parent executor (whoever calls the `terraform apply`) to cancel the request
- have an explicit configurable timeout or max number of retries

I'll add a max number of retries

Joerger (approved):

marcoandredinis:
Hey @kbence
The `Teleport-Plugins-Test-Helm-Charts (ci-account)` CI step is failing for this PR, because of a mismatch in the email access plugin in the `charts/access/email/tests/configmap_test.yaml` file

I can update the snapshots, but I can it can be worthwhile to look into the current state in order to understand what happened 

It also failed in this PR https://github.com/gravitational/teleport-plugins/pull/547 (but somehow, we managed to merge the PR anyway 🤔)

terraform/provider/provider.go
```diff
@@ -181,6 +206,9 @@ func (p *Provider) Configure(ctx context.Context, req tfsdk.ConfigureProviderReq
 	profileDir := p.stringFromConfigOrEnv(config.ProfileDir, "TF_TELEPORT_PROFILE_PATH", "")
 	identityFilePath := p.stringFromConfigOrEnv(config.IdentityFilePath, "TF_TELEPORT_IDENTITY_FILE_PATH", "")
 	identityFile := p.stringFromConfigOrEnv(config.IdentityFile, "TF_TELEPORT_IDENTITY_FILE", "")
+	retryBaseDurationStr := p.stringFromConfigOrEnv(config.IdentityFile, "TF_TELEPORT_RETRY_BASE_DURATION", "1s")
+	retryCapDurationStr := p.stringFromConfigOrEnv(config.IdentityFile, "TF_TELEPORT_RETRY_CAP_DURATION", "5s")
+	maxTriesStr := p.stringFromConfigOrEnv(config.IdentityFile, "TF_TELEPORT_RETRY_MAX_TRIES", "4")
```
r0mant, line 211: Nit: Let's maybe set it to 10 by default?

r0mant (approved):

PR #556, opened 2022-05-20
Terraform: Provider fails to update resource with cache on

marcoandredinis:
This one is very similar to 
https://github.com/gravitational/teleport-plugins/commit/6fc6288310554e70820b18cdbad3814258cf7d53
https://github.com/gravitational/teleport-plugins/pull/550
(please read the above commit/PR to have more context)

This is a variation of the issue above, but now for the Update feature
After updating a resource, we fetch its current state to save as TF state
However, the `fetch` was returning a cached version so the TF state would contain that invalid state.

The fix, we retry until we know the resource was updated:
- get resource: vA
- update
- loop
  - get the resource: vB
  - if the id of both versions is the same (`vA.id == vB.id`) we sleep for a bit and loop
  - if they are the same, we break the loop
 - save the resource's state into tf state

We can't know for sure that the new version is ours
If we have someone else update that same role, the new version can be ours or theirs
However, the likely of this happening should be low

This same problem could occur for the solution we applied when creating a resource
When we fetch by name, we could be receiving a version created by someone else or created by us but already in a new state.


marcoandredinis:
image
Trust me, sometimes it does fail
Cache works in mysterious ways 🤷‍♂️

terraform/_gen/main.go
```diff
@@ -40,6 +40,10 @@ type payload struct {
 	Kind string
 	// DefaultVersion represents the default resource version on create
 	DefaultVersion string
+	// StaticID states whether this particular resource has a static (usually 0) Metadata.ID
+	// This is relevant to cache enabled clusters: we use Metadata.ID to check if the resource was updated
+	// Currently, the resources that don't have a dynamic Metadata.ID are strong consistent: oidc, github and saml connectors
+	StaticID bool
```
tcsc, line 46: Can this be `HasStaticID` - I'd read `StaticID` as the static ID of something, rather than a predicate.

terraform/_gen/plural_resource.go.tpl
```diff
@@ -240,18 +240,44 @@ func (r resourceTeleport{{.Name}}) Update(ctx context.Context, req tfsdk.UpdateR
...
+		if {{.VarName}}Before.GetMetadata().ID != {{.VarName}}I.GetMetadata().ID || {{.StaticID}} {
+			break
+		}
+
+		if err := backoff.Do(ctx); err != nil {
+			resp.Diagnostics.Append(diagFromWrappedErr("Error reading {{.Name}}", trace.Wrap(err), "{{.Kind}}"))
+			return
+		}
+		if tries >= r.p.RetryConfig.MaxTries {
+			diagMessage := fmt.Sprintf("Error reading {{.Name}} (tried %d times)", tries)
```
tcsc, line 275: Are we fine with the calling code just carrying on without being able to detect the error in this scenario?

terraform/test/auth_preference_test.go
```diff
@@ -92,3 +104,38 @@ func (s *TerraformSuite) TestImportAuthPreference() {
...
+		Steps: []resource.TestStep{
+			{
+				Config: s.getFixture("auth_preference_0_cluster.tf"),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(name, "kind", "cluster_auth_preference"),
+					resource.TestCheckNoResourceAttr(name, "metadata.labels"),
+					resource.TestCheckResourceAttr(name, "spec.type", "oidc"),
+				),
+			},
+			// {
```
tcsc, line 123: What's up with these commented-out bits?

terraform/tfschema/time_type.go
```diff
@@ -106,6 +106,10 @@ func (t TimeType) ValueFromTerraform(ctx context.Context, in tftypes.Value) (att
 func (t TimeType) Validate(ctx context.Context, in tftypes.Value, path *tftypes.AttributePath) diag.Diagnostics {
 	var diags diag.Diagnostics
 
+	if in.Type() == nil {
```
tcsc, line 109: How does this change relate to the retry changes above?

terraform/_gen/main.go
marcoandredinis: Yes, sounds better
Thank you

terraform/test/auth_preference_test.go
marcoandredinis: Totally forgot to stage that change into the git commit
Sorry

terraform/tfschema/time_type.go
marcoandredinis: It's related with the second linked issue
It ended up being something simple and triggered when the cache is enabled

I'm fine opening up a PR for this particular change

terraform/_gen/plural_resource.go.tpl
marcoandredinis: Carrying on means we diverge the Teleport and TF state, which is not ideal.

A possible solution would be to apply some compensation measures, like trying to rollback to the `Before` state

We would call the rollback in every function exit (with error) after we update the resource's state in teleport.

However, this will lead to some complex situations: what if we fail to rollback?
Or what if we discard a hypothetical state applied while we were retrying?

Do you think it's worth exploring this idea? Or do you have another suggestion?

marcoandredinis:
> image
Trust me, sometimes it does fail Cache works in mysterious ways man_shrugging

This happened again where
https://github.com/gravitational/teleport-plugins/pull/560

We were not actually running the TF acceptance tests when running `make tests` using the root's Makefile
:facepalm:

Joerger (approved):

terraform/_gen/plural_resource.go.tpl
tcsc: My point was (I think - it's been a while...) that we don't signal error to the caller in any way (that I can see). I'm _fine_ with it bailing out here - I just wonder what is going to happen if the caller thinks things have worked when they haven't.

terraform/tfschema/time_type.go
tcsc: Ah, gotcha.

tcsc (approved):

r0mant (approved):

PR #560, opened 2022-05-25
Terraform: fix retry variables lookup

### LINKED ISSUE #558: new retry values can't be set when identity_file is set
The validation logic at https://github.com/gravitational/teleport-plugins/blob/6fc6288310554e70820b18cdbad3814258cf7d53/terraform/provider/provider.go#L209-L211 is trying to set retryBaseDurationStr, retryCapDurationStr, and maxTriesStr based on the value of config.IdentityFile if it is set. This means that if the toml file has config.IdentityFile set, the logs print an error about the data being incorrect and it prints the identityfile contents to logging.

This appears to have been introduced by the the changes for #550

Instructions to reproduce:
---

**main.tf**:
```hcl
provider "teleport" {
  identity_file = file("/tmp/terraform-identity")
  addr = "teleport.example.com:443"
}

...
```

attempting to run this results in the following output:

```
$ terraform refresh
╷
│ Error: Failed to parse Retry Base Duration
│
│   on main.tf line 10, in provider "teleport":
│   10: provider "teleport" {
│
│ Please check if retry_cap_duration (or TF_TELEPORT_RETRY_BASE_DURATION) is set correctly. Error: time: invalid duration "-----BEGIN RSA
│ PRIVATE KEY-----
│ MIIEowIBAAKCAQEAv+CVt3O4Jp2ybnogHn414V8tEAeNepq2Cy9o8a305AnUDIdl
  ...
│ J/JE24ya8jS/9ZG8crxz2TwRmPAq
│ -----END CERTIFICATE-----
│ "
╵
```

Workaround: use `identity_file_path` instead of `identity_file`

gz#5009

marcoandredinis:
This PR added a configurable retry mechanism
https://github.com/gravitational/teleport-plugins/pull/550

However, it was trying to read the value from another config property: IdentityFile

Fixes #558


marcoandredinis:
Created an issue to followup on running Integration/Acceptance tests
https://github.com/gravitational/teleport-plugins/issues/561

I ran the tests locally (`make -C terraform test`) and they passed

programmerq:
I don't believe the current integration tests would have caught this issue since they always seem to use the identity_file_path approach: https://github.com/gravitational/teleport-plugins/blob/6fc6288310554e70820b18cdbad3814258cf7d53/terraform/test/main_test.go#L156-L161

Changing line 159 to this will cause test failures:

```golang
    identity_file = file("` + s.teleportConfig.Identity + `")
```

Having at least some of the tests run with `identity_file` instead would probably be a good idea since both code paths are valid.

terraform/test/main_test.go
```diff
@@ -156,7 +156,10 @@ func (s *TerraformBaseSuite) SetupSuite() {
 	s.terraformConfig = `
 		provider "teleport" {
 			addr = "` + s.teleportConfig.Addr + `"
-			identity_file_path = "` + s.teleportConfig.Identity + `"
+			identity_file = file("` + s.teleportConfig.Identity + `")
```
programmerq, line 159: Since both `identity_file` and `identity_file_path` are valid code paths, the tests should cover both.

Joerger (approved):

programmerq (approved):
lgtm!

r0mant (approved):

PR #563, opened 2022-05-29
Add missed support for automated unittests

gozer:
Also: fix failing unittest


zmb3 (approved):

kbence (approved):
Oops, I missed those. Thanks for spotting it!

PR #566, opened 2022-05-30
Add missing plugins to helm unittests

kbence:
These Helm chart tests were missing from the PR builds


gozer (approved):
All tests are passing

PR #568, opened 2022-05-31
Add `terraform` to CI/CD and run TF acceptance tests

### LINKED ISSUE #561: Add Integration Tests as a new step in CI/CD
We only have steps for lint and unit tests

As an example: we are not running Terraform acceptance tests in the pipeline

marcoandredinis:
This PR adds the `terraform` binary which is necessary to run Terraform acceptance tests

We also change the `make tests` target to call the `make tests` inside Terraform which sets the flag to run Terraform Acceptance Tests (`TF_ACC=true`)

Fixes #561

### DIFF
--- a/.cloudbuild/ci/unit-tests-linux.yaml
+++ b/.cloudbuild/ci/unit-tests-linux.yaml
@@ -11,7 +11,7 @@ steps:
         curl -fsSL https://apt.releases.hashicorp.com/gpg | apt-key add - &&
         apt update && apt install -y software-properties-common &&
         apt-add-repository "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main" &&
-        apt update && apt install -y terraform &&
+        apt update && apt install -y terraform=1.2.2 &&
         make test
 
 availableSecrets:
--- a/terraform/Makefile
+++ b/terraform/Makefile
@@ -71,7 +71,7 @@ endif
 release: build
 	tar -C $(BUILDDIR) -czf $(RELEASE).tar.gz .
 
-TERRAFORM_EXISTS := $(shell terraform -version | grep 'Terraform v1.')
+TERRAFORM_EXISTS := $(shell terraform -version | grep 'Terraform v1.2')
 CURRENT_ULIMIT := $(shell ulimit -n)
 
 .PHONY: test


.cloudbuild/ci/unit-tests-linux.yaml
```diff
@@ -5,7 +5,15 @@ steps:
...
       - TELEPORT_ENTERPRISE_LICENSE
     entrypoint: /bin/bash
-    args: ["-c", "make test"]
+    args:
+      - "-c"
+      - |
+        curl -fsSL https://apt.releases.hashicorp.com/gpg | apt-key add - &&
+        apt update && apt install -y software-properties-common &&
+        apt-add-repository "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main" &&
+        apt update && apt install -y terraform &&
```
wadells, line 14: I'd strongly recommend pinning the terraform version.  For apt:

```
apt install -y terraform=1.1.9
```
(or 1.2.2 for the most recent version -- I use an older one to show that it works)


Alternatively:
```
curl -o terraform.zip https://releases.hashicorp.com/terraform/${TF_VERS}/terraform_${TF_VERS}_linux_386.zip
unzip terraform.zip
mv terraform /usr/bin/terraform 
```

Without pinning we may see tests start to fail in this repo due to upstream changes. These will then block completely unrelated PRs and scatterblast the work of fixing current code to match the latest upstream.  With the version pinned, the necessary local changes can be added in the same PR that updates the pin.

marcoandredinis: Thank you
You are totally right, it could break for unrelated changes
I will pin the latest version (1.2.2)

What do you think should be the makefile check for Terraform Version?
Should I check for the same minor we use in the cicd?

It used to check for `Terraform v1.1`, but I changed to only check for `Terraform v1.`
But now I wonder, if we should be more strict and match against `Terraform v1.2`

wadells (approved):

.cloudbuild/ci/unit-tests-linux.yaml
wadells: > What do you think should be the makefile check for Terraform Version?
> Should I check for the same minor we use in the cicd?

Matching the minor version sounds reasonable.  Ideally we'd force the exact version, but I don't think we have strong enough automation around the toolchain to expect this for all developers.

> Thank you You are totally right, it could break for unrelated changes I will pin the latest version (1.2.2)
> 
> What do you think should be the makefile check for Terraform Version? Should I check for the same minor we use in the cicd?
> 
> It used to check for `Terraform v1.1`, but I changed to only check for `Terraform v1.` But now I wonder, if we should be more strict and match against `Terraform v1.2`

I think matching minor version is a good compromise for now.  Ideally we'd match the full version, but we don't have good toolchain management in place for this yet, and I don't want to scope creep this improvement to solve that more difficult problem.

Joerger (approved):

PR #581, opened 2022-06-09
feat: Also index staging s3 helm bucket

gozer:


kbence:
Just for clarification: currently, the Drone job [publish-helm-charts](https://github.com/gravitational/teleport-plugins/blob/master/.drone.yml#L697) generates and uploads the index to S3 upon promotion. The following steps are executed:

1. Download artifacts from production bucket
2. Download artifacts from staging bucket
3. Generate repository index
4. Publish directory to production bucket

With this change, the index file from the production bucket will be overwritten in the second step, but it's not a problem since we regenerate it in the 3rd step anyways, so there's no way it'll be published with the wrong index file.

kbence (approved):
👍

.drone.yml
```diff
@@ -306,12 +306,29 @@ trigger:
...
+    image: amazon/aws-cli
+    environment:
+      AWS_S3_BUCKET:
+        from_secret: HELM_CHART_STAGING_BUCKET
+      AWS_ACCESS_KEY_ID:
+        from_secret: PLUGIN_BUILD_USER_STAGING_KEY
+      AWS_SECRET_ACCESS_KEY:
+        from_secret: PLUGIN_BUILD_USER_STAGING_SECRET
+    commands:
+      - aws s3 sync s3://$AWS_S3_BUCKET/ packages/
```
gozer, line 319: I guess to avoid pulling down the stage index, one could add
```--exclude index.yaml``` here

tcsc (approved):

Joerger (approved):

PR #582, opened 2022-06-14
Release 9.3.5

r0mant:


tcsc (approved):

r0mant:
Looks like there are still some issues with revendoring the api module, going to keep it at previous version while we're figuring it out.

PR #584, opened 2022-06-15
Improve secret management in Helm charts

### LINKED ISSUE #573: Make sure no secrets are stored in the Helm chart's ConfigMaps
Some of the charts store secrets in ConfigMaps. Let's generate secrets instead of them (eg. configs).

Another solution could be to use some templating mechanism to replace strings in the generated ConfigMaps during startup, but this might increase the amount of work significantly.

kbence:
The following changes were made in the charts:
- Now the Helm charts store all secrets in Kubernetes Secret in a compatible manner
- Added `*FromSecret` and `*SecretPath` variables to support reading passwords and tokens from Secrets created outside of the Helm charts
- Updated/fixed `values.schema.json` files to reflect these changes
- Added documentation for the above changes

Closes #573

### DIFF
--- a/charts/access/email/templates/deployment.yaml
+++ b/charts/access/email/templates/deployment.yaml
@@ -53,7 +53,8 @@ spec:
               {{- if .Values.mailgun.enabled }}
               mountPath: "/var/lib/teleport/plugins/email/mailgun_private_key"
               subPath: "{{ .Values.mailgun.privateKeySecretPath }}"
-              {{- else }}
+              {{- end }}
+              {{- if .Values.smtp.enabled }}
               mountPath: "/var/lib/teleport/plugins/email/smtp_password"
               subPath: "{{ .Values.smtp.passwordSecretPath }}"
               {{- end }}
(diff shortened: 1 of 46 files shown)


charts/access/email/templates/deployment.yaml
```diff
@@ -48,6 +48,16 @@ spec:
             - name: auth-id
               mountPath: /var/lib/teleport/plugins/email/auth_id
               subPath: {{ .Values.teleport.identitySecretPath }}
+            {{- if or .Values.mailgun.enabled .Values.smtp.enabled }}
+            - name: {{ .Values.secretVolumeName }}
+              {{- if .Values.mailgun.enabled }}
+              mountPath: "/var/lib/teleport/plugins/email/mailgun_private_key"
+              subPath: "{{ .Values.mailgun.privateKeySecretPath }}"
+              {{- else }}
```
gozer, line 56: Somewhat of a nit, but from a logical pov, I prefer for readability and future-proofing structure like

```
if A and B:
  [...]
  if A
    [...]
  endif
  if B
    [...]
  endif
endif
```

gozer (approved):
Large diff, but LGTM overall

kbence:
> Large diff, but LGTM overall

Yeah, I didn't realize that the snapshots will make it this huge.

logand22 (approved):

Joerger (approved):

PR #585, opened 2022-06-15
Bump Teleport API to 9.3.6

marcoandredinis:
This PR updates the Teleport API to 9.3.6

We had to do some extra changes aside from the version bump itself:
- override field name in OIDC spec: RedirectURLs


lib/tsh/tsh.go
```diff
@@ -79,6 +79,7 @@ func (tsh Tsh) Bench(ctx context.Context, flags BenchFlags, userHost, command st
 	log.Debugf("Running %s", cmd)
 	outputBytes, err := cmd.CombinedOutput()
 	output := string(outputBytes)
+	log.Println(output)
```
r0mant, line 82: Debug leftover?

terraform/protoc-gen-terraform-teleport.yaml
```diff
@@ -134,6 +134,24 @@ exclude_fields:
...
+    # We created a new field `RedirectURLs` but its jsontag value is `redirect_url`
+    # `redirect_url` already existed as `RedirectURL` and its jsontag is now `-`
+    # The generator creates two fields with the same name:
+    # - RedirectURLs has jsontag `redirect_url`, so it creates an entry for `redirect_url`
+    # - RedirectURL has jsontag `-`, so it converts the name into snake_case, creating another entry for `redirect_url`
+    #
+    # The solution applied here is to rename `RedirectURLs` from `redirect_url` to `redirect_urls`
+    # This way we don't break configurations with a minor release.
+
+    # At v10 - we will change the type of the `redirect_url` from string to list of strings <- we break user's configuration here
```
r0mant, line 151: > we break user's configuration here

Could you clarify this, what will break?

How does this change affect terraform users that have existing OIDC connectors or create new ones?

terraform/tfschema/custom_types.go
```diff
@@ -193,3 +193,53 @@ func CopyToTraits(diags diag.Diagnostics, o wrappers.Traits, t attr.Type, v attr
...
+func GenSchemaStrings(_ context.Context) tfsdk.Attribute {
+	return tfsdk.Attribute{
+		Optional: true,
+		Type: types.ListType{
+			ElemType: types.StringType,
+		},
+	}
+}
+
+func CopyFromStrings(diags diag.Diagnostics, v attr.Value, o *wrappers.Strings) {
```
r0mant, line 207: I may be missing something, but looking at the implementation, should not this function be called `CopyToStrings` and the one below `CopyFromStrings` (i.e. inverted)?

Also, please add godocs on both.

r0mant (commented):
I think we should verify a few scenarios to make sure we don't break existing customer states:

- [ ] Existing OIDC connector with a single redirect_url.
- [ ] Creating new OIDC connector with redirect_url containing a single entry (string, not array).
- [ ] Creating new OIDC connector with redirect_url containing an array.

lib/tsh/tsh.go
marcoandredinis: Yeah, sorry
Still trying to figure out what's the issue with using the `tsh ssh/bench` commands
I was trying to understand if the error I got locally was the same as in the pipeline (it is!)

terraform/tfschema/custom_types.go
marcoandredinis: It's a bit confusing because we lack the `source/destination` in the method name
But I think it's correct, I followed the other `CopyFromXYZ`, `CopyToXYZ` implementations

terraform/protoc-gen-terraform-teleport.yaml
marcoandredinis: Terraform does not allow for a field to have two types.
So, in this case, the field `redirect_url` in terraform must be a string **or** a list of strings.

As of `v<9.3.5`, its type is a String and we can't change that in a minor release.

[Teleport API 9.3.5 added](https://github.com/gravitational/teleport/pull/13046) a new field that we also must support: RedirectURLs
So, we added a new field: `redirect_urls`

So, with this PR (v9.3.5+) we have:
```
redirect_url: <string> # maps to RedirectURL
redirect_urls: <list of strings> # maps to RedirectURLs
```

The plan is to change `redirect_url`'s type from string to list of strings at most at v11.0.0 (I previously said v10)
Why v11? That's when the RedirectURL field is going away.
When it happens, users that were still using `redirect_url` field, must change it 
```tf
redirect_url: "https://example.org/redirect"
```
to
```tf
redirect_url: ["https://example.org/redirect"]
```

I tried to reword the explanation in the file
Please let me know if it makes sense

Joerger: > Terraform does not allow for a field to have two types.
So, in this case, the field redirect_url in terraform must be a string or a list of strings.

In this case, I don't think we should aim for the end result of:
```tf
redirect_url: ["https://example.org/redirect"]
```

Can we instead support the following, with no further deprecation steps necessary:
```
redirect_url: <string> # maps to RedirectURLs[0]
redirect_urls: <list of strings> # maps to RedirectURLs - takes precedence over redirect_url
```

terraform/test/fixtures/oidc_connector_2_update_only_redirect_urls.tf
```diff
@@ -0,0 +1,21 @@
...
+    spec = {
+        client_id = "client"
+        client_secret = "value"
+    
+        claims_to_roles = [{
+            claim = "test"
+            roles = ["terraform"]
+        }]
+
+        redirect_urls = [ "https://example.com/redirect", "https://example.com/redirect2" ]
```
Joerger, line 19: Since `redirect_url`/`redirect_urls` is now required by the Teleport API, we should also add it to `oidc_connector.tf.example`.

P.S. I pushed that change before our recent conversation about API compatibility, sorry...

marcoandredinis: Nice catch, thank you

marcoandredinis:
@Joerger You can do it now :sweat_smile:

Joerger (approved):

r0mant (approved):
Let's bump API to 9.3.7 now that's it's also out.

PR #591, opened 2022-06-27
Promote to public ecr

logand22:


access/email/Makefile
```diff
@@ -18,6 +18,7 @@ RELEASE_MESSAGE = "Building with GOOS=$(OS) GOARCH=$(ARCH)."
 
 DOCKER_NAME = teleport-plugin-email
 DOCKER_IMAGE = 146628656107.dkr.ecr.us-west-2.amazonaws.com/gravitational/$(DOCKER_NAME):$(VERSION)
+DOCKER_IMAGE_ECR_PUBLIC = public.ecr.aws/d7a2c0j1/$(DOCKER_NAME):$(VERSION)
```
wadells, line 21: It'd be nice to loop back and fix `d7a2c0j1` once we have `gravitational`.  It would be confusing to have push to a different url than pull, even if they both work.

wadells (commented):
Would you add a note to `RELEASE.md` about how a plugins maintainer can assume the release role which can push to this repo?  ECR isn't exactly intuitive for first time users -- especially since we want folks to use their own sso creds intead of the drone service credentials.

access/email/Makefile
logand22: Yes, I'll add a followup issue for this. I plan on replacing with `gravitational` once it is available.

gozer (approved):

tcsc (approved):

PR #596, opened 2022-07-06
PagerDuty: Add config for specifying which roles should trigger incidents

kazimierzbudzyk:
Currently PagerDuty plugin will always create incidents if the source role is annotated with a service. To add more flexibility, this PR introduces config field allowing users to specify requests to what roles should trigger incident creation. Default behavior when the new config field is not set remains the same.

### DIFF
--- a/access/pagerduty/app.go
+++ b/access/pagerduty/app.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/gravitational/teleport-plugins/lib/stringset"
 	"strings"
 	"time"
 
@@ -205,6 +206,9 @@ func (a *App) onWatcherEvent(ctx context.Context, event types.Event) error {
 
 		var err error
 		switch {
+		case !a.maybeProcessRequest(ctx, req):
+			log.Info("Request skipped.")
+			return nil
 		case req.GetState().IsPending():
 			err = a.onPendingRequest(ctx, req)
 		case req.GetState().IsApproved():
@@ -233,6 +237,25 @@ func (a *App) onWatcherEvent(ctx context.Context, event types.Event) error {
 	default:
 		return trace.BadParameter("unexpected event operation %s", op)
 	}
+}
+
+// maybeProcessRequest checks request against the configured filter and decides if request should be processed.
+// When RolesToProcess field is empty (default) all requests will be processed.
+// When RolesToProcess field is configured, requests with roles specified in config will be processed.
+func (a *App) maybeProcessRequest(ctx context.Context, req types.AccessRequest) bool {
+	rolesToProcess := stringset.New(a.conf.RolesToProcess...)
+	if len(rolesToProcess) == 0 {
+		logger.Get(ctx).Debug("RolesToProcess not specified, processing all requests.")
+		return true
+	}
+	for _, r := range req.GetRoles() {
+		if rolesToProcess.Contains(r) {
+			logger.Get(ctx).Debug("Requested role found in RolesToProcess, processing.")
+			return true
+		}
+	}
+	logger.Get(ctx).Debug("Requested role not found in RolesToProcess, skipping.")
+	return false
 }
 
 func (a *App) onPendingRequest(ctx context.Context, req types.AccessRequest) error {
--- a/access/pagerduty/app.go
+++ b/access/pagerduty/app.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/gravitational/teleport-plugins/lib/stringset"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


marcoandredinis:
Hey @kazimierzbudzyk 
Thank you for your contribution

Do you think the `role_to_recipients` could work here?
This feature allows you to map the services for each role
It already exists for Slack and Email plugins

I would rather implement that logic for PagerDuty instead of creating something entirely new.

If it doesn't work, let me know
Also, can you give me some example use cases you are trying to solve?

marcoandredinis:
I think we'll eventually need a way to ignore or allow only a subset of roles
However, the way I see this being configured is slightly different from what you have
I imagine something like this:
```toml
[roles]
allowlist = [ ] # defaults to `[ "*" ]`
ignored = [ ] # defaults to `[]`
```

What do you think?
We could implement only the `roles.allowlist`, which I think is something you currently have except for naming

kazimierzbudzyk:
Tweaked as suggested.

access/pagerduty/app.go
```diff
@@ -235,6 +239,25 @@ func (a *App) onWatcherEvent(ctx context.Context, event types.Event) error {
...
+// When Approve field is configured, requests with roles specified in config will be processed.
+func (a *App) maybeProcessRequest(ctx context.Context, req types.AccessRequest) bool {
+	rolesToProcess := stringset.New(a.conf.Roles.Allowlist...)
+	if rolesToProcess.Contains("*") {
+		logger.Get(ctx).Debug("All roles accepted, processing all requests.")
+		return true
+	}
+	for _, r := range req.GetRoles() {
+		if rolesToProcess.Contains(r) {
+			logger.Get(ctx).Debug("Requested role found in RolesToProcess, processing.")
```
marcoandredinis, line 253:
```suggestion
			logger.Get(ctx).Debug("Requested role found in 'roles.allowlist', processing.")
```

access/pagerduty/app.go
```diff
@@ -235,6 +239,25 @@ func (a *App) onWatcherEvent(ctx context.Context, event types.Event) error {
 	}
 }
 
+// maybeProcessRequest checks request against the configured filter and decides if request should be processed.
+// When Approve field is unset or contains "*", all requests will be processed.
+// When Approve field is configured, requests with roles specified in config will be processed.
```
marcoandredinis, line 244: Can you please re-word this comment?
I don't think we have an `Approve` field
We are also missing the field name we use to filter out the requests (in this case: roles)

access/pagerduty/app.go
```diff
@@ -235,6 +239,25 @@ func (a *App) onWatcherEvent(ctx context.Context, event types.Event) error {
...
+		logger.Get(ctx).Debug("All roles accepted, processing all requests.")
+		return true
+	}
+	for _, r := range req.GetRoles() {
+		if rolesToProcess.Contains(r) {
+			logger.Get(ctx).Debug("Requested role found in RolesToProcess, processing.")
+			return true
+		}
+	}
+	logger.Get(ctx).Debug("Requested role not found in RolesToProcess, skipping.")
```
marcoandredinis, line 257: Same as above (those should be leftovers from the first iteration)

access/pagerduty/app.go
marcoandredinis: It should never be unset/empty according to the `CheckAndSet` method

access/pagerduty/config.go
```diff
@@ -69,6 +74,13 @@ user_email = "me@example.com" # PagerDuty bot user email (Could be admin email)
 [log]
 output = "stderr" # Logger output. Could be "stdout", "stderr" or "/var/lib/teleport/pagerduty.log"
 severity = "INFO" # Logger severity. Could be "INFO", "ERROR", "DEBUG" or "WARN".
+
+# Plugin will only process requests containing specified roles. If contains "*" (default) all requests will be processed.
+[roles]
+allowlist = [
+  "role1", 
+  "role2"
+]
```
marcoandredinis, line 83: This section is written into the configuration file when configuring the plugin
It should contain the most reasonable/friendly configuration we can think of
In this case, can you please set the default value to be `[ "*" ]`
Otherwise, users must edit this section

marcoandredinis:
@kazimierzbudzyk Can you please add tests?

We'll need to add two other things
- updating helm charts https://github.com/gravitational/teleport-plugins/tree/master/charts/access/pagerduty
- adding documentation
  - changes go here https://github.com/gravitational/teleport/blob/master/docs/pages/enterprise/workflow/ssh-approval-pagerduty.mdx
  - they appear here https://goteleport.com/docs/enterprise/workflow/ssh-approval-pagerduty/

Are you comfortable in doing those?

r0mant:
@marcoandredinis @kazimierzbudzyk Let us discuss with the product team on what we want the UX to look like here. Then we'll take over the PR and wrap it up. Thanks.

Joerger:
> As a sidenote, we also have some roles for which always want to have requests get automatically approved (based just on target role, regardless of being on call). If this is something you would consider useful for this plugin we could get rid of the custom "autoapprover" we built to solve that.

If I'm understanding correctly, your auto-approver is completely separate from pagerduty, so it wouldn't belong in this plugin. Instead our suggestion would be to run multiple plugins simultaneously, which is what it sounds like you're doing already.

As a side note to your side note, it would be an interesting idea for us to create a "plugin-manager", so that you can run multiple plugins and configure them in a single spot and run them in a single process.

marcoandredinis:
Closing this PR
Most of the Pager Duty code moved to https://github.com/gravitational/teleport/tree/master/integrations/access/pagerduty

PR #610, opened 2022-07-18
Clean-up GitHub and OIDC connector example terraform files

reedloden:
* GitHub connector -- use fake data for client ID and update org name
* OIDC connector -- use a variable for the client secret

Addresses last (?) part of TEL-Q321-09.


zmb3 (approved):
Looks good, though it reminds me that we should see if the terraform provider has been updated to accommodate the `teams_to_logins` -> `teams_to_roles` conversion.

Joerger (approved):

PR #614, opened 2022-07-20
Add Terraform to Darwin steps in CI

marcoandredinis:
Terraform tests are now passing
We also updated the Teleport version we use for integration tests

image

### DIFF
--- a/event-handler/event_handler_test.go
+++ b/event-handler/event_handler_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"bytes"
 	"encoding/json"
 	"os"
 	"os/user"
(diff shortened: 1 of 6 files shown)


event-handler/event_handler_test.go
```diff
@@ -221,13 +227,24 @@ func (s *EventHandlerSuite) TestEvents() {
...
+	cmdStderr := &bytes.Buffer{}
+
+	cmd.Stdout = cmdStdout
+	cmd.Stderr = cmdStderr
+
 	err = cmd.Start()
 	require.NoError(t, err)
 
-	_, err = stdinPipe.Write([]byte("exit\n"))
+	_, err = stdinPipe.Write([]byte("whoami\n\r"))
```
marcoandredinis, line 239: Calling `exit` immediately caused the process to fail
Maybe it was trying to setup the shell :question:

Joerger (approved):

marcoandredinis:
@r0mant Can you please take a look when you get a chance?

r0mant (approved):

PR #617, opened 2022-07-27
MS Teams access plugin

gzigzigzeo:
Needs to be merged after #589 

Closes gravitational/teleport#11818


lib/plugindata/cas.go
```diff
@@ -0,0 +1,129 @@
...
+// a flag, which indicates that input value was valid and up to date, and the resulting value
+// must be saved.
+//
+// Please note that fn might be called several times due to CAS backoff, hence, you must be careful
+// with things like I/O ops and channels.
+//
+// Method returns modified plugin data on success, nil when data was not saved, error otherwise.
+//
+// container argument must contain the reference to an empty struct implementing PluginData interface.
+func (c *CAS) Modify(ctx context.Context, resource string, cb Cb) (interface{}, error) {
```
Joerger, line 71: Can we split this into `Modify` and `Add`, where `Modify` returns a `trace.NotFound` if `oldData == nil` and `Add` returns a `trace.AlreadyExists` if `oldData != nil`? Then callers can check relevant errors instead of checking `pluginData == nil`.

I think this would be more straightforward and easy to use, which is important given this is a new library.

access/ms-teams/Makefile
```diff
@@ -0,0 +1,77 @@
+VERSION=9.3.9
+GO_VERSION=1.17.11
```
r0mant, line 2: Should we use the latest release now that Teleport 10 is out? It also uses Go 1.18.

access/ms-teams/app.go
```diff
@@ -0,0 +1,517 @@
...
+	"github.com/gravitational/teleport/api/client/proto"
+	"github.com/gravitational/teleport/api/types"
+	"github.com/gravitational/trace"
+)
+
+const (
+	// pluginName used as Teleport plugin identifier
+	pluginName = "ms-teams"
+	// minServerVersion is the minimal teleport version the plugin supports.
+	minServerVersion = "6.1.0-beta.1"
```
r0mant, line 25: Where did this version come from? Should probably set it to something that is supported. How about `8.0.0`?

access/ms-teams/app.go
```diff
@@ -0,0 +1,517 @@
...
+	log := logger.Get(ctx)
+	log.Infof("Starting Teleport MS Teams Plugin %s:%s", Version, Gitref)
+
+	err := a.init(ctx)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	a.Process = lib.NewProcess(ctx)
+	// a.mainJob = lib.NewServiceJob(a.run)
```
r0mant, line 67: Debug leftover?

access/ms-teams/app.go
```diff
@@ -0,0 +1,517 @@
...
+	teamsApp, err := a.bot.GetTeamsApp(ctx)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	if teamsApp == nil {
+		return trace.NotFound(
+			"MS Teams app not found in org app store. Please, refer to the documentation!",
+		)
+	}
```
Joerger, line 147: Check not found error
```suggestion
	if err != nil {
		return trace.Wrap(err)
	}
```

access/ms-teams/app.go
```diff
@@ -0,0 +1,517 @@
...
+
+	pong, err := a.checkTeleportVersion(ctx)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	var webProxyAddr string
+	if pong.ServerFeatures.AdvancedAccessWorkflows {
+		webProxyAddr = pong.ProxyPublicAddr
+	}
```
r0mant, line 124: This logic is not super clear - why are we setting webProxyAddr if AdvancedAccessWorkflows are enabled? What if they're not enabled? The webProxyAddr will be empty and passed to NewBot below - what will happen then?

access/ms-teams/app.go
```diff
@@ -0,0 +1,517 @@
...
+
+	if ok {
+		log.Info("Plugin is ready")
+	} else {
+		log.Error("Plugin is not ready")
+	}
+
+	a.mainJob.SetReady(ok)
+
+	<-a.watcherJob.Done()
```
r0mant, line 201: Should this select on the ctx.Done() also?

access/ms-teams/app.go
```diff
@@ -0,0 +1,517 @@
...
+
+// getMessageRecipients returns a recipients list for the access request
+func (a *App) getMessageRecipients(ctx context.Context, req types.AccessRequest) []string {
+	log := logger.Get(ctx)
+
+	// We receive a set from GetRecipientsFor but we still might end up with duplicate channel names.
+	// This can happen if this set contains the channel `C` and the email for channel `C`.
+	recipientSet := stringset.New()
+
+	validEmaislSuggReviewers := []string{}
```
r0mant, line 499:
```suggestion
	validEmailsSuggReviewers := []string{}
```

access/ms-teams/app.go
```diff
@@ -0,0 +1,517 @@
...
+			return PluginData{}, false
+		}
+		return PluginData{AccessRequestData: data}, true
+	})
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	// If we had not sent any messages before, let's create them
+	if pd != nil {
```
Joerger, line 302: Can we instead use `GetPluginData` to check if it exists already, instead of creating and updating the plugin data in two steps? I imagine this could lead to an out of sync state if the initial plugin data is created, but `postMessages` fails to update the app message or fails to update the plugin data.

access/ms-teams/app.go
```diff
@@ -0,0 +1,517 @@
...
+	for _, data := range teamsData {
+		logger.Get(ctx).WithFields(logger.Fields{
+			"id":        data.ID,
+			"timestamp": data.Timestamp,
+			"recipient": data.RecipientID,
+		}).Info("Successfully posted to MS Teams")
+	}
+	if err != nil {
+		logger.Get(ctx).WithError(err).Error("Failed to post one or more messages to MS Teams")
+	}
```
Joerger, line 371: nit: check the error in one go right after the call
```suggestion
	teamsData, err := a.bot.PostMessages(ctx, recipients, id, data)
	if err != nil {
		len(teamsData) == 0 {
			return trace.Wrap(err)
		}
		logger.Get(ctx).WithError(err).Error("Failed to post one or more messages to MS Teams")
	}
	for _, data := range teamsData {
		logger.Get(ctx).WithFields(logger.Fields{
			"id":        data.ID,
			"timestamp": data.Timestamp,
			"recipient": data.RecipientID,
		}).Info("Successfully posted to MS Teams")
	}
```

access/ms-teams/config.go
```diff
@@ -0,0 +1,67 @@
...
+// CheckAndSetDefaults checks the config struct for any logical errors, and sets default values
+// if some values are missing.
+// If critical values are missing and we can't set defaults for them — this will return an error.
+func (c *Config) CheckAndSetDefaults() error {
+	if err := c.Teleport.CheckAndSetDefaults(); err != nil {
+		return trace.Wrap(err)
+	}
+
+	if c.MSAPI.Region == "" {
+		c.MSAPI.Region = "uk"
```
r0mant, line 51: Nit: Should probably default to US (unless I'm missing what "uk" means).

Joerger (commented):
Mostly error handling and library usability comments, but other than that LGTM.

access/ms-teams/main.go
```diff
@@ -0,0 +1,105 @@
...
+var (
+	appName                 = "teleport-ms-teams"
+	gracefulShutdownTimeout = 15 * time.Second
+)
+
+func main() {
+	logger.Init()
+	app := kingpin.New(appName, "Teleport MS Teams plugin")
+
+	app.Command("version", "Prints teleport-slack version and exits")
```
r0mant, line 23: teleport-slack -> teleport-ms-teams?

access/ms-teams/main.go
```diff
@@ -0,0 +1,105 @@
...
+
+	targetDir := configureCmd.Arg("dir", "Path to target directory").Required().String()
+	appID := configureCmd.Flag("appID", "MS App ID").Required().String()
+	appSecret := configureCmd.Flag("appSecret", "MS App Secret").Required().String()
+	tenantID := configureCmd.Flag("tenantID", "MS App Tenant ID").Required().String()
+
+	validateCmd := app.Command("validate", "Validate bot installation")
+	validateConfigPath := validateCmd.Flag("config", "TOML config file path").
+		Short('c').
+		Default("/etc/teleport-slack.toml").
```
r0mant, line 34: teleport-ms-teams.toml?

access/ms-teams/main.go
```diff
@@ -0,0 +1,105 @@
...
+		String()
+
+	validateUserID := validateCmd.Arg("userID", "Your User ID").Required().String()
+
+	startCmd := app.Command("start", "Starts Teleport MS Teams plugin")
+	startConfigPath := startCmd.Flag("config", "TOML config file path").
+		Short('c').
+		Default("/etc/teleport-slack.toml").
```
r0mant, line 42: teleport-ms-teams.toml

access/ms-teams/msapi/bot_framework_client.go
```diff
@@ -0,0 +1,135 @@
...
+	"context"
+	"net/http"
+
+	"github.com/gravitational/trace"
+)
+
+const (
+	botFrameworkAuthScope     = "https://api.botframework.com/.default"
+	botFrameworkBaseURL       = "https://smba.trafficmanager.net"
+	botFrameworkDefaultRegion = "uk"
```
r0mant, line 13: Set it to "us" by default?

access/ms-teams/msapi/bot_framework_client.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// PostAdaptiveCardActivity sends an activity to the chat, content is AdaptiveCard
+func (c *BotFrameworkClient) PostAdaptiveCardActivity(ctx context.Context, botID, chatID, card, updateID string) (string, error) {
+	body := `
+		{
+			"type": "message",
+			"from": {
+				"id": "29:` + botID + `",
+				"name": "TeleBot"
```
r0mant, line 100: What's "29"? And where does this "name" show up?

access/ms-teams/msapi/graph_client.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	var users []User
+	err = json.NewDecoder(bytes.NewReader(g.Value)).Decode(&users)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	if len(users) == 0 {
+		return nil, nil
```
r0mant, line 223: Should this return trace.NotFound?

access/ms-teams/msapi/graph_client.go
```diff
@@ -0,0 +1,265 @@
...
```
r0mant, line 257: Should this return trace.NotFound?

access/ms-teams/msapi/token.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// Bearer returns current token value and refreshes it if token is expired.
+//
+// MS Graph API issues no refresh_token for client_credentials grant type. There also is no
+// extended validity window for this grant type. Hence, there is no way of preemprive refresh
+// the current token. We have to lock all requests and wait until the new token arrives.
+func (c *tokenWithTTL) Bearer(ctx context.Context, config Config) (string, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
r0mant, line 46: I would try to avoid doing any I/O (incl. network calls as getToken below does) under lock. Any way to refactor this do we're not making an API request while holding a lock?

lib/plugindata/cas.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// EncodeFn represents a function type of plugin data serializer
+type EncodeFn = func(interface{}) map[string]string
+
+// DecodeFn represents a function type of plugin data deserializer
+type DecodeFn = func(map[string]string) interface{}
+
+// Cb represents a type for Plugin Data modification fn
+type Cb = func(interface{}) (interface{}, bool)
```
r0mant, line 33: Can we rename this to something more descriptive? Does this stand for "callback"?

access/ms-teams/_tpl/manifest.json
```diff
@@ -0,0 +1,45 @@
+{
+    "$schema": "https://developer.microsoft.com/en-us/json-schemas/teams/v1.11/MicrosoftTeams.schema.json",
+    "manifestVersion": "1.11",
```
marcoandredinis, line 3: `1.14` is the latest version
Is there a reason we are using 1.11?

access/ms-teams/_tpl/manifest.json
```diff
@@ -0,0 +1,45 @@
...
+        "outline": "outline.png"
+    },
+    "name": {
+        "short": "TeleBot",
+        "full": "TeleBot"
+    },
+    "description": {
+        "short": "Teleport bot",
+        "full": "Teleport bot sends AccessRequests to team users"
+    },
```
marcoandredinis, line 24: @r0mant We might want to double check these descriptions with Product

access/ms-teams/_tpl/teleport-ms-teams.toml
```diff
@@ -0,0 +1,14 @@
+preload = false
+
+[teleport]
+addr = "localhost:3025"
+identity = "identity"
```
marcoandredinis, line 5: Can we add some comments explaining how to obtain the `identity` file?
Similar to what we have in Slack

go.mod
```diff
@@ -35,6 +35,11 @@ require (
 	k8s.io/apimachinery v0.24.1
 )
 
+require (
```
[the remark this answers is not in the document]
marcoandredinis: This should go away when we rebase with the current `master` which brings go 1.18

lib/plugindata/access_request.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+// AccessRequestData represents generic plugin data required for access request processing
+type AccessRequestData struct {
+	User             string
+	Roles            []string
+	RequestReason    string
+	ReviewsCount     int
+	ResolutionTag    ResolutionTag
+	ResolutionReason string
```
marcoandredinis, line 25: :nail_care: 
The other plugins have those two fields in a single struct - `Resolution`
https://github.com/gravitational/teleport-plugins/blob/7ea82ce5d94119aeb5f23a40a8fe88edbed3d326/access/slack/plugindata.go#L14-L17
It would be good if we did that here as well, so that latter on, we can remove each individual implementation to this `/lib/plugindata`

access/ms-teams/configure.go
```diff
@@ -0,0 +1,165 @@
...
+	"path"
+
+	"github.com/google/uuid"
+	"github.com/gravitational/teleport-plugins/lib"
+	"github.com/gravitational/trace"
+)
+
+const (
+	// TODO: Replace with live URL
+	guideURL = "http://example.com"
```
marcoandredinis, line 19: :nail_care: 
We can use `https://goteleport.com/docs/enterprise/workflow/` for now, but let's keep the TODO

lib/plugindata/cas.go
```diff
@@ -0,0 +1,129 @@
...
+
+// CAS represents modifier struct
+type CAS struct {
```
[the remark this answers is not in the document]
gzigzigzeo: CAS stands for compare-and-swap. Renamed it to the CompareAndSwap, as sync in go/sync package.

lib/plugindata/cas.go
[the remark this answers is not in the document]
gzigzigzeo: Renamed to CallbackFn

lib/plugindata/cas.go
```diff
@@ -0,0 +1,129 @@
...
+type Client interface {
+	GetPluginData(context.Context, types.PluginDataFilter) ([]types.PluginData, error)
+	UpdatePluginData(context.Context, types.PluginDataUpdateParams) error
+}
+
+// EncodeFn represents a function type of plugin data serializer
+type EncodeFn = func(interface{}) map[string]string
+
+// DecodeFn represents a function type of plugin data deserializer
+type DecodeFn = func(map[string]string) interface{}
```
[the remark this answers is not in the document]
marcoandredinis: (`master` is already using 1.18)

lib/plugindata/cas.go
gzigzigzeo: Compare-and-swap supposed to be a single atomic instruction. I possibly get your idea wrong, but splitting this into two methods would break the atomicity. Just in case, this method is not new. It is used in every plugin like this, I just tried to DRY it into a separate module.

access/ms-teams/msapi/bot_framework_client.go
```diff
@@ -0,0 +1,135 @@
...
+	if region == "" {
+		region = botFrameworkDefaultRegion
+	}
+
+	baseURL := config.url.botFrameworkBaseURL
+	if baseURL == "" {
+		baseURL = botFrameworkBaseURL
+	}
+
+	baseURL = baseURL + "/" + region + "/" + botFrameworkVersion
```
marcoandredinis, line 63: You can use the `CheckAndSetDefaults` pattern

access/ms-teams/msapi/bot_framework_client.go
```diff
@@ -0,0 +1,135 @@
...
+		Client: Client{
+			token:   tokenWithTTL{scope: botFrameworkAuthScope, baseURL: config.url.tokenBaseURL},
+			baseURL: baseURL,
+			config:  config,
+		},
+	}
+}
+
+// GetChatMembers returns members of a chat
+func (c *BotFrameworkClient) GetChatMembers(ctx context.Context, chatID string) ([]ChatMember, error) {
```
marcoandredinis, line 75: Is this method being used?

access/ms-teams/app.go
gzigzigzeo: This would create a race condition if another instance of the plugin writes its data between get and modify calls.

access/ms-teams/msapi/bot_framework_client.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+
+	return m, nil
+}
+
+// PostAdaptiveCardActivity sends an activity to the chat, content is AdaptiveCard
+func (c *BotFrameworkClient) PostAdaptiveCardActivity(ctx context.Context, botID, chatID, card, updateID string) (string, error) {
+	body := `
```
marcoandredinis, line 95: Can we convert this into a struct and call jsonMarshal?
It's an extra call, but I would rather let go build the json automatically

access/ms-teams/msapi/bot_framework_client.go
```diff
@@ -0,0 +1,135 @@
...
+	meth := http.MethodPost
+	status := http.StatusCreated
+	if updateID != "" {
+		meth = http.MethodPut
+		status = http.StatusOK
+	}
+
+	request := request{
+		Method:      meth,
+		Path:        "conversations/" + chatID + "/activities/" + updateID,
```
marcoandredinis, line 122: Is this path valid if the `updateID` is `""` ?

I think we should set the `Path` the same way we are doing for the Method and SuccessCode

access/ms-teams/Makefile
gzigzigzeo: With pleasure :-)

access/ms-teams/app.go
gzigzigzeo: Frankly, I do not know. Looks like that minimal version was copy pasted all over the plugins from the beginning of times. I've replaced it here, but I guess we need to create an issue to replace it everywhere else.

access/ms-teams/msg_slice_test.go
```diff
@@ -0,0 +1,25 @@
+package main
```
marcoandredinis, line 1: This file ends with `_test.go` but doesn't actually contain any tests

access/ms-teams/app.go
gzigzigzeo: `webProxyAddr` is used to show a fancy "Open" button, which launches Teleport Web UI and opens the access request in question. As far as I understand, that feature exists in Enterprise version only - that's why we check `AdvancedAccessWorkflows`. Yes, empty value indicates to app that it should not render those links. It works like this for all other plugins, should we change it?

access/ms-teams/app.go
gzigzigzeo: As far as I remember, it is triggered when the parent context is done. But, just in case, I've added it.

access/ms-teams/config.go
gzigzigzeo: That's not clear here. MS does not have proper documentation on the region codes and bot framework REST API usage. I am not even sure that this region setting affects the real region. I've changed it to `emea` as it seems the default region which is used by the .NET MS SDKs themselves.

access/ms-teams/msapi/bot_framework_client.go
gzigzigzeo: Unfortinately, there's no "us", and I have failed to find the region codes list. Moreover, according to my findings in the github issues, etc. bot framework is not working with all regions. So, I set this to `emea`, which seems to be the ms default region.

access/ms-teams/msapi/bot_framework_client.go
gzigzigzeo: 29 is an internal API designator used to determine a resource type (bot in our case). Name does not show up anywhere, it is used as a placeholder here. The problem is the following. Normally, MS bot is supposed to receive events from the users. When a user (or an admin) installs a bot for a user, the "StartConversation" event is generated. This event contains "29:<id>" along with some other information required to continue the conversation from bot's side. Bot is supposed to process this event and save all the information in the persistent storage for later usage.That's the normal way of how MS bots work. In our case there's no answering part, so I have to improvise. This approach is confirmed by an article from one of MS Teams engineers, where he proposes to do the same in case an original event got lost or a bot is totally passive and can't process any evetns.

access/ms-teams/msapi/token.go
gzigzigzeo: MS Graph API issues no `refresh_token` for `client_credentials` grant type. There also is no extended validity window for this grant type. Hence, there is no way of preemptive refresh the current token. We have to lock all requests and wait until the new token arrives.

I guess, that's my mistake, there's not much harm in doing multiple concurrent token refresh requests. Refactored.

access/ms-teams/_tpl/manifest.json
gzigzigzeo: That was auto generated by the ms tool. Are you sure by 100% that it would work?

lib/plugindata/access_request.go
gzigzigzeo: My guess was that the separate Resolution struct her is redundant since it is almost never used separately. `/lib/plugindata` raises lots of questions, so, let me give one more pass to it and then we decide.

access/ms-teams/msapi/bot_framework_client.go
gzigzigzeo: I guess, `CheckAndSetDefaults` here is applicable to the region only. Will move it to the `msapi.Config#CheckAndSetDefaults`.

access/ms-teams/msapi/bot_framework_client.go
gzigzigzeo: Nope, it's the debug artifact, thanks.

access/ms-teams/msg_slice_test.go
gzigzigzeo: Yes, but it contains the code essential for tests to work. `_test.go` indicates that this file belongs to the test domain.

access/ms-teams/msapi/bot_framework_client.go
gzigzigzeo: Yes, it is valid.

access/ms-teams/Makefile
```diff
@@ -0,0 +1,77 @@
+VERSION=10.0.2
+GO_VERSION=1.18.4
```
marcoandredinis, line 2: Don't forget to update `go.mod` as well, or just rebase from `master`

marcoandredinis:
We need to update the `Makefile` at root as well

access/ms-teams/configure.go
```diff
@@ -0,0 +1,164 @@
...
+		TeamsAppID: uuid.New().String(),
+	}
+
+	lib.PrintVersion(appName, Version, Gitref)
+	fmt.Println()
+
+	_, err := os.Stat(targetDir)
+	if !os.IsNotExist(err) {
+		return trace.Wrap(err)
+	}
```
marcoandredinis, line 60: Re-using an already existing directory didn't throw an error
It only printed the version and an empty line, nothing else

access/ms-teams/Makefile
gzigzigzeo: Done

access/ms-teams/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+```sh
+$ ./teleport-ms-teams configure ~/ms_teams --appID <Microsoft App ID value> --tenantID <App Tenant ID value> --appSecret <password value saved on step #4>
+```
+
+`~ms_teams` must be a path to non existent folder. It will be created. All required files will be generated.
+
+8. There will be `app.zip` in the target directory. This archive contains MS Teams app. If you want to change the application name, URLs and other settings, please update `manifest.json` and recreate `app.zip` from `color.png`, `outline.png` and `manifest.json`.
+
+9. Upload the app to MS Teams organization store. Open "Apps", then "Manage your apps", click "Upload an app", choose "Upload an app to your org's app catalog", choose `app.zip`
```
marcoandredinis, line 118: I think it's worth mentioning that Microsoft Teams Free does not support apps at all.

I created an app and was trying to install it to my Teams org.
There's not even an `Apps` menu when using the Free version

We can add a note at the beginning.

access/ms-teams/configure.go
gzigzigzeo: Fixed

access/ms-teams/README.md
gzigzigzeo: Was not aware of that. I have free account and it seems working for me.

gzigzigzeo: I guess, you need kind of Azure subscription.

marcoandredinis: Possibly :thinking: 
I created an account just now, not even associated with my work email
I'll try to get an azure subscription
image

access/ms-teams/validate.go
```diff
@@ -0,0 +1,136 @@
...
+
+	b, err := NewBot(c.MSAPI, "local", "")
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	teamApp, err := b.GetTeamsApp(context.Background())
+	if trace.IsNotFound(err) {
+		fmt.Printf("Application %v not found in the org app store. Please, ensure that you have the application uploaded and installed for your team.", c.MSAPI.TeamsAppID)
+		return nil
```
marcoandredinis, line 38: I think we should return an error here :thinking: 
Otherwise it's _validated_

access/ms-teams/bot.go
```diff
@@ -0,0 +1,237 @@
...
+
+	return nil
+}
+
+// FetchUser fetches app id for user, installs app for a user if missing, fetches chat id and saves
+// everything to cache. This method is used for priming the cache. Returns trace.NotFound if a
+// user was not found.
+func (b Bot) FetchUser(ctx context.Context, userIDOrEmail string) (*UserData, error) {
+	if b.teamsApp == nil {
+		return nil, trace.Errorf("Bot is not configured, run GetTeamsApp first")
```
marcoandredinis, line 111: Can we just call `GetTeamsApp` here?

access/ms-teams/bot.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	b.mu.RLock()
+	d, ok := b.users[userIDOrEmail]
```
marcoandredinis, line 115: Do we need to restart the plugin if we want a newly created user to receive notifications?

access/ms-teams/bot.go
```diff
@@ -0,0 +1,237 @@
...
+		return trace.Wrap(err)
+	}
+
+	return nil
+}
+
+// FetchUser fetches app id for user, installs app for a user if missing, fetches chat id and saves
+// everything to cache. This method is used for priming the cache. Returns trace.NotFound if a
+// user was not found.
+func (b Bot) FetchUser(ctx context.Context, userIDOrEmail string) (*UserData, error) {
```
marcoandredinis, line 109: Can we rename this method to something that indicates there's side effect when calling this method?

Something like `InstallAppForUser` or something similar?

access/ms-teams/configure.go
```diff
@@ -0,0 +1,164 @@
...
+	renderTemplateTo(manifestTpl, p, path.Join(targetDir, "manifest.json"))
+
+	printStep(&step, "Generated configuration files")
+
+	a, err := assets.ReadDir("_tpl")
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	for _, a := range a {
```
marcoandredinis, line 79: We have two variables named `a`
Can we please rename one of them, if possible with something more explicit

access/ms-teams/configure.go
```diff
@@ -0,0 +1,164 @@
...
+	w, err := os.Create(targetPath)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	defer w.Close()
+
+	err = tpl.ExecuteTemplate(w, "template", payload)
+	if err != nil {
+		return trace.Wrap(err)
+	}
```
marcoandredinis, line 161:
```suggestion
	if err := tpl.ExecuteTemplate(w, "template", payload); err != nil {
		return trace.Wrap(err)
	}
```

access/ms-teams/main.go
```diff
@@ -0,0 +1,105 @@
...
+		Short('c').
+		Default("/etc/teleport-ms-teams.toml").
+		String()
+
+	validateUserID := validateCmd.Arg("userID", "Your User ID").Required().String()
+
+	startCmd := app.Command("start", "Starts Teleport MS Teams plugin")
+	startConfigPath := startCmd.Flag("config", "TOML config file path").
+		Short('c').
+		Default("/etc/teleport-ms-teams.toml").
```
marcoandredinis, line 42: We should extract this to a const, it's used in multiple places

access/ms-teams/msapi/client.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		defer resp.Body.Close()
+
+		b, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return trace.Wrap(err)
+		}
+
+		if resp.StatusCode > 499 {
```
marcoandredinis, line 102:
```suggestion
		if resp.StatusCode > 499 || resp.StatusCode == http.StatusTooManyRequests {
```

access/ms-teams/msapi/token.go
```diff
@@ -0,0 +1,132 @@
...
+			continue
+		}
+
+		defer resp.Body.Close()
+		b, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return t, trace.Wrap(err)
+		}
+
+		if resp.StatusCode != http.StatusOK {
```
marcoandredinis, line 121: This never evaluates to true, does it?
We are `continue`ing at line 112 if `resp.StatusCode != http.StatusOK`

access/ms-teams/msapi/graph_client.go
```diff
@@ -0,0 +1,265 @@
...
+		Err:         &graphError{},
+		SuccessCode: http.StatusCreated,
+	}
+
+	err := c.request(ctx, request)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	return nil
```
marcoandredinis, line 177:
```suggestion
	return trace.Wrap(c.request(ctx, request))
```

access/ms-teams/msapi/bot_framework_client.go
```diff
@@ -0,0 +1,124 @@
...
+		},
+	}
+}
+
+// PostAdaptiveCardActivity sends an activity to the chat, content is AdaptiveCard
+func (c *BotFrameworkClient) PostAdaptiveCardActivity(ctx context.Context, botID, chatID, card, updateID string) (string, error) {
+	m := postMessagePayload{
+		Type: "message",
+		From: postMessagePayloadFrom{
+			ID:   "29:` + botID + `",
```
marcoandredinis, line 86: We should move the `29` to a const and add an explanation (the link to the blog/article you talked about should do)

access/ms-teams/msapi/bot_framework_client.go
```diff
@@ -0,0 +1,124 @@
...
+			Name: "TeleBot",
+		},
+		Attachments: []postMessagePayloadAttachment{{
+			ContentType: "application/vnd.microsoft.card.adaptive",
+			Content:     []byte(card),
+		}},
+	}
+
+	body, err := json.MarshalIndent(&m, "", "    ")
```
marcoandredinis, line 95: Nothing against indenting the json, but is it required?

lib/plugindata/cas.go
Joerger: I agree that we should use compare-and-swap, but we can still move the functionality from the callback to the function by splitting it into two functions. Here's what I mean:

```
func (c *CAS) Modify(ctx context.Context, resource string, modify func(interface{}) interface{}) (interface{}, error) {
	backoff := backoff.NewDecorr(c.backoffBase, c.backoffMax, clockwork.NewRealClock())
	for {
		oldData, err := c.getPluginData(ctx, resource)
		if err != nil {
			return nil, trace.Wrap(err)
		}

		newData := modify(oldData)

		err = trace.Wrap(c.updatePluginData(ctx, resource, newData, oldData))
		if err == nil {
			return newData, nil
		}
		if !trace.IsCompareFailed(err) {
			return nil, trace.Wrap(err)
		}
		if err := backoff.Do(ctx); err != nil {
			return nil, trace.Wrap(err)
		}
	}
}

func (c *CAS) Create(ctx context.Context, resource string, pd interface{}) (interface{}, error) {
	// We don't need to do backoff, because if compare fails
	// that means the plugin data already exists. Abort creation.
	err := trace.Wrap(c.updatePluginData(ctx, resource, pd, nil))
	if err != nil {
		return nil, trace.Wrap(err)
	}
	return pd, nil
}
```

Now we can simplify our calls to Modify like this:

```
// postMessages posts initial Teams messages
func (a *App) postMessages(ctx context.Context, recipients []string, id string, data pd.AccessRequestData) error { 
	...
	// Let's update sent messages data
	_, err = a.pd.Modify(ctx, id, func(existing interface{}) (interface{}) {
		pluginData, ok := existing.(PluginData)
		if !ok {
			// maybe return error?
			return nil
		}
		pluginData.TeamsData = teamsData
		return pluginData
	})
	return trace.Wrap(err)
}
```

And we can simplify Creation like this:

```
func (a *App) onPendingRequest(ctx context.Context, req types.AccessRequest) error {
	...
	pd, err := a.pd.Modify(ctx, id, PluginData{AccessRequestData: data})
	if err != nil {
		return trace.Wrap(err)
	}
```

access/ms-teams/plugindata.go
```diff
@@ -0,0 +1,61 @@
...
+	data := PluginData{}
+
+	data.AccessRequestData = plugindata.DecodeAccessRequestData(dataMap)
+
+	if channelID, timestamp := dataMap["channel_id"], dataMap["timestamp"]; channelID != "" && timestamp != "" {
+		data.TeamsData = append(data.TeamsData, TeamsMessage{ID: channelID, Timestamp: timestamp})
+	}
+	if str := dataMap["messages"]; str != "" {
+		for _, encodedMsg := range strings.Split(str, ",") {
+			parts := strings.Split(encodedMsg, "/")
```
marcoandredinis, line 34: Can we assume that ID, Timestamp and RecipientID won't have neither a `,` or a `/`?

access/ms-teams/app.go
```diff
@@ -0,0 +1,521 @@
...
+
+// init initializes the application
+func (a *App) init(ctx context.Context) error {
+	bk := grpcbackoff.DefaultConfig
+	bk.MaxDelay = grpcBackoffMaxDelay
+
+	apiClient, err := client.New(ctx, client.Config{
+		Addrs:       a.conf.Teleport.GetAddrs(),
+		Credentials: a.conf.Teleport.Credentials(),
+		DialOpts: []grpc.DialOption{
```
marcoandredinis, line 98: Let's add another DialOption here
```go
grpc.WithReturnConnectionError(),
```
https://github.com/gravitational/teleport-plugins/pull/601

access/ms-teams/validate.go
gzigzigzeo: We do not want an error output and backtrace in this case. validate.go represents the console command which sends a fake message, it returns only fatal errors (networking, etc). Otherwise, it nicely shows the result to a user.

marcoandredinis: Got it :+1: 

I would expect the `validate` method to return an error and we would then print the message on the caller
It's fine this way

marcoandredinis:
We should run the linter, I think it has some offenses

access/ms-teams/bot.go
gzigzigzeo: This method does the following: 1) Tries to fetch user 2) Installs an app if missing 3) Gets the chat for a user. I do not think that `InstallAppForUser` fits. Do you probably have any ideas?

access/ms-teams/configure.go
gzigzigzeo: Hmm, I use the multiline style almost everywhere, could you please elaborate a bit, why to use single line here?

access/ms-teams/bot.go
marcoandredinis: `EnsureAppForUser`
Out of ideas today, sorry :joy: 

My point is that `Fetch` does not seem to indicate that we are actually installing the app
Maybe we can extract this to smaller functions

access/ms-teams/msapi/client.go
gzigzigzeo: Thanks!

access/ms-teams/configure.go
marcoandredinis: Sorry, it's ok
Ignore the other comments regarding this type of change

I usually prefer using the single line, but no need to change to it

access/ms-teams/msapi/token.go
gzigzigzeo: Right. Changed the order.

access/ms-teams/msapi/bot_framework_client.go
gzigzigzeo: It facilitates the debugging

access/ms-teams/plugindata.go
gzigzigzeo: Yes. ID and RecipientID are both UUIDs, Timestamp is RFC822 if I am not mistaken. But you're right, this possibly needs changing. I've just used the same format we used in other plugins.

gzigzigzeo:
Okay, will do.

access/ms-teams/bot.go
gzigzigzeo: There are a smaller functions already, this method composes them into the strategy :-) Let's just leave it as is until I fix all other issues :-)

access/ms-teams/README.md
gzigzigzeo: I've added the notice, anyway.

access/ms-teams/msapi/bot_framework_client.go
gzigzigzeo: Okay, just in case here's the article: https://hackandchat.com/teams-proactive-messaging/

access/ms-teams/app.go
Joerger: We should continue to use `CompareAndSwap` to avoid such race conditions, unless that's not the kind of race condition you're referring to?

The problem I'm pointing out is that when we create the initial `PluginData{AccessRequestData: data}` object, it is essentially useless until we add the `TeamsData` in the following `Modify` call. So if `a.bot.PostMessages` fails, this plugin data will remain in this useless state, and as I understand any future calls to update it would log `"Failed to post reply: plugin data is missing"`. If we somehow had another call to `onPendingRequest`, this would also fail after seeing the plugin data already exists, even though the messages were never actually posted.

Instead, after we first `GetPluginData` just to see if it already exists. If it does, we can do `a.bot.PostMessages`. If that succeeds, we can finally do `a.pd.Modify` (or `a.pd.Create` from my other comment`) to create the full plugin data `PluginData{AccessRequestData: data, TeamsData: teamsData}`. We can always be sure that if the plugin data exists, the messages have been posted. 

This still has the possible de-synced state of the messages being posted but the plugin data failing to be created, but I'm not sure there's a simple way to circumvent this.

lib/plugindata/cas.go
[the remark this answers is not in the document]
Joerger: Thanks, `ModifyFn` would also be a fitting name if you'd like.

lib/plugindata/cas.go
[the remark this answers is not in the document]
Joerger: Sorry, I had meant to delete this comment once I realized this during review. Generics would be a great upgrade though, looking forward to seeing that.

access/ms-teams/app.go
gzigzigzeo: We can not get TeamsData unless we have messages actually sent. PluginData without TeamsData marks that an instance of the plugin has acquired a lock on this AccessRequest and is responsible for sending the initial messages. Yes, the situation when postMessages fails is possible. However, if we implement it in a way you explained, we could fall in the situation when two instances of the plugin determine that there is no plugin data and send duplicate messages.

Joerger: Ah ok that makes sense, thanks for explaining

lib/plugindata/cas.go
gzigzigzeo: Ah, now I see. I've already tried to achieve the same goal with `mode` argument of the method. There are actually three situations: "Create", "Update" and "Upsert". 

Let me finish my refactoring, I get back to you and we discuss an updated version.

gzigzigzeo:
@Joerger @marcoandredinis Please, check the updated version of PluginData. Implemented generics and `Create`/`Update` instead of `Modify`. Refactored app.go a bit as well.

access/ms-teams/app.go
```diff
@@ -0,0 +1,479 @@
...
+		return trace.Wrap(err)
+	}
+
+	return nil
+}
+
+// onResolvedRequest is called when a request is resolved
+func (a *App) onResolvedRequest(ctx context.Context, req types.AccessRequest) error {
+	//	replyErr := a.postReviews(ctx, req.GetName(), req.GetReviews())
+
```
marcoandredinis, line 339:
```suggestion

```

access/ms-teams/ms_teams_test.go
```diff
@@ -0,0 +1,677 @@
...
+	reviewer := s.mockAPI.StoreUser(msapi.User{Mail: s.userNames.reviewer1})
+
+	s.startApp()
+
+	req := s.createAccessRequest([]msapi.User{reviewer})
+	msg, err := s.mockAPI.CheckNewMessage(s.Context())
+	require.NoError(t, err)
+	require.Equal(t, reviewer.ID, msg.RecipientID)
+
+	// max size of request was decreased here: https://github.com/gravitational/teleport/pull/13298
```
marcoandredinis, line 353:
```suggestion
```

access/ms-teams/ms_teams_test.go
```diff
@@ -0,0 +1,677 @@
...
+		Reason:        "okay",
+	})
+	require.NoError(t, err)
+
+	reply, err := s.mockAPI.CheckMessageUpdate(s.Context())
+	require.NoError(t, err)
+
+	require.Equal(t, msg.RecipientID, reply.RecipientID)
+	require.Equal(t, msg.ID, reply.ID)
+	require.Equal(t, gjson.Get(reply.Body, "attachments.0.content.body.4.facts.0.value").String(), "✅")
```
marcoandredinis, line 453: This paths look a bit cryptic
I think we should move them to consts with a clarifying name
What do you think?

access/ms-teams/_tpl/manifest.json
marcoandredinis: https://docs.microsoft.com/en-us/microsoftteams/platform/resources/schema/manifest-schema#manifestversion
Feel free to leave as 1.11 or update to 1.12

access/ms-teams/msapi/bot_framework_client.go
marcoandredinis: I think we introduced a bug here
This will compile to
```
29:` + botID + `
```

Instead of `29:abcd`

marcoandredinis:
PR looks very good, that's an awesome job you did here :clap: 
I still want to do a last pass, as we have a lot of changes
I'm still trying to get access to a Teams workplace to try out the plugin

access/ms-teams/ms_teams_test.go
gzigzigzeo: XSLT of the new era :-) I agree, that would increase readability.

lib/plugindata/cas.go
```diff
@@ -0,0 +1,166 @@
...
+		if !trace.IsCompareFailed(err) {
+			return emptyData, trace.Wrap(err)
+		}
+		if err := backoff.Do(ctx); err != nil {
+			return emptyData, trace.Wrap(err)
+		}
+	}
+}
+
+// TODO: Implement Upsert method
```
Joerger, line 134:
```suggestion
// Upsert tries to perform compare-and-swap update of a plugin data.
func (c *CompareAndSwap[T]) Upsert(ctx context.Context, resource string, upsertData T) (T, error) {
	emptyData := *new(T)

	backoff := backoff.NewDecorr(c.backoffBase, c.backoffMax, clockwork.NewRealClock())
	for {
		existingData, err := c.getPluginData(ctx, resource)
		if trace.IsNotFound(err) {
			existingData := emptyData
		} else if err != nil {
			return emptyData, trace.Wrap(err)
		}

		err = c.updatePluginData(ctx, resource, upsertData, existingData)
		if err == nil {
			return upsertData, nil
		} else if !trace.IsCompareFailed(err) {
			return emptyData, trace.Wrap(err)
		} else if err := backoff.Do(ctx); err != nil {
			return emptyData, trace.Wrap(err)
		}
	}
}```

File: access/ms-teams/app.go [hunk 55]
```diff
@@ -0,0 +1,477 @@
...
+	return trace.Wrap(err)
+}
+
+// postReviews updates a message with reviews
+func (a *App) postReviews(ctx context.Context, id string, reviews []types.AccessReview) error {
+	pluginData, err := a.pd.Update(ctx, id, func(existing PluginData) (PluginData, error) {
+		teamsData := existing.TeamsData
+		if len(teamsData) == 0 {
+			return PluginData{}, trace.CompareFailed("existing teamsData is empty")
```
Joerger, line 394: nit: clarifying comment
```suggestion
			// wait for the plugin data to be updated with TeamsData
			return PluginData{}, trace.CompareFailed("existing teamsData is empty")
```

access/ms-teams/app.go
```diff
@@ -0,0 +1,477 @@
...
+		}
+
+		count := len(reviews)
+		oldCount := existing.ReviewsCount
+		if oldCount >= count {
+			return PluginData{}, trace.CompareFailed("reviews are sent already")
+		}
```
Joerger, line 401: If the reviews are already sent, won't the backoff attempts to CompareAndSwap also fail? Should we just return an error other than CompareFailed to abort the update?
```suggestion
		if oldCount >= count 
			return PluginData{}, trace.AlreadyExists("reviews are sent already")
		}
```

Joerger (commented):
Great use of generics!

lib/plugindata/cas.go
gzigzigzeo: I guess, I can close this conversation

access/ms-teams/bot.go
gzigzigzeo: We could, yes, but the idea was to make user explicitly call `GetTeamsApp` and quit if it fails. It makes no sense to proceed further otherwise.

access/ms-teams/msapi/bot_framework_client.go
gzigzigzeo: Ouch, yes, surprisingly this works with a bug. My guess is that the API does not care.

gzigzigzeo: Fixed.

access/ms-teams/app.go
```diff
@@ -0,0 +1,477 @@
...
+		User:          req.GetUser(),
+		Roles:         req.GetRoles(),
+		RequestReason: req.GetRequestReason(),
+	}
+
+	// Let's try to create PluginData. This equals to locking AccessRequest to this
+	// instance of a plugin.
+	_, err := a.pd.Create(ctx, id, func() (PluginData, error) {
+		return PluginData{AccessRequestData: data}, nil
+	})
```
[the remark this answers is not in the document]
gzigzigzeo: I agree, done.

lib/plugindata/cas.go
```diff
@@ -0,0 +1,166 @@
...
+		return emptyData, trace.Wrap(err, "callback failed to update PluginData")
+	}
+
+	err = c.updatePluginData(ctx, resource, modifiedData, emptyData)
+	if err == nil {
+		return modifiedData, nil
+	}
+
+	return emptyData, trace.Wrap(err)
+}
```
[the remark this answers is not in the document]
gzigzigzeo: Done

lib/plugindata/cas.go
```diff
@@ -0,0 +1,166 @@
...
+	backoff := backoff.NewDecorr(c.backoffBase, c.backoffMax, clockwork.NewRealClock())
+	for {
+		oldData, err := c.getPluginData(ctx, resource)
+		if err != nil {
+			return emptyData, trace.Wrap(err)
+		}
+
+		if oldData == nil {
+			return emptyData, trace.NotFound("Plugin data does not exist")
+		}
```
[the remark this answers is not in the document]
gzigzigzeo: Done

lib/plugindata/cas.go
```diff
@@ -0,0 +1,166 @@
...
+
+// Update tries to perform compare-and-swap update of a plugin data assuming that it exist
+//
+// fn callback function receives current plugin data value and returns modified value and
+// error. If fn has failed to update data due to a diff, it must return CompareFailed error.
+func (c *CompareAndSwap[T]) Update(
+	ctx context.Context,
+	resource string,
+	cb func(T) (T, error),
+) (T, error) {
```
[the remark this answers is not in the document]
gzigzigzeo: Done

access/ms-teams/app.go
```diff
@@ -0,0 +1,477 @@
...
+
+	pluginData, err := a.pd.Update(ctx, reqID, func(existing PluginData) (PluginData, error) {
+		if len(existing.TeamsData) == 0 {
+			return PluginData{}, nil
+		}
+
+		// If resolution field is not empty then we already resolved the incident before. In this case we just quit.
+		if existing.AccessRequestData.ResolutionTag != pd.Unresolved {
+			return PluginData{}, nil
+		}
```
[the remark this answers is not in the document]
gzigzigzeo: Sure, we should, fixed.

lib/plugindata/cas.go
```diff
@@ -0,0 +1,166 @@
...
+			return emptyData, trace.NotFound("Plugin data does not exist")
+		}
+
+		cbData := *oldData
+		expectData := *oldData
+
+		newData, err := cb(cbData)
+		if err != nil && !trace.IsCompareFailed(err) {
+			return emptyData, trace.CompareFailed("Callback failed to update PluginData")
+		}
```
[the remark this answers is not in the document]
gzigzigzeo: It would not overwrite existing data because `updatePluginData` would not allow it. But you are correct, we should backoff immediately. Fixed.

lib/plugindata/cas.go
gzigzigzeo: Not exactly :-) We need callback and backoff because it is an extension to the update method. Let's follow YAGNI here: I update comment with "when it is required". As we are not using it right now, I do not see a good reason to add it.

access/ms-teams/app.go
gzigzigzeo: @Joerger Thank you for your thoroughness!

gzigzigzeo:
@Joerger Thank you for your thouroughness!

lib/plugindata/cas.go
```diff
@@ -0,0 +1,166 @@
...
+			return emptyData, trace.Wrap(err)
+		}
+
+		err = c.updatePluginData(ctx, resource, newData, expectData)
+		if err == nil {
+			return newData, nil
+		}
+		if !trace.IsCompareFailed(err) {
+			return emptyData, trace.Wrap(err)
+		}
```
Joerger, line 130: Don't we still need the backoff here as well?
```suggestion
		}
		if err := backoff.Do(ctx); err != nil {
			return emptyData, trace.Wrap(err)
		}
```

Joerger (approved):
LGTM, assuming Marco's manual testing doesn't uncover any issues

marcoandredinis:
Just tested this in a real (it was actually a sandboxed) account and it works as expected :+1: 
image

I just have a little suggestion: increasing the timeout of the WatcherJob
The current value of 5s was not enough to start process the request, fetch user info and send the message
My suggestion is to set it to at least 15 seconds as default
Ideally this would be part of the configuration

A workaround for this is to turn on the `preload` in configuration

marcoandredinis (approved):

tcsc (approved):

hugoShaka (approved):
LGTM, I would like to address a couple of thing but want to avoid the same giant PR than for the operator: I'll merge this PR with the typo changes and will send smaller/more reviewable PRs later.

lib/plugindata/access_request.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// DecodePluginData deserializes a string map to PluginData struct.
```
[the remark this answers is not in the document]
hugoShaka: Fixed

lib/plugindata/access_request.go
```diff
@@ -0,0 +1,61 @@
...
+	if str := dataMap["reviews_count"]; str != "" {
+		fmt.Sscanf(str, "%d", &data.ReviewsCount)
+	}
+	data.ResolutionTag = ResolutionTag(dataMap["resolution"])
+	data.ResolutionReason = dataMap["resolve_reason"]
+
+	return
+}
+
+// DecodePluginData deserializes a string map to PluginData struct.
```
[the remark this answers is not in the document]
hugoShaka: Fixed

access/ms-teams/msapi/client.go
```diff
@@ -0,0 +1,143 @@
...
+	token   tokenWithTTL
+	baseURL string
+	config  Config
+}
+
+// request represents generic request structure
+type request struct {
+	// Method HTTP method
+	Method string
+	// Path path to a resource
```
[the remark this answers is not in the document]
hugoShaka: Fixed

access/ms-teams/_tpl/teleport-ms-teams-role.yaml
```diff
@@ -0,0 +1,16 @@
+kind: user
```
hugoShaka, line 1: Should specify the role first as `tctl create -f <filename>` will fail with: 
```
ERROR: role teleport-ms-teams is not found
```

access/ms-teams/msapi/bot_framework_client.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+}
+
+// PostAdaptiveCardActivity sends an activity to the chat, content is AdaptiveCard
+func (c *BotFrameworkClient) PostAdaptiveCardActivity(ctx context.Context, botID, chatID, card, updateID string) (string, error) {
+	m := postMessagePayload{
+		Type: "message",
+		From: postMessagePayloadFrom{
+			ID:   botDesignator + botID,
+			Name: "TeleBot",
```
hugoShaka, line 90: This is weird to hardcode this here: how does this behave with the app installed as an other name ?

PR #619, opened 2022-08-01
Discord mode for Slack

gzigzigzeo:
Draft, but working version


tcsc:
I don't actually understand what functionality this is changing or adding - What actually _is_ a Discord mode for Slack? - so I'm probably not the best for review here.

Issue #604, opened 2022-07-14
[event-handler] Validate required config fields up front and pass more helpful error messages to the user


ptgott:
When configuring the Fluentd plugin, I accidentally left out the `teleport.client_key`, `teleport.client_crt`, and `teleport.root_cas` config fields. When I ran the plugin, I saw the following error message:

```
ERRO   Terminating with fatal error [1]... error:[
ERROR REPORT:
Original Error: *fs.PathError open : no such file or directory
Stack Trace:
        /go/pkg/mod/github.com/gravitational/teleport/api@v0.0.0-20220701151950-c3f9fd84a104/client/credentials.go:115 github.com/gravitational/teleport/api/client.(*keypairCreds).TLSConfig
        /go/pkg/mod/github.com/gravitational/teleport/api@v0.0.0-20220701151950-c3f9fd84a104/client/client.go:186 github.com/gravitational/teleport/api/client.connect.func3              /usr/local/go/src/runtime/asm_amd64.s:1581 runtime.goexit
User Message:
        open : no such file or directory] lib/bail.go:14
```

I had to debug the event handler with `dlv` and poke around the source to discover the issue. Frustratingly, it had to do with the missing config fields, which were passed as empty strings to `LoadX509KeyPair`. 

The event handler should perform some quick validation before it attempts to load the credentials files and call `client.New`:

https://github.com/gravitational/teleport-plugins/blob/90a961bd152a042a2fcbf50533e3ca5e50a29b94/event-handler/teleport_events_watcher.go#L75-L86

We should also see if there are any other config fields the event handler should validate earlier on startup.

PR #627, opened 2022-08-05
Check that required files in StartCmdConfig exist

EdwardDowling:
closes [604](https://github.com/gravitational/teleport-plugins/issues/604)

### DIFF
--- a/event-handler/cli.go
+++ b/event-handler/cli.go
@@ -65,7 +65,7 @@ type TeleportConfig struct {
 	TeleportKey string `help:"Teleport TLS key file" type:"existingfile" env:"FDFWD_TELEPORT_KEY"`
 }
 
-// Validate performs some basic validation on TeleportConfig that can't be handled using the kong defaulting
+// Check verifies that a valid configuration is set
 func (cfg *TeleportConfig) Check() error {
 	provided := stringset.NewWithCap(3)
 	missing := stringset.NewWithCap(3)
@@ -220,7 +220,7 @@ func (c *StartCmdConfig) Validate() error {
 		t := c.StartTime.Truncate(time.Second)
 		c.StartTime = &t
 	}
-	if err := c.TeleportConfig.Validate(); err != nil {
+	if err := c.TeleportConfig.Check(); err != nil {
 		return trace.Wrap(err)
 	}
 	c.SkipSessionTypes = lib.SliceToAnonymousMap(c.SkipSessionTypesRaw)
--- a/event-handler/cli.go
+++ b/event-handler/cli.go
@@ -94,6 +94,10 @@ func (cfg *TeleportConfig) Validate() error {
 			strings.Join(missing.ToSlice(), ", "),
 		)
 	}
+
+	if cfg.TeleportIdentityFile != "" && len(provided) != 0 {
+		return trace.BadParameter("configuration setting `identity` is mutually exclusive with the `cert`, `key` and `ca` settings")
+	}
 	if len(provided) == 0 && cfg.TeleportIdentityFile == "" {
 		return trace.BadParameter("neither `identity` file nor `cert`, `key` and `ca` files configured")
 	}
--- a/event-handler/cli.go
+++ b/event-handler/cli.go
@@ -18,12 +18,15 @@ package main
 
 import (
 	"context"
+	"strings"
 	"time"
 
 	"github.com/alecthomas/kong"
 
 	"github.com/gravitational/teleport-plugins/event-handler/lib"
 	"github.com/gravitational/teleport-plugins/lib/logger"
+	"github.com/gravitational/teleport-plugins/lib/stringset"
+	"github.com/gravitational/trace"
 )
 
 // FluentdConfig represents fluentd instance configuration
@@ -47,7 +50,7 @@ type FluentdConfig struct {
 // TeleportConfig is Teleport instance configuration
 type TeleportConfig struct {
 	// TeleportAddr is a Teleport addr
-	TeleportAddr string `help:"Teleport addr" env:"FDFWD_TELEPORT_ADDR"`
+	TeleportAddr string `help:"Teleport addr" env:"FDFWD_TELEPORT_ADDR" defualt:"localhost:3025"`
 
 	// TeleportIdentityFile is a path to Teleport identity file
 	TeleportIdentityFile string `help:"Teleport identity file" type:"existingfile" name:"teleport-identity" env:"FDFWD_TELEPORT_IDENTITY"`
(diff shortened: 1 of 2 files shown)


Joerger (commented):
IIUC the consensus on reflect is to only use it when the generalization it provides is necessary or profoundly useful, otherwise using basic Go is preferred. I see that `event-handler` is using Kong, which relies on reflection in a similar way, so I'm open to reflection being the solution if you feel strongly about it.

But... instead of using refection here, what do you think of making a `CheckAndSetDefaults` for the config structs and checking field by field? Like this: 
 - https://github.com/gravitational/teleport-plugins/blob/a373950beb9564953c9b215aed9155ff8a068ac6/lib/config.go#L36
 - https://github.com/gravitational/teleport-plugins/blob/a373950beb9564953c9b215aed9155ff8a068ac6/lib/config.go#L54

event-handler/teleport_events_watcher.go
```diff
@@ -97,6 +103,31 @@ func NewTeleportEventsWatcher(
 	return &tc, nil
 }
 
+// validateConfig validates the given StartCmdConfig
+func validateConfig(cfg StartCmdConfig) error {
+	errs := []error{}
+	// Check all config fields with the tag existingfile point to an existing file.
+	for _, config := range []interface{}{cfg.FluentdConfig, cfg.TeleportConfig, cfg.IngestConfig} {
```
r0mant, line 110: Not sure I understand, why do we need to use reflection here instead of checking that these fields aren't empty?

r0mant (changes requested):
Let's please not use reflection for this.

event-handler/teleport_events_watcher_test.go
```diff
@@ -108,3 +109,97 @@ func TestNext(t *testing.T) {
...
+				IngestConfig{},
+				LockConfig{},
+			},
+			errorMessage: "required file \"tempFile.Name()\" does not exist, required config field \"TeleportKey\" not set",
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			err := validateConfig(tc.cfg)
+			if tc.errorMessage != "" {
+				require.EqualError(t, err, tc.errorMessage)
```
zmb3, line 199: I wouldn't assert on a specific error message - this is easy to break if we change the wording of an error.

In general, I would expect that invalid configuration would result in a `trace.BadParameter` error - so how about we just assert that `trace.IsBadParameter(err)` is true?

event-handler/teleport_events_watcher.go
EdwardDowling: Was thinking it may be handy if any more eixistingfile fields were added, in hindsight this was kind of made no sense. Will just remove the reflection and have a check for each field.

EdwardDowling:
> IIUC the consensus on reflect is to only use it when the generalization it provides is necessary or profoundly useful, otherwise using basic Go is preferred. I see that `event-handler` is using Kong, which relies on reflection in a similar way, so I'm open to reflection being the solution if you feel strongly about it.
> 
> But... instead of using refection here, what do you think of making a `CheckAndSetDefaults` for the config structs and checking field by field? Like this:
> 
>     * https://github.com/gravitational/teleport-plugins/blob/a373950beb9564953c9b215aed9155ff8a068ac6/lib/config.go#L36
> 
>     * https://github.com/gravitational/teleport-plugins/blob/a373950beb9564953c9b215aed9155ff8a068ac6/lib/config.go#L54

I am not attached to the reflection at all, no idea why I thought that was a good idea before. I've added in a Validation func for the fields without any reflection. I'm not sure the CheckAndSetDefaults makes sense if kong is being used as the defaults should probably use the defaulting functionality kong provides instead.

event-handler/cli.go
```diff
@@ -47,7 +50,7 @@ type FluentdConfig struct {
 // TeleportConfig is Teleport instance configuration
 type TeleportConfig struct {
 	// TeleportAddr is a Teleport addr
-	TeleportAddr string `help:"Teleport addr" env:"FDFWD_TELEPORT_ADDR"`
+	TeleportAddr string `help:"Teleport addr" env:"FDFWD_TELEPORT_ADDR" defualt:"localhost:3025"`
```
Joerger, line 53:
```suggestion
	TeleportAddr string `help:"Teleport addr" env:"FDFWD_TELEPORT_ADDR" default:"localhost:3025"`
```

event-handler/teleport_events_watcher_test.go
```diff
@@ -108,3 +109,56 @@ func TestNext(t *testing.T) {
...
+			cfg: StartCmdConfig{
+				FluentdConfig{},
+				TeleportConfig{
+					TeleportIdentityFile: "not_empty_string",
+					TeleportCA:           "not_empty_string",
+				},
+				IngestConfig{},
+				LockConfig{},
+			},
+			wantError: false,
```
Joerger, line 152:
```suggestion
			wantError: true,
```

event-handler/teleport_events_watcher_test.go
```diff
@@ -108,3 +109,56 @@ func TestNext(t *testing.T) {
...
+	}{
+		{
+			name: "Identity and teleport cert/ca/key files configured",
+			cfg: StartCmdConfig{
+				FluentdConfig{},
+				TeleportConfig{
+					TeleportIdentityFile: "not_empty_string",
+					TeleportCA:           "not_empty_string",
+					TeleportCert:         "not_empty_string",
+					TeleportKey:          "not_empty_string",
```
Joerger, line 127: Let's make identity file mutually exclusive with CA/Cert/Key, like in the `CheckAndSetDefaults` example.

event-handler/cli.go
```diff
@@ -18,12 +18,15 @@ package main
...
 	"context"
+	"strings"
 	"time"
 
 	"github.com/alecthomas/kong"
 
 	"github.com/gravitational/teleport-plugins/event-handler/lib"
 	"github.com/gravitational/teleport-plugins/lib/logger"
+	"github.com/gravitational/teleport-plugins/lib/stringset"
+	"github.com/gravitational/trace"
```
Joerger, line 29: nit: import grouping for external packages
```suggestion
	"github.com/gravitational/teleport-plugins/event-handler/lib"
	"github.com/gravitational/teleport-plugins/lib/logger"
	"github.com/gravitational/teleport-plugins/lib/stringset"

	"github.com/alecthomas/kong"
	"github.com/gravitational/trace"
```

event-handler/teleport_events_watcher_test.go
```diff
@@ -108,3 +109,80 @@ func TestNext(t *testing.T) {
...
+			IngestConfig{},
+			LockConfig{},
+		},
+		wantError: true,
+	}, {
+		name: "Some of teleport cert/key/ca unset",
+		cfg: StartCmdConfig{
+			FluentdConfig{},
+			TeleportConfig{
+				TeleportIdentityFile: "not_empty_string",
```
Joerger, line 170:
```suggestion
```

event-handler/teleport_events_watcher_test.go
```diff
@@ -22,6 +22,7 @@ import (
 
 	"github.com/gravitational/teleport/api/types"
 	"github.com/gravitational/teleport/api/types/events"
+	"github.com/gravitational/trace"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/net/context"
```
Joerger, line 27: nit:
```suggestion
	"github.com/gravitational/teleport/api/types"
	"github.com/gravitational/teleport/api/types/events"
	
	"github.com/gravitational/trace"
	"github.com/stretchr/testify/require"
	"golang.org/x/net/context"
```

Joerger (approved):

event-handler/cli.go
```diff
@@ -62,6 +65,45 @@ type TeleportConfig struct {
 	TeleportKey string `help:"Teleport TLS key file" type:"existingfile" env:"FDFWD_TELEPORT_KEY"`
 }
 
+// Validate performs some basic validation on TeleportConfig that can't be handled using the kong defaulting
+func (cfg *TeleportConfig) Validate() error {
```
r0mant, line 69: Nit: To make it consistent with how we usually name other methods that validate config structs in our codebase.
```suggestion
func (cfg *TeleportConfig) Check() error {
```

event-handler/cli.go
```diff
@@ -62,6 +65,45 @@ type TeleportConfig struct {
 	TeleportKey string `help:"Teleport TLS key file" type:"existingfile" env:"FDFWD_TELEPORT_KEY"`
 }
 
+// Validate performs some basic validation on TeleportConfig that can't be handled using the kong defaulting
```
r0mant, line 68: What's "kong defaulting"?

r0mant (approved):

event-handler/cli.go
EdwardDowling: Kong supports using struct tags to set default values. Have updated comment to less verbose

PR #634, opened 2022-08-17
Truncate reason field for jira bot

EdwardDowling:
closes [159](https://github.com/gravitational/teleport-private/issues/159)


access/jira/client.go
```diff
@@ -39,6 +39,9 @@ const (
 
 	jiraMaxConns    = 100
 	jiraHTTPTimeout = 10 * time.Second
+	// Jira has a 300,000 character limit for the reason field so we
+	// truncate all reasons to a generous but conservative limit
+	jiraReasonLimit = 3000
```
r0mant, line 44: If Jira has a limit of 300K, why do we limit to 3K?

r0mant (commented):
@EdwardDowling Have you tested this with real Jira setup?

access/jira/client.go
EdwardDowling: 30k sounds good to me.

r0mant:
@Joerger Can you please take a look as well?

access/jira/client.go
```diff
@@ -39,6 +39,9 @@ const (
 
 	jiraMaxConns    = 100
 	jiraHTTPTimeout = 10 * time.Second
+	// Jira has a 300,000 character limit for the reason field so we
+	// truncate all reasons to a generous but conservative limit
+	jiraReasonLimit = 30000
```
Joerger, line 44: It looks like the limit might be 4096, not 300,000. When I run the test below, I get:

```
Error Trace:	jira_test.go:346
Error:      	Received unexpected error:
            	access request reason is too long, max 4096 bytes
Test:       	TestJira/TestIssueCreationWithLargeRequestReason
```

Can you add a link to any documentation in the comment which describes the limit?

access/jira/client.go
EdwardDowling: The actual limit appears to be much smaller than what was listed in the ticket. Seemingly 4096 instead of 300000, I have updated the comment to reflect this.

access/jira/client.go
EdwardDowling: Have set the limit back to 3000 and updated the comment. 
Tried to find some documentation specifying the reason length but was unable to, beyond the info in the initial ticket.

r0mant:
@Joerger Can you take another look?

Joerger (approved):

tcsc (approved):

access/jira/client.go
zmb3: The 4k limit is not imposed by Jira, it's a Teleport-enforced limit.

https://github.com/gravitational/teleport/blob/368f3c31abfd74456267ba1dadd1c5e5914bbd7e/lib/services/access_request.go#L40

This was set in response to a security finding that an attacker could force Teleport to allocate large amounts of memory by submitting access requests with a huge reason field.

access/jira/client.go
```diff
@@ -39,6 +39,9 @@ const (
 
 	jiraMaxConns    = 100
 	jiraHTTPTimeout = 10 * time.Second
+	// Jira has a 4096 character limit for the reason field so we
```
zmb3, line 42:
```suggestion
	// Teleport has a 4096 character limit for the reason field so we
```

PR #637, opened 2022-08-25
access/msteams: Add uninstall command

hugoShaka:
Followup of the `access/msteams` PR #617, part of https://github.com/gravitational/teleport/issues/11818

The way the personal teams notification system work is the following:
- add app to MsTeams instance catalog
- install the app without user interaction to all recipients (apps have to be enabled per-user)
- get the canal ID that was created during installation
- speak with the user through this cana

However if we want to uninstall the App (something went wrong, a new app instance was created, ...) we need to disable it for all users that were contacted or in the recipient list. This PR introduces a command that will uninstall the apps for all users in the `recipients` list.

### DIFF
--- a/access/ms-teams/validate.go
+++ b/access/ms-teams/validate.go
@@ -20,26 +20,9 @@ func validate(configPath, userID string) error {
 	lib.PrintVersion(appName, Version, Gitref)
 	fmt.Println()
 
-	c, err := LoadConfig(configPath)
+	b, _, err := loadConfigAndCheckApp(configPath)
 	if err != nil {
 		return trace.Wrap(err)
-	}
-
-	fmt.Printf(" - Checking application %v status...\n", c.MSAPI.TeamsAppID)
-
-	b, err := NewBot(c.MSAPI, "local", "")
-	if err != nil {
-		return trace.Wrap(err)
-	}
-
-	teamApp, err := b.GetTeamsApp(context.Background())
-	if trace.IsNotFound(err) {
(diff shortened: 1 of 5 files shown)


access/ms-teams/validate.go
```diff
@@ -134,3 +117,28 @@ func validate(configPath, userID string) error {
...
+	}
+
+	fmt.Printf(" - Checking application %v status...\n", c.MSAPI.TeamsAppID)
+
+	b, err := NewBot(c.MSAPI, "local", "")
+	if err != nil {
+		return nil, nil, trace.Wrap(err)
+	}
+
+	teamApp, err := b.GetTeamsApp(context.Background())
```
r0mant, line 134: I would pass context from the caller to `loadConfigAndCheckApp` method and use it here.

r0mant (approved):

access/ms-teams/validate.go
```diff
@@ -134,3 +117,28 @@ func validate(configPath, userID string) error {
 
 	return nil
 }
+
+func loadConfigAndCheckApp(configPath string) (*Bot, *Config, error) {
```
marcoandredinis, line 121: :nail_care: 
I would probably separate this in two methods
`LoadConfig` and `CheckApp`

marcoandredinis (approved):

access/ms-teams/validate.go
hugoShaka: Done

access/ms-teams/validate.go
hugoShaka: Done

PR #646, opened 2022-09-01
access/ms-teams: make the plugin able to post to team channels

hugoShaka:
Part of https://github.com/gravitational/teleport/issues/11818

This commit extends the user and userData concepts to support channels. They have been renamed respectively recipient and recipientData.

Channels can be specified in the configuration using their URL.
Everything is not perfect, Microsoft might change the URL format (though we'll still be able to parse old URLs, so service won't break if this happens, only new channel addition will be harder until we update our parsing logic), but this is the easiest way to have:
- a single string (to avoid having a custom `role_to_recipient` format)
- a way to uniquely point to a channel (channel names can change)
- a way for the user to look at the configuration and know what recipient this is
- have required information to check app installation (channel ID can be sufficient for the bot SDK, but not for the azure graph client).

This PR also adds teams/group app installation check. This is not mandatory (we could send message without) but this is a better User Experience to detect permissions issues on startup. With our permissions we could also install ourselves in the Team like we do for users but this feels a bit intrusive. If you think this is a good idea I'll do it.

### DIFF
--- a/access/ms-teams/app.go
+++ b/access/ms-teams/app.go
@@ -148,7 +148,9 @@ func (a *App) initBot(ctx context.Context) error {
 		return trace.Wrap(err)
 	}
 
-	log.WithField("name", teamsApp.DisplayName).WithField("id", teamsApp.ID).Info("MS Teams app found in org app store")
+	log.WithField("name", teamsApp.DisplayName).
+		WithField("id", teamsApp.ID).
+		Info("MS Teams app found in org app store")
 
 	if !a.conf.Preload {
 		return nil
@@ -161,7 +163,10 @@ func (a *App) initBot(ctx context.Context) error {
 		if err != nil {
 			return trace.Wrap(err)
 		}
-		log.WithField("recipient", recipient).WithField("chat_id", recipientData.Chat.ID).Info("Recipient found, chat found")
+		log.WithField("recipient", recipient).
+			WithField("chat_id", recipientData.Chat.ID).
+			WithField("kind", recipientData.Kind).
+			Info("Recipient found, chat found")
 	}
 
 	log.Info("Recipient data preloaded and cached.")
@@ -455,7 +460,7 @@ func (a *App) getMessageRecipients(ctx context.Context, req types.AccessRequest)
 	// This can happen if this set contains the channel `C` and the email for channel `C`.
 	recipientSet := stringset.New()
 
-	validEmailsSuggReviewers := []string{}
+	var validEmailsSuggReviewers []string
 	for _, reviewer := range req.GetSuggestedReviewers() {
 		if !lib.IsEmail(reviewer) {
 			log.Warningf("Failed to notify a suggested reviewer: %q does not look like a valid email", reviewer)
--- a/access/ms-teams/app.go
+++ b/access/ms-teams/app.go
@@ -154,17 +154,17 @@ func (a *App) initBot(ctx context.Context) error {
 		return nil
 	}
 
-	log.Info("Preloading user data...")
+	log.Info("Preloading recipient data...")
 
-	for _, userID := range a.conf.Recipients.GetAllRecipients() {
-		userData, err := a.bot.FetchUser(ctx, userID)
+	for _, recipient := range a.conf.Recipients.GetAllRecipients() {
+		recipientData, err := a.bot.FetchRecipient(ctx, recipient)
 		if err != nil {
 			return trace.Wrap(err)
 		}
-		log.WithField("user_id", userID).WithField("chat_id", userData.Chat.ID).Info("User found, chat found")
+		log.WithField("recipient", recipient).WithField("chat_id", recipientData.Chat.ID).Info("Recipient found, chat found")
 	}
 
-	log.Info("User data preloaded and cached.")
+	log.Info("Recipient data preloaded and cached.")
 
 	return nil
 }
--- a/access/ms-teams/bot.go
+++ b/access/ms-teams/bot.go
@@ -55,7 +55,7 @@ type Bot struct {
 	botClient *msapi.BotFrameworkClient
 	// mu recipients access mutex
 	mu *sync.RWMutex
-	// apps represents the cache of apps
+	// recipients represents the cache of potential message recipients
 	recipients map[string]RecipientData
 	// webProxyURL represents Web UI address, if enabled
 	webProxyURL *url.URL
--- a/access/ms-teams/bot.go
+++ b/access/ms-teams/bot.go
@@ -2,7 +2,9 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"net/url"
+	"strings"
 	"sync"
 	"time"
 
@@ -13,13 +15,13 @@ import (
 	"github.com/gravitational/trace"
 )
 
-// UserData represents cached data for a user
-type UserData struct {
-	// User ID user id
+// RecipientData represents cached data for a recipient (user or channel)
+type RecipientData struct {
+	// ID identifies the recipient, for users it is the UserID, for channels it is "tenant/group/channelName"
 	ID string
-	// App app installation for user
+	// App installation for the recipient
 	App msapi.InstalledApp
-	// Chat chat for user
+	// Chat for the recipient
 	Chat msapi.Chat
 }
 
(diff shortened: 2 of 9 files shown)


access/ms-teams/bot.go
```diff
@@ -33,10 +44,10 @@ type Bot struct {
 	graphClient *msapi.GraphClient
 	// botClient represents MS Bot Framework client
 	botClient *msapi.BotFrameworkClient
-	// mu users access mutex
+	// mu recipients access mutex
 	mu *sync.RWMutex
 	// apps represents the cache of apps
```
tcsc, line 49: Update comment to match code?
```suggestion
	// recipients represents the cache of potential message recipients
```

access/ms-teams/bot.go
```diff
@@ -61,7 +72,7 @@ func NewBot(c msapi.Config, clusterName, webProxyAddr string) (*Bot, error) {
 		Config:      c,
 		graphClient: msapi.NewGraphClient(c),
 		botClient:   msapi.NewBotFrameworkClient(c),
-		users:       make(map[string]UserData),
+		recipients:  make(map[string]RecipientData),
```
tcsc, line 75: I'm assuming that the fundamental point of this change is that access requests can now be sent to a channel or user, so just calling the field `users ` is now misleading?

access/ms-teams/bot.go
```diff
@@ -125,52 +136,72 @@ func (b Bot) UninstallAppForUser(ctx context.Context, userIDOrEmail string) erro
...
-	if err != nil {
-		return &UserData{}, trace.Wrap(err)
-	}
-
-	var installedApp *msapi.InstalledApp
-
-	installedApp, err = b.graphClient.GetAppForUser(ctx, b.teamsApp, userID)
-	if trace.IsNotFound(err) {
-		err := b.graphClient.InstallAppForUser(ctx, userID, b.teamsApp.ID)
+	// Check if the recipient is a channel
```
tcsc, line 154: What happens in the case of 2 simultaneous calls to `FetchRecipient` that miss the cache? I get that the `recipients` map itself is protected by `mu`, but does it matter that you may have multiple copies of the recipient data being generated, and the last write to the cache wins?

access/ms-teams/bot.go
```diff
@@ -193,15 +224,56 @@ func (b Bot) getUserID(ctx context.Context, userIDOrEmail string) (string, error
 	return userIDOrEmail, nil
 }
 
+// checkChannelURL receives a recipient and checks if it is a channel URL.
+// If it is the case, the URL is parsed and the channel RecipientData is returned
+func (b Bot) checkChannelURL(recipient string) (*Channel, bool) {
```
tcsc, line 229: Does this need to be a method?

access/ms-teams/bot.go
```diff
@@ -125,52 +136,72 @@ func (b Bot) UninstallAppForUser(ctx context.Context, userIDOrEmail string) erro
 	return trace.Wrap(err)
 }
 
-// FetchUser fetches app id for user, installs app for a user if missing, fetches chat id and saves
-// everything to cache. This method is used for priming the cache. Returns trace.NotFound if a
+// FetchRecipient checks if recipient is a user or a channel, installs app for a user if missing, fetches chat id
+// and saves everything to cache. This method is used for priming the cache. Returns trace.NotFound if a
 // user was not found.
-func (b Bot) FetchUser(ctx context.Context, userIDOrEmail string) (*UserData, error) {
+func (b Bot) FetchRecipient(ctx context.Context, recipient string) (*RecipientData, error) {
```
tcsc, line 142: Umm.... Perhaps a silly question, but is there any reason why we're using Value receivers on these methods?

I mean, yes, it works OK because `mu` is a pointer to a `*Mutex` rather than the `Mutex` itself (and thus avoids the copy restriction on `Mutex`es. Just wondering if there was any architectural reason for not having a pointer receiver?

access/ms-teams/bot.go
marcoandredinis: Maps work by reference (as well as the `*Mutex`)
So, when the map is updated we are actually mutating the map and not some copy.
Being protected by the *Mutex means we are not getting two different states for the map.
I might be missing something, but seems correct to me

But I do agree with you that we should probably change this method to a pointer receiver.

access/ms-teams/bot.go
```diff
@@ -125,52 +136,72 @@ func (b Bot) UninstallAppForUser(ctx context.Context, userIDOrEmail string) erro
...
-	var installedApp *msapi.InstalledApp
-
-	installedApp, err = b.graphClient.GetAppForUser(ctx, b.teamsApp, userID)
-	if trace.IsNotFound(err) {
-		err := b.graphClient.InstallAppForUser(ctx, userID, b.teamsApp.ID)
+	// Check if the recipient is a channel
+	channel, isChannel := b.checkChannelURL(recipient)
+	if isChannel {
+		// A team and a group are different but in MsTeams the team is associated to a group and will have the same id.
+		installedApp, err := b.graphClient.GetAppForTeam(ctx, b.teamsApp, channel.Group)
```
marcoandredinis, line 158: When we do the same for the user we also create the app_user chat
The same doesn't happen here: we only "get" the channel but there's no creation if it doesn't exist

marcoandredinis (approved):

access/ms-teams/app.go
```diff
@@ -154,17 +154,17 @@ func (a *App) initBot(ctx context.Context) error {
...
 
-	for _, userID := range a.conf.Recipients.GetAllRecipients() {
-		userData, err := a.bot.FetchUser(ctx, userID)
+	for _, recipient := range a.conf.Recipients.GetAllRecipients() {
+		recipientData, err := a.bot.FetchRecipient(ctx, recipient)
 		if err != nil {
 			return trace.Wrap(err)
 		}
-		log.WithField("user_id", userID).WithField("chat_id", userData.Chat.ID).Info("User found, chat found")
+		log.WithField("recipient", recipient).WithField("chat_id", recipientData.Chat.ID).Info("Recipient found, chat found")
```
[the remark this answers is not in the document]
hugoShaka: Done.

access/ms-teams/bot.go
marcoandredinis: So, it might happen that we try to use a channel but it won't work because the Admin didn't grant permission for it when installing the bot :thinking: 

If so, I think we should add a check for write access for that particular channel when starting up the app (when `preload=true`)

hugoShaka: But, FetchRecipient is already called during preload, and the check logic is in FetchRecipient. Maybe I'm entirely missing the point but I think we already are doing this preload check

access/ms-teams/bot.go
tcsc: @marcoandredinis - It's not the `map` update I was concerned about, that's totally fine. What I was concerned about is was the generation of multiple copies of the `RecipientData`, with them being replaced in the map one after the other. Is it possible someone will have a copy of the stale data? Will it matter? Are there any side effects of that?

And I think @hugoShaka's says "no, it won't" - so with that I'm all good.

access/ms-teams/bot.go
```diff
@@ -125,52 +136,72 @@ func (b Bot) UninstallAppForUser(ctx context.Context, userIDOrEmail string) erro
...
+		userID, err := b.getUserID(ctx, recipient)
+		if err != nil {
+			return &RecipientData{}, trace.Wrap(err)
+		}
+
+		var installedApp *msapi.InstalledApp
 
 		installedApp, err = b.graphClient.GetAppForUser(ctx, b.teamsApp, userID)
+		if trace.IsNotFound(err) {
+			err := b.graphClient.InstallAppForUser(ctx, userID, b.teamsApp.ID)
```
tcsc, line 192: Is it OK that this could be called more than once due to cache misses? Does "already installed" manifest as an error?

access/ms-teams/bot.go
marcoandredinis: Oh right :+1: 
Thank you :bow:

access/ms-teams/bot.go
marcoandredinis: Oh I'm sorry
Totally missed the point here
We should be good then :+1:

access/ms-teams/bot.go
hugoShaka: Yup, this could be an issue as discussed here: https://github.com/gravitational/teleport-plugins/pull/646#discussion_r961470167

https://github.com/gravitational/teleport-plugins/pull/646/commits/1fbe4bd0d517d176b498740c7f6999e04f981416 should handle this case

tcsc (approved):

PR #660, opened 2022-09-19
ci: Implement code scanning with CodeQL and add linting for terraform

reedloden:
Use CodeQL to scan all pull requests and pushes for known security issues.

https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/about-code-scanning-with-codeql

Use the shared workflow for linting terraform changes.


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,55 @@
...
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+    paths-ignore:
+      - 'rfd/**'
```
zmb3, line 11: If CodeQL doesn't do anything for helm charts, I'd also ignore `charts/**`.

Also, the plugins all use toml for configuration, so you can ignore `**.toml` as well.

zmb3 (approved):

.github/workflows/codeql-analysis.yml
reedloden: I'm hesitant to ignore any code changes, as it would require us maintaining all our CodeQL workflows across tons of different repositories. But perhaps. Realistically, we should add https://github.com/helm/chart-testing-action to have proper testing.

TOML seems like it might be supported under Python scanning support in CodeQL -- https://codeql.github.com/docs/codeql-overview/supported-languages-and-frameworks/#python-built-in-support -- would need to double-check for sure.

wadells (approved):
LGTM.

PR #662, opened 2022-09-21
Update helm publishing pipeline

tcsc:
Updates the helm publishing pipeline to use the new charts repo.

The new repo has a different permission set for the deploy user than the old one, so we need to switch from the drone S2 plugin to just running `s3 sync`


tcsc:
Made obsolete by role assumption changes

PR #667, opened 2022-09-27
Discord access-request plugin

hugoShaka:
This PR introduces the Discord access-request plugin. Part of https://github.com/gravitational/teleport/issues/9563

It relies on the access-request messaging refactoring from PR #666.

Documentation, charts and release-pipeline will come in a subsequent PR to avoid overloading this one.

The bot has the following limitations:
- it cannot initiate a DM with a user (I was not able to use the Discord endpoint with a bot token)
- it canot lookup a channel by its pretty name (it would require implementing a cache) and have a startup logic


hugoShaka:
I am not sure what testing strategy to adopt here. The `common` logic is already covered by the existing slack tests, the untested parts are the Discord client (`access/discord/bot.go`) and the config parsing.

I can duplicate the slack test suite and sightly adapt it to make it work with Discord. Ideally the messaging testing suite would be reusable, but I don't think another refactoring is worth the time right now.

WDYT? @r0mant @Joerger @marcoandredinis

access/discord/Makefile
```diff
@@ -0,0 +1,77 @@
+VERSION=10.2.2
```
hugoShaka, line 1: TODO: update version

access/discord/bot.go
```diff
@@ -0,0 +1,155 @@
...
+	for _, review := range reviews {
+		text, err := common.MsgReview(review)
+		if err != nil {
+			return ""
+		}
+
+		result += text + "\n"
+	}
+
+	return "\n" + result
```
Joerger, line 143: nit: joining a list seems better than appending to a string here. Also this TODO seems stale.
```suggestion
	reviewMessages := make([]string, len(reviews)
	for _, review := range reviews {
		msg, err := common.MsgReview(review)
		if err != nil {
			return ""
		}
		reviewMessages[i] = msg
	}

	return "\n" + strings.Join(reviewMessages, "\n")
```

access/discord/bot.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func (b DiscordBot) FetchRecipient(ctx context.Context, recipient string) (*common.Recipient, error) {
+	// Discord does not support resolving email address, we only return the channel name
+	// TODO: check if channel exists ?
```
Joerger, line 148: Stale TODO?
```suggestion
```

access/discord/example_config.toml
```diff
@@ -0,0 +1,36 @@
...
+# When using --format=file:
+# identity = "/var/lib/teleport/plugins/discord/auth_id"    # Identity file
+#
+# When using --format=tls:
+# client_key = "/var/lib/teleport/plugins/discord/auth.key" # Teleport TLS secret key
+# client_crt = "/var/lib/teleport/plugins/discord/auth.crt" # Teleport TLS certificate
+# root_cas = "/var/lib/teleport/plugins/discord/auth.cas"   # Teleport CA certs
+
+[discord]
+# Slack Bot OAuth token
```
Joerger, line 21:
```suggestion
# Discord Bot OAuth token
```

access/discord/example_config.toml
```diff
@@ -0,0 +1,36 @@
...
+token = "xoxb-11xx"
+
+[role_to_recipients]
+# Map roles to recipients.
+#
+# Provide discord user_email/channel recipients for access requests for specific roles.
```
Joerger, line 27: What does `user_email` do if sending DMs is not supported?

access/discord/config.go
```diff
@@ -0,0 +1,139 @@
...
+
+	if len(c.Discord.Recipients) > 0 {
+		if len(c.Recipients) > 0 {
+			return trace.BadParameter("provide either discord.recipients or role_to_recipients, not both.")
+		}
+
+		c.Recipients = common.RawRecipientsMap{
+			types.Wildcard: c.Discord.Recipients,
+		}
+	}
```
Joerger, line 89: Let's not support the old recipients field in this new bot.
```suggestion
```

access/discord/main.go
```diff
@@ -0,0 +1,99 @@
...
+	if err = logger.Setup(logConfig); err != nil {
+		return trace.Wrap(err)
+	}
+	if debug {
+		logger.Standard().Debugf("DEBUG logging enabled")
+	}
+
+	if conf.Discord.Recipients != nil {
+		logger.Standard().Warn("The discord.recipients config option is deprecated, set role_to_recipients[\"*\"] instead for the same functionality")
+	}
```
Joerger, line 90:
```suggestion
```

Joerger (commented):
For testing, I wouldn't worry about the config parsting, In fact I think you can remove `slack/config_test.go` as that test is covered by `common/recipient_test.go`. But we probably do want some Discord versions of the other slack tests in `slack_test.go` since that logic isn't generalized for now.

access/discord/bot.go
hugoShaka: Yeah, I wanted to do channel lookup but this involes doing queries that must be cached. To avoid complexifying the plugin and have a first working release I chose to omit the feature

hugoShaka:
@Joerger I added the test suite, it took a bit longer than I expected. Everything should be ready for a second review.

I also tried to improved discord's formatting which was really ugly. Now reviews appear as message embeds colored based on the result. I'll run a last test and post a screenshot here.

access/discord/main.go
```diff
@@ -0,0 +1,95 @@
...
+		return trace.Wrap(err)
+	}
+	if debug {
+		logger.Standard().Debugf("DEBUG logging enabled")
+	}
+
+	app := NewDiscordApp(conf)
+	go lib.ServeSignals(app, 15*time.Second)
+
+	logger.Standard().Infof("Starting Teleport Access Slack Plugin %s:%s", Version, Gitref)
```
marcoandredinis, line 91:
```suggestion
	logger.Standard().Infof("Starting Teleport Access Discord Plugin %s:%s", Version, Gitref)
```

access/discord/bot.go
```diff
@@ -0,0 +1,183 @@
...
+			}).
+			Patch("/channels/" + msg.ChannelID + "/messages/" + msg.MessageID)
+		if err != nil {
+			errors = append(errors, trace.Wrap(err))
+		}
+	}
+
+	if len(errors) > 0 {
+		return errors[0]
+	}
```
marcoandredinis, line 129: We're losing errors here
Could we print them out? or return all of them using `trace.Aggregate`

access/discord/config.go
```diff
@@ -0,0 +1,129 @@
...
+		}).
+		SetHeader("Content-Type", "application/json").
+		SetHeader("Accept", "application/json").
+		SetHeader("Authorization", token)
+
+	// APIURL parameter is set only in tests
+	if endpoint := c.Discord.APIURL; endpoint != "" {
+		client.SetHostURL(endpoint)
+	} else {
+		client.SetHostURL("https://discord.com/api/")
```
marcoandredinis, line 120: :nail_care: 
const

access/discord/config.go
```diff
@@ -0,0 +1,129 @@
...
+		client.OnAfterResponse(onAfterResponseDiscord)
```
marcoandredinis, line 121: Why don't we set this handler when running tests?

access/discord/bot.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func (b DiscordBot) discordMsgText(reqID string, reqData pd.AccessRequestData) string {
+	return "You have a new Role Request:\n" +
+		common.MsgFields(reqID, reqData, b.clusterName, b.webProxyURL) +
+		common.MsgStatusText(reqData.ResolutionTag, reqData.ResolutionReason)
+}
+
+func (b DiscordBot) FetchRecipient(ctx context.Context, recipient string) (*common.Recipient, error) {
+	// Discord does not support resolving email address, we only return the channel name
```
marcoandredinis, line 176: It's actually the channel ID, right?

marcoandredinis:
Overall looks good :+1: 
image

Can you please add a tutorial on how to setup?

access/discord/config.go
hugoShaka: It's taken from the slack access plugin, I think it'sbecause the mockserver does not set `code` when returning errors, which would cause the unmarshalling to `DiscordResponse` to fail.

access/discord/bot.go
hugoShaka: Done

access/discord/config.go
hugoShaka: Done

access/discord/bot.go
hugoShaka: Yup, I improved the comment

marcoandredinis (approved):

Joerger (approved):

PR #669, opened 2022-09-27
Add Machine ID Bot resource for the terraform provider

timothyb89:
This adds a new `teleport_bot` resource to the Terraform provider which allows creation of Teleport bots.

This resource is a bit unusual as bots do not have any server-side state of their own. As such, this includes a small hand-written resource which makes the appropriate API calls and tracks state locally. It doesn't support updates as bot fields are currently immutable; fields users might want to change are marked with `tfsdk.RequiresReplace()` causing the resource to be recreated upon modification.

No data source is currently provided, due to the previously mentioned lack of a server side resource to query for. It's difficult to populate a bot's roles (though technically doable) but not really sane to recover, say, the token ID. This and the (technically) unnecessary replacements may be rectified in the future if new Teleport RPCs are added to support updating and querying for bots.

A new usage example is also included for this resource type.

Fixes gravitational/teleport#14575


timothyb89:
Posting now mainly to solicit feedback. At a minimum this needs tests, but I still want to look into import support. I think we can probably hold off on proper `Update()` support since `RequiresReplace()` mostly does the right thing now.

terraform/test/fixtures/bot_0_create.tf
```diff
@@ -0,0 +1,26 @@
...
+  spec = {
+    roles = ["Bot"]
+    bot_name = local.bot_name
+    join_method = "token"
+  }
+}
+
+resource "teleport_bot" "test" {
+  name = local.bot_name
+  token_id = "bot-test"
```
timothyb89, line 20: I'm a bit bothered that using a reference here fails in CI but works locally, I'd ideally prefer to have:

```terraform
resource "teleport_bot" "test" {
  name = local.bot_name
  token_id = teleport_provision_token.bot_test.metadata.name
  roles = ["terraform"]
}
```
... but CI fails every time with:

```
    bot_test.go:45: Step 1/4 error: Error running pre-apply refresh: exit status 1
        
        Error: Unsupported attribute
        
          on terraform_plugin_test.tf line 29, in resource "teleport_bot" "test":
          29:   token_id = teleport_provision_token.bot_test.metadata.name
        
        This value does not have any attributes.
```

Maybe an ordering issue or something? Hrm.

terraform/example/bot.tf.example
```diff
@@ -0,0 +1,35 @@
...
+resource "random_password" "bot_token" {
+  length           = 32
+  special          = false
+}
+
+resource "time_offset" "bot_example_token_expiry" {
+  offset_hours = 1
+}
+
+resource "teleport_provision_token" "bot_example" {
```
strideynet, line 16: I think I definitely agree with separating the creation of the token and bot for Terraform.

terraform/provider/resource_teleport_bot.go
```diff
@@ -0,0 +1,241 @@
...
+				Description: "The desired TTL for the token if one is created. If unset, a server default is used",
+			},
+			"token_id": {
+				Type:        types.StringType,
+				Optional:    true,
+				Computed:    true,
+				Description: "The bot joining token. If unset, a new random token is created and its name returned, otherwise a preexisting Bot token may be provided for IAM/OIDC joining.",
+
+				// TODO: Consider dropping RequiresReplace() in the future if a
+				// ResetBot() API becomes available.
```
strideynet, line 71: Can we create an issue outlining what `ResetBot()` would do so we just keep track of this as a piece of work we'd want to engage with in future ?

terraform/provider/resource_teleport_bot.go
```diff
@@ -0,0 +1,241 @@
...
+			},
+			"token_ttl": {
+				Type:        types.StringType,
+				Optional:    true,
+				Computed:    true,
+				Description: "The desired TTL for the token if one is created. If unset, a server default is used",
+			},
+			"token_id": {
+				Type:        types.StringType,
+				Optional:    true,
```
strideynet, line 66: I think we should potentially make this a required field and direct people towards specifying a token they've already created. What do you think ?

terraform/provider/resource_teleport_bot.go
```diff
@@ -0,0 +1,241 @@
...
+				Computed:    true,
+				Description: "The name of the generated bot role",
+			},
+			"token_ttl": {
+				Type:        types.StringType,
+				Optional:    true,
+				Computed:    true,
+				Description: "The desired TTL for the token if one is created. If unset, a server default is used",
+			},
+			"token_id": {
```
strideynet, line 64: Is there anything we can do to mark this field as "sensitive" so we don't accidentally output this (as with standard tokens this will be a secret value) ?

terraform/provider/resource_teleport_bot.go
timothyb89: That's a good point. Even though it doesn't "properly" represent the API, it's the only thing that really makes sense with Terraform.

terraform/provider/resource_teleport_bot.go
timothyb89: d'oh, yup, I'll fix that

terraform/provider/resource_teleport_bot.go
timothyb89: This is covered in https://github.com/gravitational/teleport/issues/13091, I'll link it in a comment here

terraform/provider/resource_teleport_bot.go
timothyb89: I've made the field required and left a note explaining why it the provider behaves differently than the API.

terraform/example/bot.tf.example
```diff
@@ -0,0 +1,35 @@
...
+}
+
+resource "random_password" "bot_token" {
+  length           = 32
+  special          = false
+}
+
+resource "time_offset" "bot_example_token_expiry" {
+  offset_hours = 1
+}
```
hugoShaka, line 14: This comment is non blocking, I don't want to hold the feature but there's something I don't understand about the user flow,

As far as I understand, the bot only use the token on join and continue to work while the token is expired? So the token should be short-lived, or maybe long-lived but deleted on first join?

In this example, the offset is based on the first apply time. Hence the same terraform code cannot be re-applied after 1 hour and will fail with:

```
╷
│ Error: Time validation error
│
│   with teleport_provision_token.bot_example,
│   on main.tf line 101, in resource "teleport_provision_token" "bot_example":
│  101: resource "teleport_provision_token" "bot_example" {
│
│ Attribute metadata.expires value must be in the future
╵
```

If I understand correctly, the users have to re-generate a token during each terraform execution (given the time between two execution is higher than the token expiry) which is not this example's behaviour. If the token have a different name than the previous execution it also means the bot will have to be re-registered.

In its current state, I think the way tokens and bots interact with the descriptive Terraform approach might harm the end user. Users likely work around by creating long-lived tokens, somewhat defeating the Teleport pitch "get rid of long-lived secrets". Maybe allowing a bot to be defined without a token and doing nothing if token expiry is in the past would help, but this is outside of this PR's scope.

hugoShaka (commented):
LGTM, but the tests are going to break because some test cases have hardcoded expiry="2022-10-12", which was yesterday. Could you update them in your PR, so they will pass on master?

terraform/example/bot.tf.example
timothyb89: Machine ID supports a couple different token types. As you mentioned randomized tokens are problematic (though functional, barely) with Terraform. IAM and (as of v11) GitHub OIDC tokens are reusable and are not secret, so the long expiration time doesn't matter.

The plan when we document this is to strongly recommend only using reusable tokens with Terraform.

timothyb89:
> LGTM, but the tests are going to break because some test cases have hardcoded expiry="2022-10-12", which was yesterday. Could you update them in your PR, so they will pass on master?

I believe the tests use a mock clock so they should continue to work - it's the same expiry as all the other test fixtures so everything should be broken otherwise.

r0mant:
@timothyb89 @hugoShaka The tests were failing because of expiration date but I fixed them as a part of https://github.com/gravitational/teleport-plugins/pull/677.

timothyb89:
> @timothyb89 @hugoShaka The tests were failing because of expiration date but I fixed them as a part of #677. So you just need to rebase off of master.

Excuse me while I go eat my words :upside_down_face: Will fix!

hugoShaka (approved):

zmb3 (approved):
Bot.

timothyb89:
> I'm late to the PR's discussion, sorry
> 
> The file `terraform/provider/resource_teleport_bot.go` should be auto generated by the template The same happens today for all the other resources.
> 
> I'm not sure this was discussed, sorry if it was.

Unfortunately that's not really an option with bots, they don't cleanly map to any single server-side resource and have special (if simple) creation/deletion logic. We can always revert if you think this is a blocking issue but I don't see a Machine ID resource happening that isn't handwritten, unfortunately.

marcoandredinis:
Thank you :pray: 

We knew that auto generation would not work for all cases
We have a lib upgrade in the plugin framework that brings breaking changes
I'll try to understand if it makes sense to create a specific template for this resource (we might have other similar resources in the future, not sure)

Thank you

PR #672, opened 2022-09-29
Fail early if the credentials are expired

### LINKED ISSUE #628: Plugins should check for credential expiration at startup
Currently, if a plugin starts with a credential that is expired, it will proceed to attempt to connect using that credential.

At the very least, this should be a warn-then-continue, or better yet-- refuse to start with an expired credential.

https://github.com/gravitational/teleport-plugins/pull/601 - this did give more stacktrace output when it ran into a problem, but it still doesn't have much advice about looking at the credential expiration to resolve the EOF.

All the plugins should do this type of sanity check after loading the credential specified by the config file, but before attempting to connect.

hugoShaka:
Fixes #628

This PR introduces a credential check on plugin startup. Failing fast with a meaningful error message is a better user experience than hanging for several seconds and returning cryptic errors.


Joerger:
Can we actually add this to `teleport/api` instead? We definitely want to catch more errors like this for all API usages.

hugoShaka:
I agree. I will check with @r0mant tomorrow to see if it's OK to do this kinda-breaking change in `teleport/api` in v11.

lib/credentials/credentials.go
```diff
@@ -0,0 +1,65 @@
...
+					),
+				)
+			}
+		}
+		if isValid {
+			validCredentials = append(validCredentials, credential)
+		}
+	}
+
+	return len(validCredentials) != 0, trace.NewAggregate(errors...)
```
Joerger, line 64: nit: can we make `validCredentials` a boolean instead?

Joerger (approved):
I'm ok with merging this here for now (for v11), but it would be good to move this to `teleport/api` in the future.

hugoShaka:
Based on the comments here and a discussion with roman, it doesn't hurt to do the change now and will address client's need.

We will want to add the feature in `teleport/api`, this should be in a major release. An issue has been filed here to upstream this behavior: https://github.com/gravitational/teleport/issues/17139

lib/credentials/credentials.go
hugoShaka: done

r0mant (approved):

Issue #652, opened 2022-09-08
Check error returned by semaphore acquire


nklaassen:
We currently don't check the error returned by the semaphore acquire here https://github.com/gravitational/teleport-plugins/blob/bec9a9af07972eacdab605d1f6225bf41239cd53/event-handler/session_events_job.go#L74

This can cause a panic if the context is cancelled (likely because the event handler is already terminating for another reason).
```
INFO   Finished session events ingest id:04cfba49-6620-4180-8d44-bf5a72e7f4eb event-handler/session_events_job.go:181
panic: semaphore: released more than held

goroutine 8369 [running]:
golang.org/x/sync/semaphore.(*Weighted).Release(0xc000466190, 0x0?)
        /go/pkg/mod/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/semaphore/semaphore.go:103 +0xbd
main.(*SessionEventsJob).run.func2.1({0x12a2f50, 0xc00097c340})
        /go/src/github.com/gravitational/teleport-plugins/event-handler/session_events_job.go:116 +0x862
github.com/gravitational/teleport-plugins/lib.jobFunc.DoJob(0x0?, {0x12a2f50?, 0xc00097c340?})
        /go/src/github.com/gravitational/teleport-plugins/lib/process.go:195 +0x2c
github.com/gravitational/teleport-plugins/lib.NewProcess.func2.1()
        /go/src/github.com/gravitational/teleport-plugins/lib/process.go:83 +0x51
created by github.com/gravitational/teleport-plugins/lib.NewProcess.func2
        /go/src/github.com/gravitational/teleport-plugins/lib/process.go:82 +0x197 
```

We should check and handle the error, and add a lint rule to make sure we aren't silently ignoring any other errors

PR #681, opened 2022-10-20
Check err when aquiring semaphore

EdwardDowling:
closes [652](https://github.com/gravitational/teleport-plugins/issues/652)


r0mant (commented):
@EdwardDowling Can you add a linter rule to our `make lint` target that would catch errors like this? I think there's `errcheck` linter plugin that we may need to enable.

tcsc (dismissed):
Looks good, but I agree that there should be a lint rule to try and catch this statically.

r0mant (approved):

Joerger (approved):

PR #682, opened 2022-10-24
Bump Teleport API to v11

marcoandredinis:
This commit bumps de Teleport API to v11 rc2, preparing for the v11 release.

Go was updated to 1.19.2

Terraform schema was re-generated according to the latest api changes. Because of the above, some tests were removed.

**Breaking Changes**
Terraform: OIDC Connectors - redirect url
We moved from a single Redirect URL to multiple Redirect URLs If you were using the prop `redirect_url` you need to change it to a single element array:
```tf
redirect_url = "https://example.com"
```
Should now be:
```tf
redirect_url = ["https://example.com"]
```

If you were already using the array property - `redirect_urls` You must change it as well:

```tf
redirect_urls = ["https://example.com"]
```
Should now be (loses the `s` in `urls`)
```tf
redirect_url = ["https://example.com"]
```


terraform/protoc-gen-terraform-teleport.yaml
```diff
@@ -135,25 +135,29 @@ exclude_fields:
...
     # Going forward we want to:
-    # - change the type of `redirect_url` from string into list of strings to match the resource's yaml representation in Teleport
-    # - after doing the above, remove `redirect_urls` field
+    # - change the type of `require_session_mfa` from boolean into an enum to match the resource's yaml representation in Teleport
+    # - after doing the above, remove `require_mfa_type` field
     
-    # When we change the type of `redirect_url` from string to list of strings we break user's configurations
+    # When we change the type of `require_session_mfa` from boolean to enum we break user's configurations
     # They'll need to change
-    # redirect_url: "https://example.org/redirect"
```
r0mant, line 150: @Joerger Did we remove support for singular `redirect_url` field in Teleport 11?

r0mant (commented):
@Joerger Can you please review this PR?

terraform/protoc-gen-terraform-teleport.yaml
Joerger: Teleport 11 Still supports `redirect_url` for JSON marshalling, aka for dynamic resources and file conf. Only Terraform is impacted by this change since it does not use JSON marshaling and therefore does not respect the `github.com/gravitational/teleport/api/types/wrappers.Strings` type.

Here's the recent commit that deprecates the `RedirectURL` proto field - [cc5931d1014915f641b232956901df6b9cbbe08a](https://github.com/gravitational/teleport/pull/17047/commits/2a12068cd01b79f149b1731fcf75d6827ee41a3f)

Joerger (approved):

r0mant (approved):
@marcoandredinis Do not merge yet please, we'll merge it after Teleport 11 release.

PR #686, opened 2022-10-31
Update go.mod's version to 1.19

marcoandredinis:


zmb3:
Are you actually using Go 1.19 features in the codebase yet? If not, there's no need to do this (in fact, we _shouldn't_ do this as it creates additional unnecessary barriers).

marcoandredinis:
No 1.19 features are being used.

Closing this

My goal was to have the same version being used both in `teleport` and `teleport-plugins`

PR #690, opened 2022-11-02
Terraform: remove requirement of `teams_to_logins` field in Github connector

### LINKED ISSUE #687: Inappropriate value for attribute "spec": attribute "teams_to_logins" is required.
Starting with Teleport 12, we'll depreate GitHub connector's teams_to_logins field https://github.com/gravitational/teleport/discussions/17951

I tried to change this with the 11.0.1 Terraform Provider but ened out with this message. team_to_roles is support in Teleport 11, but we should update the Terraform provider to support both fields during the transition period. 

```
│ Error: Incorrect attribute value type
│
│   on teleport.tf line 146, in resource "teleport_github_connector" "github":
│  146:   spec = {
│  147:     client_id     = var.gh_ae_client_id
│  148:     client_secret = var.gh_ae_client_secret
│  149:     display       = "github"
│  150:     redirect_url  = "${var.teleport_url}:443/v1/webapi/github/callback"
│  151:     teams_to_roles = [{
│  152:       roles       = ["access", "aws-console-access-iam"]
│  153:       team         = var.gh_team
│  154:       organization = var.gh_org
│  155:     }]
│  156:   }
│     ├────────────────
│     │ var.gh_ae_client_id has a sensitive value
│     │ var.gh_ae_client_secret has a sensitive value
│     │ var.gh_org is a string, known only after apply
│     │ var.gh_team is a string, known only after apply
│     │ var.teleport_url is a string, known only after apply
│
│ Inappropriate value for attribute "spec": attribute "teams_to_logins" is required.
```

marcoandredinis:
Teleport is moving from `teams_to_logins` to `teams_to_roles`. As of V10, both are configurable but Terraform was requiring the deprecated one(`teams_to_logins`).

This commit allows for both to be set and paves the way for the removal of the property `teams_to_logins` when we release v12.

Fixes #687

### DIFF
--- a/terraform/example/github_connector.tf.example
+++ b/terraform/example/github_connector.tf.example
@@ -19,10 +19,10 @@ resource "teleport_github_connector" "github" {
     client_id = "client"
     client_secret = var.github_secret
 
-    teams_to_logins = [{
+    teams_to_roles = [{
        organization = "gravitational"
        team = "devs"
-       logins = ["example"]
+       roles = ["example"]
     }]
   }
 }
--- a/terraform/protoc-gen-terraform-teleport.yaml
+++ b/terraform/protoc-gen-terraform-teleport.yaml
@@ -254,9 +254,6 @@ required_fields:
     - "GithubConnectorV3.Spec"
     - "GithubConnectorV3.Spec.ClientID"
     - "GithubConnectorV3.Spec.ClientSecret"
-    - "GithubConnectorV3.Spec.TeamsToLogins"
-    - "GithubConnectorV3.Spec.TeamsToLogins.Team"
-    - "GithubConnectorV3.Spec.TeamsToLogins.Logins"
     - "GithubConnectorV3.Metadata.Name"
 
     # OIDC connector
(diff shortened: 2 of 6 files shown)


zmb3 (approved):
Looks good to me, though now that both fields are optional, what happens if the user fails to specify either of `teams_to_logins` or `teams_to_roles`? Do we just let the auth server catch the error at that point, or is there a way to mark one of these required in the schema?

marcoandredinis:
> Looks good to me, though now that both fields are optional, what happens if the user fails to specify either of `teams_to_logins` or `teams_to_roles`? Do we just let the auth server catch the error at that point, 

We let the auth server catch the error
It returns something like this to the end user
`team_to_logins or team_to_roles mapping is invalid, no mappings defined`

> or is there a way to mark one of these required in the schema?

I think it's possible but would require us to change the proto file and explicitly add the `OneOf` directive to the fields
https://github.com/gravitational/teleport/blob/698e377fb412824d48d908a5f8229bc20586e2ca/api/proto/teleport/legacy/types/types.proto#L3482-L3496

zmb3:
Ok, no problem. This is good as-is since this is only a temporary situation anyway.

Joerger (approved):

PR #692, opened 2022-11-03
access/msteams: Fix pluginData encoding

hugoShaka:
Fixes https://github.com/gravitational/teleport/issues/18068

Encoding/decoding of messages in pluginData was breaking if the data contained `/` or `,`. This was the case in the issue above as the customer used the teams URL channel.

To fix this I switched to a more robust encoding: each message is json-encoded and `messages` is a list of the base64-encoded mesasge. This will increase the storage size needed for the plugin data by approx 33%. I don't think this will be an issue.

The commit also contains the following changes:
- fix what I think was a bug in the plugin data update function in `access/msteams/app.go`
- do not swallow errors when retrying in the `cas`


access/msteams/msapi/token.go
```diff
@@ -45,7 +45,7 @@ func (c *tokenWithTTL) Bearer(ctx context.Context, config Config) (string, error
 	expiresAt := c.expiresAt
 	c.mu.RUnlock()
 
-	if expiresAt == 0 || expiresAt > time.Now().UnixNano() {
+	if expiresAt == 0 || expiresAt < time.Now().UnixNano() {
```
marcoandredinis, line 48: :see_no_evil:

access/msteams/plugindata.go
```diff
@@ -26,14 +27,22 @@ func DecodePluginData(dataMap map[string]string) PluginData {
...
+				parts := strings.Split(encodedMsg, "/")
+				if len(parts) == 3 {
+					data.TeamsData = append(data.TeamsData, TeamsMessage{ID: parts[0], Timestamp: parts[1], RecipientID: parts[2]})
+				}
+			} else {
+				msg := &TeamsMessage{}
+				err = json.Unmarshal(decodedMsg, msg)
+				if err != nil {
+					continue
+				}
```
marcoandredinis, line 44: I think we should add at least a log here :thinking:

access/msteams/plugindata.go
```diff
@@ -26,14 +27,22 @@ func DecodePluginData(dataMap map[string]string) PluginData {
...
-			parts := strings.Split(encodedMsg, "/")
-			if len(parts) == 3 {
-				data.TeamsData = append(data.TeamsData, TeamsMessage{ID: parts[0], Timestamp: parts[1], RecipientID: parts[2]})
+			decodedMsg, err := base64.StdEncoding.DecodeString(encodedMsg)
+			if err != nil {
+				// Backward compatibility
+				parts := strings.Split(encodedMsg, "/")
+				if len(parts) == 3 {
+					data.TeamsData = append(data.TeamsData, TeamsMessage{ID: parts[0], Timestamp: parts[1], RecipientID: parts[2]})
+				}
```
marcoandredinis, line 38: We should remove this path later on right?

Can we add a TODO here? Maybe to remove this when we release v12 ?

access/msteams/plugindata.go
```diff
@@ -47,7 +56,12 @@ func EncodePluginData(data PluginData) map[string]string {
 
 	var encodedMessages []string
 	for _, msg := range data.TeamsData {
-		encodedMessages = append(encodedMessages, fmt.Sprintf("%s/%s/%s", msg.ID, msg.Timestamp, msg.RecipientID))
+		jsonMessage, err := json.Marshal(msg)
+		if err != nil {
+			continue
+		}
```
marcoandredinis, line 62: We discard this error
Can we at least add a log?

hugoShaka:
@marcoandredinis I changed the error handling to hard fail as soon as we can't serialize/deserialize plugin data. My first try was to unmarshall as much as possible and ignore failures, but this will only need to infinite retry loops when doing the compare and swap.

r0mant (approved):

access/msteams/msapi/token.go
marcoandredinis, line 48: I think we should add an offset here
If the token expires in the next `retry count * timeout seconds` then renew the token.

As an alternative, we might change `access/msteams/msapi/client.go` to (possibly) renew the token when we receive a 401.

So, close to this `if` statement
```go
		if resp.StatusCode > 499 || resp.StatusCode == http.StatusTooManyRequests {
```

access/msteams/plugindata.go
```diff
@@ -15,42 +18,58 @@ type PluginData struct {
...
+					data.TeamsData = append(data.TeamsData, TeamsMessage{ID: parts[0], Timestamp: parts[1], RecipientID: parts[2]})
+				}
+			} else {
+				msg := &TeamsMessage{}
+				err = json.Unmarshal(decodedMsg, msg)
+				if err != nil {
+					errors = append(errors, err)
+				}
+				data.TeamsData = append(data.TeamsData, *msg)
 			}
```
marcoandredinis, line 50: :nail_care: 
Swap the `else` with a `continue` in the `if` block
This reduces the indentation level which helps readability (a subjective term :sweat_smile: )

```suggestion
				continue
			}
			
			msg := &TeamsMessage{}
			err = json.Unmarshal(decodedMsg, msg)
			if err != nil {
				errors = append(errors, err)
			}
			data.TeamsData = append(data.TeamsData, *msg)
```

lib/plugindata/cas.go
```diff
@@ -109,33 +111,40 @@ func (c *CompareAndSwap[T]) Update(
...
+				failedAttempts = append(failedAttempts, trace.Wrap(backoffErr))
+				return emptyData, trace.NewAggregate(failedAttempts...)
 			}
 
 			continue
 		} else if err != nil {
 			return emptyData, trace.Wrap(err)
 		}
 
+		// SUbmit modifications
```
marcoandredinis, line 129:
```suggestion
		// Submit modifications
```

access/msteams/msapi/token.go
marcoandredinis: :+1:

hugoShaka: Well, I was wrong and we renew the token 1 minute before expiry. This is conservative and should not cause us any time skew or slow call issue :)

PR #694, opened 2022-11-07
Terraform: allow identity to be provided as base64

### LINKED ISSUE #693: Terraform Provider should be able to communicate with a web proxy without files
When running terraform in some automated systems, such as terraform cloud, it is not always possible to load files for credentials. Since the addr can point to a teleport auth service or to a teleport proxy service, the terraform provider attempts to connect in different ways.

Given the constraint of no files to load, and an addr that is a teleport web proxy endpoint, we run into a problem.

Currently, the terraform provider provides the following combinations to configure it with credentials:

* `cert_path`, `key_path`, `root_ca_path`, and `addr` to connect using key files.
  Requires file ❌, treats addr like auth service ❌

* `cert_base64`, `key_base64`, `root_ca_base64`, and `addr` to connect using a base64-encoded key.
  doesn't require file ✅, treats addr like an auth service ❌

* `identity_file_path` and `addr` to connect using identity file.
  requires file ❌, treats addr like a web proxy ✅

* `profile_name` and `profile_dir` (both can be empty) and Teleport will try to connect using current profile from ~/.tsh
  requires file ❌, can treat addr like a web proxy ✅

It should be possible to force the terraform provider to connect with credentials not populated from a file, and also to connect to a web proxy rather than an auth endpoint. It seems like the easiest way to achieve this might be to add `identity_file_base64`

Currently, if one tries to use the base64 method to configure the provider, it fails with the following output:

```
Terraform v1.3.4
on linux_amd64
Initializing plugins and modules...
╷
│ Error: Error connecting to Teleport!
│ 
│   on main.tf line 22, in provider "teleport":
│   22: provider "teleport" {
│ 
│ all connection methods failed
│ 	
│ 	profile could not be decoded: current-profile is not set, 
│ 	failed to connect to addr teleport.example.com:443 as an auth server
│ 		context deadline exceeded: connection error: desc = "transport:
│ authentication handshake failed: x509: certificate is valid for
│ teleport.example.com, not teleport.cluster.local"
╵
Operation failed: failed running terraform plan (exit 1)
```

A packet capture on the proxy side confirms that the terraform provider never attempts to connect to the web proxy as a web proxy. it instead always sets the SNI to teleport.cluster.local.

```
provider "teleport" {
  addr        = "teleport.redacted.net443"
  cert_base64 = "redacted"
  key_base64 = "redacted"
  root_ca_base64 = "redacted"
}
```

marcoandredinis:
When configuring the Teleport Provider we have multiple methods. One of them is to provide the Identity as a file path.

With this PR we can now provide the identity as a base64 encoded string.

This PR also adds a `dial_timeout_duration` configuration for the provider to ensure we can set our own limit instead of always using 30s.

Fixes #693

### DIFF
--- a/terraform/provider/provider.go
+++ b/terraform/provider/provider.go
@@ -82,12 +82,16 @@ type providerData struct {
 	IdentityFilePath types.String `tfsdk:"identity_file_path"`
 	// IdentityFile identity file content
 	IdentityFile types.String `tfsdk:"identity_file"`
+	// IdentityFile identity file content encoded in base64
+	IdentityFileBase64 types.String `tfsdk:"identity_file_base64"`
 	// RetryBaseDuration is used to setup the retry algorithm when the API returns 'not found'
 	RetryBaseDuration types.String `tfsdk:"retry_base_duration"`
 	// RetryCapDuration is used to setup the retry algorithm when the API returns 'not found'
 	RetryCapDuration types.String `tfsdk:"retry_cap_duration"`
 	// RetryMaxTries sets the max number of tries when retrying
 	RetryMaxTries types.String `tfsdk:"retry_max_tries"`
+	// DialTimeout sets timeout when trying to connect to the server.
+	DialTimeoutDuration types.String `tfsdk:"dial_timeout_duration"`
 }
 
 // New returns an empty provider struct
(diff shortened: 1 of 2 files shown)


dangreenisrael:
May I humbly request we rename the PR from `bas64` to `base64`. It's not that I'm OCD, it's just that this might give me an aneurysm

terraform/provider/provider.go
```diff
@@ -267,12 +285,36 @@ func (p *Provider) Configure(ctx context.Context, req tfsdk.ConfigureProviderReq
 		creds = append(creds, client.LoadIdentityFileFromString(identityFile))
 	}
 
+	if identityFileBase64 != "" {
```
r0mant, line 288: Nit: I'd use strings.TrimSpace here just in case.

r0mant (approved):

PR #695, opened 2022-11-09
Introduced GitHub actions workflows

gzigzigzeo:
What's inside:

- [x] GitHub workflows for unit tests.
- [x] GitHub workflows for golang lint.
- [x] Helm charts unit tests.
- [x] Fixes for linting errors.
- [x] Terraform tests.
- [x] GitHub workflows for integration tests.

Terraform test fails. Looks like my license is invalid for CI. 

`TELEPORT_ENTERPRISE_LICENSE` secret must be provided after this gets merged.

I found out that `Makefile` `make test` does not run either lib/* tests or specific plugins tests. I guess, this needs to be addressed in a separate PR.

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -1,0 +1,44 @@
+name: tests-integration
+run-name: Tests (integration)
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+env:
+  TELEPORT_GET_VERSION: v10.2.0
+  TELEPORT_ENTERPRISE_LICENSE: ${{ secrets.TELEPORT_ENTERPRISE_LICENSE }}
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+
+    strategy:
+      matrix:
+        plugin:
+          - access/discord
+          - access/email
+          - access/jira
+          - access/mattermost
+          - access/msteams
+          - access/pagerduty
+          - access/slack
+          - event-handler
+
+    steps:
+      - name: Checkout Teleport Plugins
+        uses: actions/checkout@v3
+
+      - name: Setup Go 1.19.3
+        uses: actions/setup-go@v3
+        with:
+          go-version: '1.19.3'
+
+      - name: Run tests (${{ matrix.plugin }})
+        run: |
+          make -C ${{ matrix.plugin }} test
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -31,5 +31,5 @@ jobs:
       - name: make test-tooling
         run: make test-tooling
 
-      - name: go test lib
-        run: cd lib && go test -v ./...
+      - name: make test-unit
+        run: make test-unit
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -1,0 +1,35 @@
+name: tests-unit
+run-name: Tests (unit)
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+env:
+  TELEPORT_GET_VERSION: v10.2.0
+  TELEPORT_ENTERPRISE_LICENSE: ${{ secrets.TELEPORT_ENTERPRISE_LICENSE }}
+
+jobs:
+  test:
+    name: Tests (unit)
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout Teleport Plugins
+        uses: actions/checkout@v3
+
+      - name: Setup Go 1.19.3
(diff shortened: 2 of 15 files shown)


.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+      - name: Setup Go 1.19.3
+        uses: actions/setup-go@v3
+        with:
+          go-version: '1.19.3'
+
+      - name: make test-tooling
+        run: make test-tooling
+
+      - name: go test lib
```
gozer, line 31: No make target to cover this ?

gozer: > I found out that Makefile make test does not run either lib/* tests or specific plugins tests. I guess, this needs to be addressed in a separate PR.

Just noticed that, why make it a separate PR, I'd lump it in here to make sure it doesn't get forgotten

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+env:
+  TELEPORT_GET_VERSION: v10.2.0
+  TELEPORT_ENTERPRISE_LICENSE: ${{ secrets.TELEPORT_ENTERPRISE_LICENSE }}
```
gozer, line 12: Doesn't seem used/needed here.

gozer:
> `TELEPORT_ENTERPRISE_LICENSE` secret must be provided after this gets merged.

From what I can see, `TELEPORT_ENTERPRISE_LICENSE` is supposed to be a path to a license file, not the license content iself.

.github/workflows/integration-tests.yaml
tcsc: @gozer - unless things have changed, the Teleport plugin integration tests actually run full, compiled versions of Teleport to test against (as opposed to the Teleport integration tests, which essentially use Teleport as a library). 

`TELEPORT_ENTERPRISE_LICENSE` will be picked up by the test setup code and written to a file for the Teleport instances under test to find.

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,40 @@
...
+    steps:
+      - name: Checkout Teleport Plugins
+        uses: actions/checkout@v3
+
+      - name: Setup Go 1.19.3
+        uses: actions/setup-go@v3
+        with:
+          go-version: '1.19.3'
+
+      - name: Run tests (${{ matrix.plugin }})
```
tcsc, line 38: Cute.

.github/workflows/integration-tests.yaml
tcsc: Also, `TELEPORT_ENTERPRISE_LICENSE` is given an exemption from our _no secrets in public CI_ policy, as it's not really a secret, it's just a convenient way to inject a license into the build. 

(This license is well known as a debug/test license so if found in the wild we know it's not valid.)

tcsc: ... although, on second thoughts, it might be an idea to talk with whoever is currently working on license enforcement and make sure that these assumptions are still valid.

camscale: I can't (yet) see anything that would change here with license enforcement. License enforcement will disable the enterprise features when the license has expired (with a grace period) and should not affect the way the license is provided.

.github/workflows/unit-tests.yaml
gzigzigzeo: Okay. Introduced test-unit target.

gzigzigzeo:
License taken from teleport.e fixed the failing tests.

gozer (approved):

r0mant (approved):
lgtm! @tcsc @gozer Any other comments?

@marcoandredinis I think would be useful if you took a look as well as we're migrating teleport-plugins tests to Github Actions.

Makefile
```diff
@@ -114,14 +114,18 @@ event-handler:
...
 	@echo Testing plugins against Teleport $(TELEPORT_GET_VERSION)
 	go test -race -count 1 $(shell go list ./... | grep -v -e '/terraform/' -e '/tooling/')
 
 .PHONY: test-tooling
 test-tooling:
 	(cd tooling; go test -v -race ./...)
 
+.PHONY: test-unit
+test-unit:
+	(cd lib; go test -v -race ./...)
```
marcoandredinis, line 127: :nail_care: 
We might be testing twice the `lib/`

We can change the `go test` of the `test` target to:
```shell
	go test -race -count 1 $(shell go list ./... | grep -v -e '/terraform/' -e '/tooling/' -e '/lib/')
```

marcoandredinis (approved):

r0mant:
This was merged in https://github.com/gravitational/teleport-plugins/pull/705.

PR #697, opened 2022-11-10
Add discord to release pipeline

hugoShaka:
Part of https://github.com/gravitational/teleport/issues/9563

This PR adds discord access plugin to the release pipeline.

~Blocked by https://github.com/gravitational/cloud-terraform/pull/1112~


Joerger (approved):

access/discord/Makefile
```diff
@@ -44,7 +44,6 @@ release: clean $(BINARY)
 	mkdir $(RELEASE_NAME)
 	cp -rf $(BINARY) \
 		README.md \
-		CHANGELOG.md \
```
r0mant, line 47: Any reason we do not want to include changelog with discord like with other plugins?

r0mant (approved):

access/discord/Makefile
hugoShaka: It's empty on all the plugins I checked, the versions are wrong and it's not maintained as it's not part of our release process. I think we should remove them all.

PR #729, opened 2023-01-03
Extract slack access plugin to own package

justinas:
Contributes to https://github.com/gravitational/cloud/issues/2506 , see also RFD in https://github.com/gravitational/cloud/pull/2678.

Moves most of the code to a non-main package, in order to be able to depend on this code in `teleport.e`.


tcsc (approved):
I'm not sure about having the plugin cmd live _inside_ the `slack` package.

Would it make sense to have a structure where the common slack library went into `./pkg/slack` and `./access/slack` stayed as the slack plugin entry point? That seems to be a common pattern for [exposing library code](https://github.com/golang-standards/project-layout#pkg).

justinas:
> I'm not sure about having the plugin cmd live _inside_ the `slack` package.
> 
> Would it make sense to have a structure where the common slack library went into `./pkg/slack` and `./access/slack` stayed as the slack plugin entry point? That seems to be a common pattern for [exposing library code](https://github.com/golang-standards/project-layout#pkg).

Just to note, golang-standards is not [without controversy](https://github.com/golang-standards/project-layout/issues/117) :)

We seem to have at least a couple of slightly different patterns:

* teleport:

  ```
  api/
    ...
  lib/
    ...
  tool/
    tbot/
    tctl/
    teleport/
  ```

* cloud:

  ```
  api/
    ...
  pkg/
    ...
  tools/
    salescenter/
    tenantoperator/
  ```

I don't care much either way as long as the goal is achieved, so I'll defer this judgement to someone with a stronger opinion. :)

justinas:
I've noticed the naming lints:

```
Error: type name will be used as slack.SlackBot by other packages, and that stutters; consider calling this Bot (golint)
  Error: type name will be used as slack.SlackConfig by other packages, and that stutters; consider calling this Config (golint)
  Error: type name will be used as slack.SlackResponse by other packages, and that stutters; consider calling this Response (golint)
  Error: type name will be used as slack.SlackMsg by other packages, and that stutters; consider calling this Msg (golint)
  Error: type name will be used as slack.SlackUser by other packages, and that stutters; consider calling this User (golint)
```

Will proceed with the following renames if they seem good. There should only be internal uses of those, as it is not an importable package.

Joerger (approved):

PR #738, opened 2023-01-13
Add foundation for short-lived OAuth token support to Slack plugin

justinas:
Contributes to https://github.com/gravitational/cloud/issues/2506, see also RFD https://github.com/gravitational/cloud/pull/2678 .

This adds interfaces for the most common OAuth2 flows (authorization code, refresh token), and a single concrete implementation for the Slack API. This allows using [token rotation](https://api.slack.com/authentication/rotation) as opposed to a fixed authentication token.

We plan to use this when running plugins in Cloud. The standalone self-hosted plugin itself is unchanged here, it still uses the old way (static auth token). However the design of these changes should suit the self-hosted use case as well with additional changes: an implementation of the initial OAuth flow and some config boilerplate.

## Testing

* Unit tests for new components
* Manual testing of the dynamic tokens (exchange, refresh) with an ad-hoc Go script
* Manual testing of the standalone Slack plugin to ensure it still works as expected with a static token


access/common/auth/token_provider_test.go
```diff
@@ -0,0 +1,169 @@
...
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/gravitational/teleport-plugins/access/common/auth/state"
+	"github.com/gravitational/trace"
+	"github.com/jonboulle/clockwork"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/suite"
```
Joerger, line 12: Is there a reason why you chose to use suite style tests here? I believe we are still trying to migrate our tests to simpler `testify/require` assertions as opposed to suites, using built in go test buildup/breakdown mechanisms instead - https://github.com/gravitational/teleport/blob/master/rfd/0001-testing-guidelines.md#details

Let me know if this RFD has fallen out of date and we have started using suite style tests again.

access/slack/oauth.go
```diff
@@ -0,0 +1,88 @@
...
+func newAuthorizer(client *resty.Client, clientID string, clientSecret string) *Authorizer {
+	return &Authorizer{
+		client:       client,
+		clientID:     clientID,
+		clientSecret: clientSecret,
+	}
+}
+
+// NewAuthorizer returns a new Authorizer
+func NewAuthorizer(clientID string, clientSecret string) *Authorizer {
```
Joerger, line 30: What are `clientID` and `clientSecret`? Can you add the description to the godoc?

access/slack/config.go
```diff
@@ -34,7 +34,8 @@ import (
 // Config stores the full configuration for the teleport-slack plugin to run.
 type Config struct {
 	common.BaseConfig
-	Slack common.GenericAPIConfig
+	Slack               common.GenericAPIConfig
+	AccessTokenProvider auth.AccessTokenProvider
```
Joerger, line 38: Currently `Config` can be parsed from a `toml` file for self hosted plugins, can we modify this so that `AccessTokenProvider` can also be parsed from toml?

```
type Config struct {
	Slack SlackConfig
}

type SlackConfig struct {
	common.GenericAPIConfig
	ClientID     string
	ClientSecret string
}
```

If these new config options will be used for other plugins as well as a common OAuth2 flow, we should add those files to `common.GenericAPIConfig` instead.

In addition to making this usable to self hosted, this removes the need for the cloud use case to call `NewAuthorizer` itself, so we can remove that exported function.

We should also update the example toml file to include this feature option.

```
# Example slack plugin configuration TOML file

[teleport]
# Teleport Auth/Proxy Server address.
# addr = "example.com:3025"
#
# Should be port 3025 for Auth Server and 3080 or 443 for Proxy.
# For Teleport Cloud, should be in the form "your-account.teleport.sh:443".

# Credentials generated with `tctl auth sign`.
#
# When using --format=file:
# identity = "/var/lib/teleport/plugins/slack/auth_id"    # Identity file
#
# When using --format=tls:
# client_key = "/var/lib/teleport/plugins/slack/auth.key" # Teleport TLS secret key
# client_crt = "/var/lib/teleport/plugins/slack/auth.crt" # Teleport TLS certificate
# root_cas = "/var/lib/teleport/plugins/slack/auth.cas"   # Teleport CA certs

[slack]
# Slack Bot OAuth token
token = "xoxb-11xx"

# description
client_id = "example"
client_secret = "example"

[role_to_recipients]
# Map roles to recipients.
#
# Provide slack user_email/channel recipients for access requests for specific roles. 
# role.suggested_reviewers will automatically be treated as additional email recipients.
# "*" must be provided to match non-specified roles.
#
# "dev" = "devs-slack-channel"
# "*" = ["admin@email.com", "admin-slack-channel"]

[log]
output = "stderr" # Logger output. Could be "stdout", "stderr" or "/var/lib/teleport/slack.log"
```

access/common/auth/token_provider.go
```diff
@@ -0,0 +1,221 @@
...
+				r.creds = creds
+				r.lock.Unlock()
+
+				interval := r.getRefreshInterval(creds)
+				timer.Reset(interval)
+				r.log.Infof("Next refresh in: %s", interval)
+				continue
+			}
+
+			creds, err := r.refresh(r.ctx)
```
Joerger, line 175: It looks like there is a race condition where two HA instances can refresh the credentials. The logic above doesn't quite prevent this, as both instances could have called `GetCredentials` before either of them refresh the credentials. Will this cause the first instance that refreshed to have its credentials overwritten by the other instance refresh, or is this a race condition that we are ok with ignoring?

access/common/auth/token_provider.go
```diff
@@ -0,0 +1,221 @@
...
+	defer timer.Stop()
+	r.log.Infof("Will attempt token refresh in: %s", interval)
+
+	for {
+		select {
+		case <-r.ctx.Done():
+			r.log.Info("Shutting down")
+			return
+		case <-timer.Chan():
+			creds, _ := r.state.GetCredentials(r.ctx)
```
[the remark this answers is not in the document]
justinas: We could log the error here.

I think I omitted the check since it checks `creds != nil` further along the way. This is just an optimistic check whether the creds are already fresh enough, and in case of an error we will still proceed as usual and try to refresh + save new creds.

access/common/auth/token_provider_test.go
justinas: Ah, I mostly structured this after existing tests. Will move to vanilla `testing.T` :+1:

access/common/auth/token_provider.go
justinas: A couple of possible issues with this simultaneous refresh/overwrite of credentials come to mind:
* Validity of refresh token - from my testing, these are not "one-time-use", you can use the same refresh token multiple times (in its validity period). This is probably by design: if one *used up* the refresh token, but failed to save the new credentials, the app would be broken for good and the user would need to re-initiate the authorization. :heavy_check_mark: 
* Validity of renewed credentials, i.e. whether refresh via the same "refresh token" invalidates previous credentials. Will need to test this one :question:

access/slack/config.go
justinas: We weren't exactly planning to extend this "proper" OAuth2 support to self-hosted version just yet. This implementation is mostly to scratch the Cloud's itch :slightly_smiling_face: 

One problem is that the Slack plugin itself does not implement the initial "exchange" flow (this is planned to be implemented by Proxy/Auth in Cloud). We would need to add a HTTP server to receive the auth callback (in a similar vein to `tsh` browser login flow).

In theory, we could support this, but for now the users would need to write their own shim to do the initial auth/exchange, then configure client-id/client-secret/token. Not sure if it's worth half-baking such solution into self-hosted version. /cc @jimbishopp

access/slack/oauth_test.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func TestOAuth(t *testing.T) {
+	const clientID = "my-client-id"
+	const clientSecret = "my-client-secret"
+	const authorizationCode = "12345678"
+	const redirectURI = "https://foobar.com/callback"
+	const refreshToken = "my-refresh-token1"
+
+	const expiresInSeconds = 43200
```
Joerger, line 82: nit:
```suggestion
	const (
		clientID = "my-client-id"
		clientSecret = "my-client-secret"
		authorizationCode = "12345678"
		redirectURI = "https://foobar.com/callback"
		refreshToken = "my-refresh-token1"
		expiresInSeconds = 43200
	)
	
```

access/common/auth/token_provider.go
Joerger: Makes sense, can we add a comment so this doesn't get confused for a race condition down the road? Just looking at the code and comment above, I didn't realize the `shouldRefresh` logic is just an optimistic check, rather than core logic.

Joerger (approved):

access/common/auth/state/state.go
```diff
@@ -0,0 +1,17 @@
+package state
+
+import (
+	"context"
+	"time"
+)
+
+type Credentials struct {
```
marcoandredinis, line 8: Can you please add some comments to the Struct and interface below?

access/common/auth/token_provider.go
```diff
@@ -0,0 +1,217 @@
...
+
+	r.creds, err = r.state.GetCredentials(ctx)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	return nil
+}
+
+// GetAccessToken implements AccessTokenProvider()
+func (r *RotatedAccessTokenProvider) GetAccessToken() (string, error) {
```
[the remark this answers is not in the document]
justinas: The problem with "refresh-on-demand" is that `GetAccessToken()` might not get called for a very long time. The Slack plugin does not actively call any Slack API endpoints as some sort of "keep-alive". So if no access requests come in for X hours* and thus we don't get a chance to refresh the credentials, the refresh token might expire.

\* I could not find a definitive answer on how long the refresh tokens issued by Slack are valid for.

access/common/auth/token_provider.go
```diff
@@ -0,0 +1,217 @@
...
+func (s *StaticAccessTokenProvider) GetAccessToken() (string, error) {
+	return s.token, nil
+}
+
+// RotatedAccessTokenProviderConfig contains parameters and dependencies for RotatedAccessTokenProvider
+type RotatedAccessTokenProviderConfig struct {
+	RetryInterval       time.Duration
+	TokenBufferInterval time.Duration
+
+	State     state.State
```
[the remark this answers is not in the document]
marcoandredinis: Thank you :+1:

access/common/auth/token_provider.go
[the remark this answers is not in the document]
marcoandredinis: Got it :+1: 
Thank you :+1:

marcoandredinis (approved):
LGTM
Is there a way for me to test this end to end?

justinas:
> LGTM Is there a way for me to test this end to end?

I don't have a self-contained example with `RotatedTokenProvider` currently, but I tested exchange & refresh using Authorizer via such shim:

<details>
<summary>Go code</summary>

```go
package main

import (
    "context"
    "fmt"
    "log"
    "net/http"
    "strings"

    "github.com/gravitational/teleport-plugins/access/common/auth/oauth"
    "github.com/gravitational/teleport-plugins/access/common/auth/storage"
    "github.com/gravitational/teleport-plugins/access/slack"
    "github.com/sirupsen/logrus"
)

const clientID = "<...>"
const clientSecret = "<...>"

var scopes = []string{
    "chat:write",
    "users:read",
    "users:read.email",
}

const callbackURL = "https://localhost:5000/callback?tenant=mytenant"

var authorizer oauth.Authorizer

func printCreds(pre string, creds *storage.Credentials) {
    log.Printf("%s\naccess: %s\nrefresh: %s\nexpire: %s", pre, creds.AccessToken, creds.RefreshToken, creds.ExpiresAt)
}

func index(w http.ResponseWriter, r *http.Request) {
    authURL := fmt.Sprintf("https://slack.com/oauth/v2/authorize?scope=%s&client_id=%s&redirect_uri=%s", strings.Join(scopes, ","), clientID, callbackURL)
    http.Redirect(w, r, authURL, http.StatusFound)
}

func callback(w http.ResponseWriter, r *http.Request) {
    q := r.URL.Query()
    if err := q.Get("error"); err != "" {
        log.Println(err)
        return
    }

    log.Printf("Initial creds:\ncode: %s\nredirect URL:%s\n", q.Get("code"), callbackURL)

    creds, err := authorizer.Exchange(context.Background(), q.Get("code"), callbackURL)
    if err != nil {
        log.Println(err)
        return
    }
    printCreds("Successfully exchanged creds:", creds)

    creds, err = authorizer.Refresh(context.Background(), creds.RefreshToken)
    if err != nil {
        log.Println(err)
        return
    }
    printCreds("Successfully refreshed creds:", creds)
}

func main() {
    logrus.SetLevel(logrus.DebugLevel)
    authorizer = slack.NewAuthorizer(clientID, clientSecret)

    log.Println("Listening at https://127.0.0.1:5000")
    http.HandleFunc("/", index)
    http.HandleFunc("/callback", callback)
    err := http.ListenAndServeTLS(":5000", "cert.pem", "key.pem", nil)
    if err != nil {
        panic(err)
    }
}
```

</details>

You will also need a Slack app with [token rotation](https://api.slack.com/authentication/rotation) and allowed redirect URL of `https://localhost:5000/callback`, as well as self-signed TLS cert for the HTTP server.

PR #739, opened 2023-01-14
Patched()fix jsonwebtoken has insecure input validation in jwt.verify function

imhunterand:
## Overview
 jsonwebtoken library, if a malicious actor has the ability to modify the key retrieval parameter (referring to the secretOrPublicKey argument from the [readme link](https://github.com/auth0/node-jsonwebtoken#jwtverifytoken-secretorpublickey-options-callback)) of the `jwt.verify()` function, they can gain remote code execution (RCE). You are affected only if you allow untrusted entities to modify the key retrieval parameter of the `jwt.verify()` on a host that you control.

JWT Header
As shown in Figure 1, the header consists mostly of two parameters that indicate the type of the token and the signing algorithm.
```js
The header and payload are each Base64Url encoded to form the first and second parts of a JWT string (Header.Payload.Signature):

eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9 and eyJ1c2VyX25hbWUiOiJKb2huIERvZSIsImFkbWluIjp0cnVlfQ.

The third part of the JWT, the signature, is then computed (signed) by the following formula using a secret key. A signature is used to verify that the token isn’t forged or manipulated. Because it is signed with a secret key, we can validate the authenticity of the sender. Here is an example of a signature calculation: 

base64Url(HMACSHA256(base64Url(header) + ”.” + base64Url(payload), secret_key)) = epQym-1JoN9eep458VBZw4-cVhwfmsI1cfaGa6PE818
```
https://unit42.paloaltonetworks.com/jsonwebtoken-vulnerability-cve-2022-23529/
[CWE-20](https://cwe.mitre.org/data/definitions/20.html)
`CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:L/I:H/A:L`

### DIFF
--- a/apps/example-nodejs-jwt/package-lock.json
+++ b/apps/example-nodejs-jwt/package-lock.json
@@ -184,7 +184,7 @@
     "buffer-equal-constant-time": {
       "version": "1.0.1",
       "resolved": "https://registry.npmjs.org/buffer-equal-constant-time/-/buffer-equal-constant-time-1.0.1.tgz",
-      "integrity": "sha1-+OcRMvf/5uAaXJaXpMbz5I1cyBk="
+      "integrity": "sha512-zRpUiDwd/xk6ADqPMATG8vc9VPrkck7T07OIx0gnjmJAnHnTVXNQG3vfvWNuiZIkwu9KrKdA1iJKfsfTVxE6NA=="
     },
     "bytes": {
       "version": "3.1.0",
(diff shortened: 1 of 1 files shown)


marcoandredinis:
Thank you for your contribution.
Closing in favor of this PR https://github.com/gravitational/teleport-plugins/pull/753

PR #741, opened 2023-01-18
 Swap to using StdEncoding over RawStdEncoding to handle padding 1243c51 

### LINKED ISSUE #702: The new `identity_file_base64` doesn't decode base64 properly
I was excited to see this issue was addressed as I needed it to automate our use case, so I got around to testing it after it was released. Unfortunately, I'm seeing the provider return this error:

```text
│ Error: Failed to decode Identity file using base 64
│ 
│   on test.tf line 20, in provider "teleport":
│   20: provider "teleport" {
│ 
│ Error when trying to decode: illegal base64 data at input byte 9854
```

I can use `base64 -d` on my Mac to decode the same base64 string without any errors, so I'm not too sure why the provider is having difficulties with it.

EdwardDowling:
Padding is not expected by RawStdEncoding, swap to StdEncoding so padding is handled.
RawStdEncoding doesn't expect the '=' at the end of the identity file that is added for padding.
Fixes #702


zmb3:
This PR (and associated commit message) could use a better title/description.

zmb3 (approved):

EdwardDowling:
> This PR (and associated commit message) could use a better title/description.

Added some extra explanation.

Joerger (approved):

r0mant (approved):

PR #742, opened 2023-01-18
Update Teleport API to v12

marcoandredinis:
Change list:
- update docs/examples/code to use RoleV6
- gen Terraform schema using the new api version
- update Terraform tests to use the new RoleV6
- remove `tsh bench` as it was only used in tests
- fix event handler test (mismatch intance join events)
- add `-o "StrictHostKeyChecking no"` to `tsh ssh` commands
- update `lib/testing/integration/download.go/downloadVersions` to match the current release

Manual tests:
Terraform
Start teleport 11.2.2 and add a role via terraform (using 11.2.1 plugin)
Upgrade to teleport 12.0.0.alpha and `tf plan` using 12.0.0 returned no changes
Set the version to v6 and added Kube Resource Rules, issued `tf apply` and no changes were returned.


access/slack/README.md
```diff
@@ -69,7 +69,7 @@ spec:
     rules:
       - resources: ['access_request']
         verbs: ['list', 'read', 'update']
```
greedy52, line 71: The official guide has these in the allow
```
spec:
  allow:
    rules:
      - resources: ['access_request']
        verbs: ['list', 'read']
      - resources: ['access_plugin_data']
        verbs: ['update']
```
Which is one is more up-to-date? During my testing, it doesn't seem to make a difference.

marcoandredinis: tldr: only list and read are required for `access_request`

Access plugins used to allow for the acceptance of an `access_request` as part of the plugin within the plugin's App.
Example in slack:
When an access request was created, slack plugin would send a message to a specific workspace/channel and users would be able to click on a Accept/Deny buttons.
That relaxed the security we were aiming for (not exactly sure about the full details).
So, now, we only send a message with a link to teleport web ui itself.

Joerger (approved):

zmb3:
Not super familiar with plugins. Why do we do this before v12 releases?

r0mant:
Yes, we're holding this off until 12 GA is cut.

r0mant (approved):
@marcoandredinis lgtm but please update version to 12.0.0 before merging

PR #746, opened 2023-01-24
Extract Teleport client interface in access plugins

justinas:
Contributes to https://github.com/gravitational/cloud/issues/2506, see also RFD https://github.com/gravitational/cloud/pull/2678 .

This extracts the methods used from the Teleport API Client into an internal interface. This interface type (rather than a concrete `*client.Client`) is passed to components. This (with a [small change in Auth](https://github.com/gravitational/teleport/commit/4c888737c28f254b754889dac17fe4dbd24bfbdf)) will allow us to plug-in the Auth server directly when running plugins in Cloud, removing the need to use gRPC.


access/common/config.go
```diff
@@ -41,6 +50,36 @@ func (c BaseConfig) GetTeleportConfig() lib.TeleportConfig {
...
+		Credentials: c.GetTeleportConfig().Credentials(),
+		DialOpts: []grpc.DialOption{
+			grpc.WithConnectParams(grpc.ConnectParams{Backoff: bk, MinConnectTimeout: initTimeout}),
+			grpc.WithReturnConnectionError(),
+		},
+	})
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	clt = clt.WithCallOptions(grpc.WaitForReady(true))
```
justinas, line 78: `WaitForReady` is now set globally for the entire client. Previously, it was set in some places:

* [Ping](https://github.com/gravitational/teleport-plugins/blob/931c006afe119289b17adcb3a8d564094f16df44/access/common/app.go#L96)
* [WatcherJob](https://github.com/gravitational/teleport-plugins/blob/a18e314d11457f2262ecff25e0621d4fbc955085/lib/watcherjob/watcherjob.go#L55)

But not in:

* [plugindata](https://github.com/gravitational/teleport-plugins/blob/a18e314d11457f2262ecff25e0621d4fbc955085/lib/plugindata/cas.go).

Not sure if this difference is intentional, please do point out if plugindata *requires* this option to be absent.

marcoandredinis: I'm not aware of anything that could be different here

access/common/teleport/client.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"context"
+
+	"github.com/gravitational/teleport-plugins/lib/plugindata"
+	"github.com/gravitational/teleport/api/client/proto"
+	"github.com/gravitational/teleport/api/types"
+)
+
+type Client interface {
```
marcoandredinis, line 25: Missing godoc

marcoandredinis (approved):

tcsc (approved):

PR #753, opened 2023-01-31
Update NodeJS Example app dependencies

marcoandredinis:
This is the result of updating the libs we use and then
- `npm update`
- `npm audit fix --force`


Joerger (approved):

jentfoo (approved):
These updates look to have slipped through the cracks.  Can we merge this PR?

PR #759, opened 2023-02-08
Improve TestSSH assertions

marcoandredinis:
Uses a more deterministic way to assert that it printed the expected string.

The previous version was flaky based on the PS1 and how fancy the prompt was.
It failed when running on the Drone's Darwin machine.

### DIFF
--- a/event-handler/event_handler_test.go
+++ b/event-handler/event_handler_test.go
@@ -279,9 +279,6 @@ func (s *EventHandlerSuite) TestEvents() {
 	t.Log("STDERR", cmdStderr.String())
 	require.NoError(t, err)
 
-	err = stdinPipe.Close()
-	require.NoError(t, err)
-
 	// Our test session is very simple. There would be to copies of the same messages: one copy is supposed to be received
 	// via audit log, other one - via session log.
 	counters := make(map[string]int)
--- a/lib/testing/integration/ssh_test.go
+++ b/lib/testing/integration/ssh_test.go
@@ -69,10 +69,10 @@ func (s *IntegrationSSHSuite) TestSSH() {
 	err = cmd.Start()
 	require.NoError(t, err)
 
-	_, err = stdinPipe.Write([]byte("whoami\n\r"))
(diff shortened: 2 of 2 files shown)


r0mant (approved):

event-handler/event_handler_test.go
```diff
@@ -279,9 +279,6 @@ func (s *EventHandlerSuite) TestEvents() {
 	t.Log("STDERR", cmdStderr.String())
 	require.NoError(t, err)
 
-	err = stdinPipe.Close()
```
tcsc, line 282: Is this what was failing on Darwin? Or is this what was causing failures on a fancy prompt?

lib/testing/integration/ssh_test.go
```diff
@@ -69,22 +69,16 @@ func (s *IntegrationSSHSuite) TestSSH() {
 	err = cmd.Start()
 	require.NoError(t, err)
 
-	_, err = stdinPipe.Write([]byte("whoami\n\r"))
```
tcsc, line 72: Is `whoami` not available on one of our platforms?

event-handler/event_handler_test.go
marcoandredinis: This failed when I ran the tests in Darwin using go1.20, with an already closed like error.
It works on linux (at least using go.19)

However, according to docs, we should only close if we want to close it sooner than the program ends.
https://pkg.go.dev/os/exec#Cmd.StdinPipe
> A caller need only call Close to force the pipe to close sooner.

lib/testing/integration/ssh_test.go
marcoandredinis: It is available.
I changed this in order to simplify the check we do.

Depending on the prompt/shell configuration it could print one of the following:
- `\r<user>\r\n`
- `\n<user>\r\n`

So, instead of only matching for the `<user>\r\n` which would relax the check (this string could appear in the prompt), I changed this to print the user with a prefix and match on that.

tcsc (approved):

PR #788, opened 2023-03-15
CodeQL: Switch from `make` to auto build

jentfoo:
In `teleport` we saw a significant speed increase in using the autobuild.  In testing in this repo we see the build time cut in half (4min reduction), while still scanning just as much code (actually slightly more).  So it appears that this change makes sense for this repo too.

PR proposed as part of this issue: https://github.com/gravitational/SecOps/issues/269

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -29,27 +29,18 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v3
 
-    - name: Set up Go
-      uses: actions/setup-go@v3
-      with:
-        go-version-file: 'go.mod'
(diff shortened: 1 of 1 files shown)


wadells (commented):
Seems reasonable.

My only concern is that for some reason the output of make and the output of `autobuild` don't end up the same, and cause the results to be less valid for some reason.  I don't think this is likely given the current complexity of `teleport-plugins`, but it seems like an easy thing to forget if the build notably changes in the future.

Saying this as someone who first encountered `autobuild` in this PR -- I haven't done any real homework on it yet.

wadells (approved):

r0mant (approved):

PR #789, opened 2023-03-16
Warn users when an unsupported event is received by `event-handler`

### LINKED ISSUE #783: Improve error message when plugins receive an unknown event
## Current behavior:

When a Teleport plugin receives an unknown event from the Auth Server it prints a horrifying message and crashes:
`*trace.BadParameterError failed to parse event, session record is corrupted`

Teleport transfers audit events using a Protobuf [oneof](https://github.com/gravitational/teleport/blob/42fd2e19112be9326628b2688c5fd5ff377a2eba/api/types/events/events.pb.go#L5358) message wrapper.

When the Auth Server is running with a newer version than the plugins and it has new event types that the older plugins version doesn't support, the event is unmarshaled as `nil` and we return an [error](https://github.com/gravitational/teleport/blob/42fd2e19112be9326628b2688c5fd5ff377a2eba/api/types/events/oneof.go#L538).

When the plugin's package loads the Auth events to send them to the upstream service, it receives an unsupported event - protobuf `Event` returns `nil`, and the unknown event payload is stored under [`XXX_unrecognized`](https://github.com/gravitational/teleport/blob/42fd2e19112be9326628b2688c5fd5ff377a2eba/api/types/events/events.pb.go#L5482) as raw bytes.

## Expected behavior:

Teleport Plugins should print a better message to the user specifying that an incompatible event was received from Auth server and an upgrade of the plugin handler to the same version as the auth server must happen.

We can check the `XXX_unrecognized` and return a different message saying that the plugins must be upgraded to work correctly. It should also mention that during the downtime no events are lost and once the plugin's code is updated, it continues from the last processed event.

Auth Server: v11.3.5
Plugins: v10.3.2

tigrato:
When a trace.BadParameter error is returned from SearchEvents, which means that the teleport-plugins protobuf version is incompatible with the Teleport Auth protobuf version. This is a fatal error and the plugin should exit because it won't be able to parse the event that is supported by the newer version of Teleport Auth but not by the older version of teleport-plugins.
teleport-plugins compatibility is strictly tied to the Teleport Auth version and the plugin should be updated to the latest version when the Teleport Auth version is updated. plugins break our compatibility promise of supporting clients 1 major version behind Auth. We don't support older versions of teleport-plugins with newer versions of Teleport Auth even if they are in the same major version because we can not guarantee that the plugin will be able to parse the events introduced between minor or patch versions of Teleport Auth. This is a temporary solution until we have a better way to handle this.

Fixes #783

### DIFF
--- a/event-handler/teleport_events_watcher.go
+++ b/event-handler/teleport_events_watcher.go
@@ -130,9 +130,27 @@ func (t *TeleportEventsWatcher) flipPage() bool {
 // fetch fetches the page and sets the position to the event after latest known
 func (t *TeleportEventsWatcher) fetch(ctx context.Context) error {
 	log := logger.Get(ctx)
-
 	b, nextCursor, err := t.getEvents(ctx)
-	if err != nil {
+	// When a trace.BadParameter error is returned, it means that the teleport-plugins
+	// protobuf version is incompatible with the Teleport Auth protobuf version.
+	// This is a fatal error and the plugin should exit because it won't be able to parse the
+	// event that is supported by the newer version of Teleport Auth but not by the
+	// older version of teleport-plugins.
+	// teleport-plugins compatibility is strictly tied to the Teleport Auth version
+	// and the plugin should be updated to the latest version when the Teleport Auth
+	// version is updated. plugins breaks our compatibility promise of supporting
+	// clients 1 major version behind Auth. We don't support older versions of teleport-plugins
+	// with newer versions of Teleport Auth even if they are in the same major version
+	// because we can not guarantee that the plugin will be able to parse the events
+	// introduced between minor or patch versions of Teleport Auth.
+	// This is a temporary solution until we have a better way to handle this.
+	if trace.IsBadParameter(err) {
+		return trace.BadParameter(
+			"teleport-plugins version is incompatible with the Teleport Auth version. " +
+				"Please update teleport-plugins to the same version as the Teleport Auth server." +
+				"Please upgrade the plugins exporter to resume exporting events.",
+		)
+	} else if err != nil {
 		return trace.Wrap(err)
 	}
 
--- a/event-handler/teleport_events_watcher.go
+++ b/event-handler/teleport_events_watcher.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/gravitational/teleport/api/client"
+	"github.com/gravitational/teleport/api/client/proto"
 	"github.com/gravitational/teleport/api/types"
 	"github.com/gravitational/teleport/api/types/events"
 	"github.com/gravitational/trace"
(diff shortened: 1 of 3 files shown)


event-handler/teleport_events_watcher.go
```diff
@@ -130,9 +131,27 @@ func (t *TeleportEventsWatcher) flipPage() bool {
...
+	// event that is supported by the newer version of Teleport Auth but not by the
+	// older version of teleport-plugins.
+	// teleport-plugins compatibility is strictly tied to the Teleport Auth version
+	// and the plugin should be updated to the latest version when the Teleport Auth
+	// version is updated. plugins breaks our compatibility promise of supporting
+	// clients 1 major version behind Auth. We don't support older versions of teleport-plugins
+	// with newer versions of Teleport Auth even if they are in the same major version
+	// because we can not guarantee that the plugin will be able to parse the events
+	// introduced between minor or patch versions of Teleport Auth.
+	// This is a temporary solution until we have a better way to handle this.
```
tigrato, line 147: It might be worth checking if we do not lose events when Auth Server and any other component are in the same major version but auth is one minor/patch version behind that component  -eg auth in 12.0 but kube agent in 12.1 and introduced a new audit event.
From our compatibility matrix, they are compatible and customers are allowed to use that setup.

event-handler/teleport_events_watcher.go
```diff
@@ -130,9 +130,27 @@ func (t *TeleportEventsWatcher) flipPage() bool {
...
+	if trace.IsBadParameter(err) {
+		return trace.BadParameter(
+			"teleport-plugins version is incompatible with the Teleport Auth version. " +
+				"Please update teleport-plugins to the same version as the Teleport Auth server." +
+				"Please upgrade the plugins exporter to resume exporting events.",
```
hugoShaka, line 151: Non-blocking suggestions: it would be great to output the plugin version and the Teleport auth version. It might also help to perform this check on startup (but might cause crashloops after a cluster update, which is a painful user experience, and there would be no upgrade path without disruption)

hugoShaka (approved):

tigrato:
This PR depends on https://github.com/gravitational/teleport/pull/23166 to work.
It is blocked until we have the final product decision.

event-handler/teleport_events_watcher.go
```diff
@@ -130,9 +132,27 @@ func (t *TeleportEventsWatcher) flipPage() bool {
...
+	// and the plugin should be updated to the latest version when the Teleport Auth
+	// version is updated. plugins breaks our compatibility promise of supporting
+	// clients 1 major version behind Auth. We don't support older versions of teleport-plugins
+	// with newer versions of Teleport Auth even if they are in the same major version
+	// because we can not guarantee that the plugin will be able to parse the events
+	// introduced between minor or patch versions of Teleport Auth.
+	// This is a temporary solution until we have a better way to handle this.
+	if trace.IsBadParameter(err) {
+		return trace.BadParameter(
+			"teleport-plugins version is incompatible with the Teleport Auth version. " +
```
r0mant, line 151: Instead of "teleport-plugins", can we say "Teleport event handler" explicitly everywhere?

tigrato: done in d346fd255950e0acc18427e611662f4b1df8c3ad

PR #793, opened 2023-03-23
Replace slack and supporting libs with teleport dependencies

justinas:
That code was vendored in teleport in https://github.com/gravitational/teleport/pull/23045 , backport is pending in https://github.com/gravitational/teleport/pull/23475 .

This makes the remaining code to depend on those packages in teleport.

TODO: when https://github.com/gravitational/teleport/pull/23475 merges, I'll replace dependencies in `go.mod` to point to `branch/v12` or `v12.X.Y`.


justinas:
Dependency review is failing due to a supposed CVE in teleport, however I think this is due to the fact that we do not use Go major version scheme with subfolders, and this makes Go.mod pseudo-version `1.3.3-0.20230323122326-68aaa58860d1`. The lint thinks we're using teleport v1.3.3.

justinas:
I'm opening this for initial review, keep in mind the TODOs in description.

r0mant (approved):
lgtm once TODOs are addressed

justinas:
> lgtm once TODOs are addressed

@r0mant I've added an update script and bumped the dependencies to `branch/v12` (since there is no v12 tag with these changes yet). PTAL.

r0mant (approved):
+243 -10,459 🥇

PR #797, opened 2023-04-04
Fix integration tests in Drone

justinas:
After #793 , the tests on Drone master builds broke, because dependency subdirectories may not be writable. This utilizes https://github.com/gravitational/teleport/pull/23773 to specify a download path for Teleport explicitly.

`RUNNER_TEMP` is a GitHub Actions variable, however it is easier to define it rather than add a new one, and I imagine we'll be moving more stuff into GHA as time goes on.

Test build: https://drone.platform.teleport.sh/gravitational/teleport-plugins/2391/2/3

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -26,6 +26,7 @@ steps:
   - name: Run tests
     image: golang:1.20.2
     environment:
+      RUNNER_TEMP: /tmp
       TELEPORT_ENTERPRISE_LICENSE:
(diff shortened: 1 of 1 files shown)


tigrato (approved):

hugoShaka (approved):

wadells (approved):
Thanks!

r0mant (approved):

PR #810, opened 2023-04-26
feat: Masks the ID with a random string instead of using the token as…

pkeecom:
… ID which is a security issue appearing in the plan.


marcoandredinis:
Hey :wave: 
Thank you for your contribution.
I've created a PR based on your work but slightly different (we auto generate the files you changed, so we have to change the template instead).

I'll add you as a reviewer.
Again, thank you 🙏

https://github.com/gravitational/teleport-plugins/pull/815

Issue #539, opened 2022-05-04
ARM/M1 builds for plugins


AgDude:
Please add ARM builds for Apple M1 to the published releases so we don't have to build the terraform plugin manually.

gz#5327

gz#5468

programmerq:
Related to #235

alexlafreniere:
+1, we have a mix of arm64 and amd64 machines amongst our dev team and having published builds for both CPU architectures would make our lives much much easier.

zarko-a:
+1

cbnorman:
+1

benarent:
+1 but I did find `make install` for the terraform provider fixed my issue. Still an extra step, I would love to see this.

pschisa:
related: https://github.com/gravitational/teleport/issues/4226

camscale:
This is done for the terraform plugin, and there are no plans to do it for the event handler plugin. No other plugins are built for Mac. I'll leave this open until a new release of the plugins is released.

sklirg:
Just cross-posting from a recent Slack-thread to here.

There is now a [release supporting ARM/Mac M1 builds out](https://github.com/gravitational/teleport-plugins/releases/tag/v13.0.0), but the terraform provider still doesn't work on an M1 Mac:
```
❯ terraform init

Initializing the backend...

Initializing provider plugins...
- Finding terraform.releases.teleport.dev/gravitational/teleport versions matching ">= 13.0.0"...
╷
│ Error: Incompatible provider version
│
│ Provider terraform.releases.teleport.dev/gravitational/teleport v13.0.0 does not have a package available for your current platform, darwin_arm64.
│
│ Provider releases are separate from Terraform CLI releases, so not all providers are available for all platforms. Other versions of this provider may have
│ different platforms supported.

❯ bat main.tf --plain
terraform {
  required_providers {
     teleport = {
       version = ">= 13.0.0"
       source = "terraform.releases.teleport.dev/gravitational/teleport"
     }
  }
}
```

webvictim:
@camscale AFAIK the Terraform provider is now published for M1 Macs and the instructions in docs will work out of the box. We should be able to close this issue, I think?

dm3ch:
I just tested the official doc - https://goteleport.com/docs/management/dynamic-resources/terraform-provider/#step-23-create-a-terraform-configuration .

Unfortunately, I got an error:
```
❯ terraform init

Initializing the backend...

Initializing provider plugins...
[...]

PR #812, opened 2023-05-03
Build universal binary for MacOS terraform plugin

camscale:
Build the terraform plugin for arm64 and universal (arm64 + amd64 in the
same binary) for MacOS, in addition to the existing amd64
(runner-native) build.

Issue: https://github.com/gravitational/teleport-plugins/issues/539


r0mant (approved):

terraform/Makefile
```diff
@@ -30,6 +30,18 @@ clean: tfclean
 build: clean
 	GOOS=$(OS) GOARCH=$(ARCH) $(CGOFLAG) go build -o $(BUILDDIR)/terraform-provider-teleport $(BUILDFLAGS)
 
+build-darwin-universal: $(addprefix $(BUILDDIR)/terraform-provider-teleport_,arm64 amd64)
+	lipo -create -output $(BUILDDIR)/terraform-provider-teleport $^
+	rm $^
+
+# Architecture-specific binaries for the universal binary are extracted from
+# the release tarball. make will not automatically build this; you will need
```
zmb3, line 38: Will this at least fail if you forget to build the release tarball first?

zmb3 (approved):

terraform/Makefile
camscale: Yes. Make will say:
```
make[1]: *** No rule to make target 'build/terraform-provider-teleport_arm64', needed by 'build-darwin-universal'.  Stop.
```

I did try to add a rule to build it automatically, but due to ordering and that `release` does a clean which removes the `build` directory, it ends up deleting the first binary when it builds the second, and I didn't want to get into refactoring the `clean` targets.

PR #821, opened 2023-05-11
RFD 3: Event handler unknown events

tigrato:
This RFD proposes a new process for the event handler to be able to transform events into JSON even when the structure of the received event is unknown to the event handler protobuf version.

Part of https://github.com/gravitational/teleport/issues/23388


rfd/0003-plugins-unknown-events.md
```diff
@@ -0,0 +1,143 @@
...
+that wasn't able to be deserialized. If the message with `XXX_unrecognized` is marshaled
+back to protobuf, the `XXX_unrecognized` field will be included in the marshaled message.
+This means that the raw -> unmashaled -> marshaled -> raw path will be the same even if the
+message is unknown to the event handler.
+
+With this in mind, the event handler will be able to send the raw protobuf event to the
+WASM module that will deserialize the raw protobuf event - using the same protobuf version Auth has -
+and serialize it to JSON and return it to the caller that will send it to the upstream service.
+
+The WASM module will be served by the Auth server and will be downloaded by the event handler
```
r0mant, line 67: How does event handler know auth has been up-/downgraded?

Also, is it possible to "refresh" (redownload) the module upon receiving an error trying to parse event (i.e. for scenario when auth has been upgraded but event handler hasn't fetched the new schema yet)? Basically, lazy loading - then we wouldn't need to worry about auth lifecycle.

tigrato: Event handlers check the Auth's ping endpoint for auth's version. 
If the returned version differs from the previously stored version, it will re-download the file. The same happens when a new event fails the marshaling.

rfd/0003-plugins-unknown-events.md
```diff
@@ -0,0 +1,143 @@
...
+download the WASM module each time the Auth server is upgraded/downgraded.
+
+When the event handler receives an unknown event to its protobuf version, the protobuf
+message includes a field called `XXX_unrecognized` that contains the raw slice of bytes
+that wasn't able to be deserialized. If the message with `XXX_unrecognized` is marshaled
+back to protobuf, the `XXX_unrecognized` field will be included in the marshaled message.
+This means that the raw -> unmashaled -> marshaled -> raw path will be the same even if the
+message is unknown to the event handler.
+
+With this in mind, the event handler will be able to send the raw protobuf event to the
```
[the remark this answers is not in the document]
tigrato: It's possible but not sure if it's the ideal.

While writing this, I spoke with  @hugoShaka about it and he prefers to have a single path - not having a path for known and another for unknown events.  This would simplify the code and allow us to understand bugs.

I changed the RFD so that everything will always hit the WASM routine. 

The current time is:
- `json directly`: 132.699µs
- `via wasm`: 534.154µs -> but involves 2 extra marshals and unmarshals.

rfd/0003-plugins-unknown-events.md
```diff
@@ -0,0 +1,143 @@
...
+WASM module that will deserialize the raw protobuf event - using the same protobuf version Auth has -
```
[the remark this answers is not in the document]
tigrato: Partially replyed above

espadolini (commented):
Why can't we just have APIs that return events in JSON instead? Taking what's almost always JSON (or JSON-like) data, converting it to protobuf just to return it together with a WASM module to convert the data back into JSON seems far too convoluted.

rfd/0003-plugins-unknown-events.md
```diff
@@ -0,0 +1,384 @@
+---
+authors: Tiago Silva (tiago.silva@goteleport.com)
+state: draft
+---
+
+# RFD 3 - Event handler unknown events
+
+## Required Approvers
+
+* Engineering: @r0mant && ( @rosstimothy || @edoardo )
```
espadolini, line 10:
```suggestion
* Engineering: @r0mant && ( @rosstimothy || @espadolini )
```

rfd/0003-plugins-unknown-events.md
```diff
@@ -0,0 +1,384 @@
...
+The proposal consists of having Auth server serve the events as JSON when the
+client requests the JSON endpoints. This will allow the event handler to receive
+the events as JSON and send them to the upstream service without having to
+deserialize the Protobuf and serialize it again to JSON.
+
+To achieve this, we will add two new endpoints to the Auth server that will return
+the events as JSON. The event handler will use these endpoints to retrieve the events
+and session events as JSON.
+
+The new endpoints will be:
```
espadolini, line 57: It doesn't change the nature of this proposal as it's just an implementation detail, but we've been trying to migrate away from the monolithic protobuf service approach and into split services, so maybe this should serve as the start of the `teleport.auditlog.v1.AuditLogService`?

rfd/0003-plugins-unknown-events.md
```diff
@@ -0,0 +1,384 @@
...
+    string ID = 2;
+    // Time is the time when the event was generated.
+    google.protobuf.Timestamp Time = 3 [
+        (gogoproto.stdtime) = true,
+        (gogoproto.nullable) = false
+    ];
+    // Index is the index of the event.
+    int64 Index = 4;
+    // Payload is the JSON event payload.
+    bytes Payload = 5;
```
espadolini, line 96: I think using a `google.protobuf.Struct` rather than JSON as the format here would be better; perhaps we should refer to `EventUnstructured` rather than `EventJSON` or some other name like that?

In practice it's basically the same thing, but at least with a Struct we're guaranteed to have a well-formed object (in the json sense) in flight.

rfd/0003-plugins-unknown-events.md
```diff
@@ -0,0 +1,384 @@
...
+}
+```
+
+Auth server serializes the events to JSON and returns them using `EventJSON` struct.
+The serialization is done using the default event JSON representation for each event type
+besides the `print` event type. The `print` event type is serialized using a custom
+JSON representation because the default JSON representation does not include the
+`data` field which is required by the event handler.
```
espadolini, line 104: Again, implementation detail, but almost all if not all audit log backends have events start off as JSON then parsed into a `map[string]any`; it might be possible to avoid roundtripping through the oneof events and just keep the event data unstructured, which would even allow auth servers that don't know about an event to forward it through correctly.

espadolini (commented):
I think that allowing clients to fetch unstructured event data if they don't need to do anything but store it unstructured somewhere else (like what the event plugins do) makes perfect sense.

rfd/0003-plugins-unknown-events.md
tigrato: done

rfd/0003-plugins-unknown-events.md
tigrato: done

espadolini (approved):

rfd/0003-plugins-unknown-events.md
```diff
@@ -0,0 +1,352 @@
...
+When the event handler is not up to date with the Auth server, it will not be
+able to handle new events that are not known to the event handler. This can
+cause the event handler to crash (previous behavior) or ignore the unknown events.
+
+This is problematic because the event handler is a critical component that
+should be able to handle all types of events. If the event handler is not able
+to handle new events, it can miss important events and drift from the Auth server truth.
+
+The Teleport event handler receives events from the Auth server, serializes them
+using JSON and sends them to the upstream service (e.g. Splunk, FluentD, etc).
```
tcsc, line 35:
```suggestion
using JSON and sends them to the downstream service (e.g. Splunk, FluentD, etc).
```

rfd/0003-plugins-unknown-events.md
```diff
@@ -0,0 +1,352 @@
...
+Currently, the event handler ignores unknown events.
+
+## Proposal
+
+The proposal consists of having Auth server serve unstructured events when the
+client requests those endpoints. This will allow the event handler to receive
+the events with a JSON-like structure and send them to the upstream service without having to
+deserialize the Protobuf and serialize it again to JSON.
+
+To achieve this, we will add a new Protobuf service - `AuditLogService - to Auth server that will return
```
tcsc, line 53:
```suggestion
To achieve this, we will add a new Protobuf service - `AuditLogService` - to Auth server that will return
```

tcsc (approved):

tcsc:
The main question I initially had was _why not just emit JSON at the source?_, i.e. directly from the auth server, rather than wrap unstructured data in a protobuf envelope. 

I think the comments & answers in this PR answered that pretty well, but it might be worth adding "Emit unstructured JSON from the Auth server" option in the **_Alternatives_** section, with a couple of sentences about why using GRPC + protobuf `struct` is better.

rfd/0003-plugins-unknown-events.md
```diff
@@ -0,0 +1,352 @@
...
+}
+
+```
+
+Auth server converts the events to `structpb.Struct` and returns them using `EventUnstructured` struct.
+The conversion is done using the default event JSON representation for each event type
+besides the `print` event type. For the `print` event type, an extra field
+`data` is appended to the `structpb.Struct` because the default JSON representation does not include the
+`data` field which is required by the event handler.
+
```
tcsc, line 150: How does this encoding process look on the `auth` side? I am imagining something that is the moral equivalent of this :

```go
func MarshalEvent(structuredEvent *someTeleportEventType) *EventUnstructured {
  jsonText, _ := json.Marshal(&structuredEvent)
  
  var structMap map[string]any
  _ = json.Unmarshal(jsonText, &structMap)

  return &EventUnstructured {
    unstructured: structpb.NewStruct(structMap),
  }
}
```
(although hopefully without necessarily having to go all the way down to text to perform the `eventType` to `map[string]any` conversion).

Or have I got the wrong end of the stick entirely?

tigrato: The process is similar but you can unmarshal the JSON directly into the struct ([real code](https://github.com/gravitational/teleport/pull/27026/files#diff-0d992726e983ecd7ccd935e9769652763907ad1a88ae462050d2770e557d3747R5224-R5257)):

```go

func MarshalEvent(structuredEvent *someTeleportEventType) *EventUnstructured {
  jsonText, _ := json.Marshal(&structuredEvent)
  
  var struct structpb.Struct
  _ = struct.UnmarshalJSON(jsonText)

  return &EventUnstructured {
    unstructured: struct,
  }
}


```

rfd/0003-plugins-unknown-events.md
```diff
@@ -0,0 +1,352 @@
...
+message EventsUnstructured {
+  // Items is a list of unstructured formatted audit events.
+  repeated EventUnstructured items = 1;
+  // the key of the last event if the returned set did not contain all events found i.e limit <
+  // actual amount. this is the key clients can supply in another API request to continue fetching
+  // events from the previous last position
+  string last_key = 2;
+}
+
+// EventUnstructured represents a siggle events.AuditEvent in an unstructured format.
```
r0mant, line 127:
```suggestion
// EventUnstructured represents a single events.AuditEvent in an unstructured format.
```

rfd/0003-plugins-unknown-events.md
```diff
@@ -0,0 +1,352 @@
...
+import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/gravitational/teleport/api/gen/proto/go/teleport/auditlog/v1;auditlogv1";
+
+// AuditLogService provides methods to access audit log.
+service AuditLogService {
+  // StreamUnstructuredSessionEvents streams audit events from a given session recording in an unstructured format.
+  // This endpoint is used by the event handler to retrieve the session events as JSON.
+  rpc StreamUnstructuredSessionEvents(StreamUnstructuredSessionEventsRequest) returns (stream EventUnstructured);
```
r0mant, line 74: A naming nit, but what do you think about replacing generic "unstructured" word everywhere with just explicit "json" since it's basically what it returns?

rfd/0003-plugins-unknown-events.md
```diff
@@ -0,0 +1,352 @@
...
+  // GetUnstructuredEvents gets events from the audit log in an unstructured format.
+  // This endpoint is used by the event handler to retrieve the events as JSON.
+  rpc GetUnstructuredEvents(GetUnstructuredEventsRequest) returns (EventsUnstructured);
```
r0mant, line 77: Have you considered instead of adding another set of "unstructured" counterparts for everything, just adding "unstructured" or "raw" flag to existing GetEventsRequest and then just adding "UnstructuredEvent" (or "RawEvent" or "JSONEvent") to the events.OneOf?

PR #824, opened 2023-05-14
Fixes arm64 terraform provider promotion

tcsc:
The code used to deduce the CPU architecture from a filename used `aarch64` as a 64-bit ARM architecture tag. The terraform provider (and other Teleport artifacts) uses `arm64` to mean the same thing.

This patch updates the plugin filename parser to match.

### DIFF
--- a/tooling/internal/filename/parse.go
+++ b/tooling/internal/filename/parse.go
@@ -25,7 +25,7 @@ import (
 )
 
 var (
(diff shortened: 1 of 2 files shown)


webvictim (approved):
Thanks Trent!

fheinecke (approved):

camscale (approved):

r0mant (approved):

PR #826, opened 2023-05-16
Fix email plugin values.yaml example in the doc

PLJanicki:
There is a wrong key used in the teleport-plugin-email-values.yaml example


r0mant:
@PLJanicki Thank you for your PR, the change will be merged in https://github.com/gravitational/teleport-plugins/pull/828.

PR #832, opened 2023-05-26
TF Ref Docs: Update to 13.0.3

marcoandredinis:
We released 13.0.3
This PR updates the reference docs for it.

It also changes the way we escape the fields description. It was using html escape previsouly, but then some words would trigger the cspell lint check we have in `Lint(Docs)` CI Step. Eg `doesn't` became `doesn&#39;t` and `doesn` isn't a valid word.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -198,8 +198,8 @@ update-version:
 	$(SED) '1s/.*/VERSION=$(VERSION)/' access/email/Makefile
 	$(SED) '1s/.*/VERSION=$(VERSION)/' terraform/install.mk
 	$(MAKE) update-helm-version
-	$(MAKE) terraform-gen-tfschema
 	$(MAKE) update-teleport-dep-version
+	$(MAKE) terraform-gen-tfschema
 
 # Update all charts to VERSION
 .PHONY: update-helm-version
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -198,8 +198,8 @@ update-version:
 	$(SED) '1s/.*/VERSION=$(VERSION)/' access/email/Makefile
 	$(SED) '1s/.*/VERSION=$(VERSION)/' terraform/install.mk
 	$(MAKE) update-helm-version
-	$(MAKE) terraform-gen-tfschema
 	$(MAKE) update-teleport-dep-version
+	$(MAKE) terraform-gen-tfschema
```
marcoandredinis, line 202: @r0mant I had to change the order so that we generate the tf schema for the version we are about to release.
The previous order meant that we were releasing the previous version's fields.
Eg, when releasing 13.0.3 we would be using the teleport/api 13.0.2

With this change we should get the latest schema when doing a release.
So, when you release 13.0.4, we should get the terraform schema of teleport/api 13.0.4.

r0mant (approved):

Joerger (approved):

PR #836, opened 2023-05-29
Use unstructured endpoints for fetching audit log events

tigrato:
This PR converts the Teleport Event handler to use the unstructured events to fetch events from the Audit log. This change, allows the older events handlers to continue exporting events from Auth's audit log into the upstream service without requiring manual upgrades each time auth version changes.

With the current implementation, the event handler will no longer skip events each time the event is unknown to its protobuf representation and can still receive and push them as JSON.

Requires https://github.com/gravitational/teleport/pull/27026 
Part of https://github.com/gravitational/teleport-plugins/pull/821 
Completes https://github.com/gravitational/teleport/issues/23388


event-handler/teleport_events_watcher.go
```diff
@@ -38,8 +39,14 @@ const (
 
 // TeleportSearchEventsClient is an interface for client.Client, required for testing
 type TeleportSearchEventsClient interface {
+	// SearchEvents searches for events in the audit log and returns them using their protobuf representation.
 	SearchEvents(ctx context.Context, fromUTC, toUTC time.Time, namespace string, eventTypes []string, limit int, order types.EventOrder, startKey string) ([]events.AuditEvent, string, error)
+	// StreamSessionEvents returns session events stream for a given session ID using their protobuf representation.
 	StreamSessionEvents(ctx context.Context, sessionID string, startIndex int64) (chan events.AuditEvent, chan error)
```
marcoandredinis, line 45: Do we still use those?

event-handler/teleport_event.go
```diff
@@ -67,109 +63,62 @@ type TeleportEvent struct {
...
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
 	evt := &TeleportEvent{
 		Cursor: cursor,
 		Type:   e.GetType(),
-		Time:   e.GetTime(),
+		Time:   e.Time.AsTime(),
 		Index:  e.GetIndex(),
+		ID:     e.Id,
```
marcoandredinis, line 77: There was some extra logic when setting the `id`. Is it no longer required?

marcoandredinis (approved):
LGTM :pray: 

Can we please add a test that ensures unknown events get forwarded?

event-handler/teleport_events_watcher.go
tigrato: I left them if someone wants to write a custom exporter based on protobuf structures

event-handler/teleport_event.go
tigrato: That logic was moved to auth server

event-handler/teleport_events_watcher.go
marcoandredinis: I think we should remove it, but not blocking because of it
In the future™️, someone will ask why we have those and will probably remove them :joy:

.github/workflows/unit-tests.yaml
```diff
@@ -11,8 +11,15 @@ env:
...
-  test:
-    name: Plugins Tests (Go)
+# This test exists to ensure that we don't break compatibility of the event handler
+# when we merge the unstructured events support.
+# This is a temporary test and should be removed in Teleport 14.
+# Since unstructured events API was intoduced in 13.1.0, this test fallbacks to
+# the old SearchEvents and StreamSessionIDEvents API and should be able to complete
+# without errors.
+# DELETE IN 14.0.0
+  test-pre-unstructured:
```
r0mant, line 21:
```suggestion
  test-structured:
```

Issue #837, opened 2023-05-30
Allow setting no expiry on `teleport_provision_token` resource


AClarkie:
The `teleport_provision_token` resource requires a `metadata.expires` field, however it's now possible to create tokens without an expiry, for example the below yaml can be used:

```
kind: token
version: v2
metadata:
  name: am-token
spec:
  roles: [Bot]

  join_method: iam
```

I've spoken with @zmb3 and @stevenGravy and they believe this is a missing feature potentially following the change [here](https://goteleport.com/docs/changelog/#join-token-api-changes), more context can be found on the below Slack thread
https://goteleport.slack.com/archives/CEZH6UL64/p1685450022714919

PR #838, opened 2023-05-30
TF: Provision Token does not require `Metadata.Expires` fields

marcoandredinis:
Provision Tokens can be created without an expiration date.

Expireless tokens are possible since https://github.com/gravitational/teleport/pull/21133
Fixes https://github.com/gravitational/teleport-plugins/issues/837

### DIFF
--- a/terraform/example/provision_token.tf.example
+++ b/terraform/example/provision_token.tf.example
@@ -15,3 +15,17 @@ resource "teleport_provision_token" "example" {
     roles = ["Node", "Auth"]
   }
 }
+
+resource "teleport_provision_token" "iam-token" {
+  metadata = {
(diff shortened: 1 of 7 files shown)


marcoandredinis:
> Could we update https://github.com/gravitational/teleport-plugins/blob/master/terraform/example/provision_token.tf.example#L5 to add an example of a token without an expiry time?

Sure :+1: Added another example

> Also https://goteleport.com/docs/reference/terraform-provider/#metadata-6 still seems to suggest that the `expires` field is required.

We'll update the reference as soon as we get a new release :+1:

r0mant (approved):

Joerger (approved):

PR #840, opened 2023-05-31
Patch install scripts

alexfornuto:
While testing [docs](https://github.com/gravitational/teleport/pull/26783) I found that the install scripts are broken, because the look for the binaries in the same directory instead of `./build`. I testing this for the Discord, email, and Jira plugins, and feel safe assuming the rest of the access plugins behave the same.

### DIFF
--- a/access/discord/install
+++ b/access/discord/install
@@ -14,7 +14,7 @@ DATADIR=/var/lib/teleport/plugins/discord
 [ ! $(id -u) != "0" ] || { echo "ERROR: You must be root"; exit 1; }
 cd "$(dirname "$0")"
 mkdir -p "$BINDIR" "$DATADIR" || exit 1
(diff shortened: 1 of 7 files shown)


hugoShaka:
`install` scripts are shipped in the tar.gz archives for installation, not for local installation after compilation. I opened current archives for Slack and Discord plugins and there are no `build` directory in the artifact. The suggested change would break the installation process.

```
$ curl -L -O https://get.gravitational.com/teleport-access-discord-v13.0.3-linux-amd64-bin.tar.gz
$ tar -xzf teleport-access-discord-v13.0.3-linux-amd64-bin.tar.gz
$ ls teleport-access-discord
README.md         VERSION           install*          teleport-discord*
```

alexfornuto:
@hugoShaka I see, thanks for clarifying that. I'll close this and note in the docs PR not to use the script when compiling.

PR #841, opened 2023-06-01
Ignore error when session doesn't return events.

tigrato:
If the given session doesn't have any events, thee session index file wasn't created and cleaning results in error.

This PR checks if the error is `os.IsNotExist` and returns nil if it's the case so it can resume the normal operation.

### DIFF
--- a/event-handler/session_events_job_test.go
+++ b/event-handler/session_events_job_test.go
@@ -1,0 +1,59 @@
+// Copyright 2023 Gravitational, Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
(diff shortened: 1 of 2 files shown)


zmb3 (approved):
Any possibility of getting some test coverage?

tigrato:
> Any possibility of getting some test coverage?

Added in 9f5ad2be9ec52a061346923101ec34a63325335f

Other cases (non-empty events) are already covered by other tests

Joerger (approved):

event-handler/session_events_job_test.go
```diff
@@ -0,0 +1,59 @@
...
+package main
+
+import (
+	"context"
+	"testing"
+
+	"github.com/gravitational/teleport/api/client"
+	"github.com/gravitational/teleport/api/types/events"
+	"github.com/peterbourgon/diskv/v3"
+	"github.com/stretchr/testify/require"
```
Joerger, line 24: nit: goimports
```suggestion
	"github.com/peterbourgon/diskv/v3"
	"github.com/stretchr/testify/require"
	
	"github.com/gravitational/teleport/api/client"
	"github.com/gravitational/teleport/api/types/events"
```

tigrato: That always confuses me but given the current rules, my version is correct. 
Our main OSS repo is external

```
  gci:
    sections:
      - standard # Standard section: captures all standard packages.
      - default # Default section: contains all imports that could not be matched to another section type.
      - prefix(github.com/gravitational/teleport-plugins) # Custom section: groups all imports with the specified Prefix.
    skip-generated: true # Skip generated files.
    custom-order: true # Required for "sections" to take effect.
```

PR #869, opened 2023-07-19
Add standard annotation support to Slack plugin

dancmeyers:
The `teleport-kube-agent` and `teleport-cluster` charts have a standard config pattern of

```
annotations:
  resourceType1:
    keyA: valueA
  resourceType2:
    keyB: valueB
```

etc. The Slack plugin chart currently doesn't support this, having a top level `podAnnotations` value (undocumented on
https://goteleport.com/docs/reference/helm-reference/teleport-plugin-slack/) only.

I need to be able to add annotations to the `ConfigMap` and `Deployment` so that our install of `stakater/reloader` picks up when the `roleToRecipients` map changes (because we've added a new role that goes to a non-default Slack channel) and restarts the container so that it fetches its new config.

This was a problem we ran into recently, where it took a bit of time to realise that Slack notifications for some Teleport roles were going to the wrong channel.

This PR brings the Slack plugin into line with the more recent versions of the core cluster and agent helm charts, and allows setting annotations on any of the created resources via a consistent block.

For backwards compatibility, the old `podAnnotations` value has been retained but the newer `annotations.pod` version takes priority.

### DIFF
--- a/charts/access/slack/values.yaml
+++ b/charts/access/slack/values.yaml
@@ -23,6 +23,18 @@ log:
 
 secretVolumeName: "password-file"
 
+# Kubernetes annotations to apply
+# https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
+annotations:
+  # Annotations for the ConfigMap
+  config: {}
+  # Annotations for the Deployment
+  deployment: {}
+  # Annotations for each Pod in the Deployment
+  pod: {}
+  # Annotations for the Secret
+  secret: {}
+
 #
 # Deployment
 #
--- a/charts/access/slack/values.yaml
+++ b/charts/access/slack/values.yaml
@@ -48,6 +48,7 @@ imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
 
+# Deprecated way to set pod annotations. `annotations.pod` should be preferred.
 podAnnotations: {}
 
 podSecurityContext: {}
--- a/charts/access/slack/README.md
+++ b/charts/access/slack/README.md
@@ -193,4 +193,41 @@ The following values can be set for the Helm chart:
     <td><code>"INFO"</code></td>
     <td>no</td>
   </tr>
+
+  <tr>
+    <td><code>annotations.config</code></td>
+    <td>
+      Annotations to add to the configmap.
+    </td>
+    <td>map</td>
+    <td><code>{}</code></td>
+    <td>no</td>
+  </tr>
+  <tr>
+    <td><code>annotations.deployment</code></td>
+    <td>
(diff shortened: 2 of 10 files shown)


dancmeyers:
The errors from all the CI failures is the same:

```
tests on CI should run with enterprise license
```

I don't think I can do anything to fix that. Surely it's up to the CI process to define the license key to be used or similar?

charts/access/slack/values.yaml
```diff
@@ -23,6 +23,18 @@ log:
...
+# https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
+annotations:
+  # Annotations for the ConfigMap
+  config: {}
+  # Annotations for the Deployment
+  deployment: {}
+  # Annotations for each Pod in the Deployment
+  pod: {}
+  # Annotations for the Secret
+  secret: {}
```
tigrato, line 36: Can you please include a deprecation NOTE for podAnnotations?

charts/access/slack/values.yaml
```diff
@@ -23,6 +23,18 @@ log:
 
 secretVolumeName: "password-file"
 
+# Kubernetes annotations to apply
+# https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
+annotations:
+  # Annotations for the ConfigMap
+  config: {}
```
tigrato, line 30: can you please rename this to confirmap

tigrato (commented):
Thank you for your contribution! We really appreciate it!

I left some minor comments, but generally LGTM.

charts/access/slack/values.yaml
dancmeyers: I don't think that should be done.

I was explicitly following the existing pattern for the [teleport-kube-agent](https://goteleport.com/docs/reference/helm-reference/teleport-kube-agent/#annotationsconfig) and [teleport-cluster](https://goteleport.com/docs/reference/helm-reference/teleport-cluster/#annotationsconfig) helm charts, both of which use `config` rather than `configmap`.

If I'm going to change this to `configmap`, both of those (and any other charts using the same pattern which I'm not aware of), should also be altered. I think that's probably outside the scope of this change.

tigrato: Oh, you are correct.
I completely missed that we name everything as config :cry:

charts/access/slack/values.yaml
dancmeyers: Done! 👍

dancmeyers:
N.B. a buddy PR has been created in https://github.com/gravitational/teleport-plugins/pull/872, because I don't have the necessary permissions to cause the tests to pass.

PR #872, opened 2023-07-20
[buddy] Add standard annotation support to Slack plugin

tigrato:
Buddies #869 
---

The `teleport-kube-agent` and `teleport-cluster` charts have a standard config pattern of

```
annotations:
  resourceType1:
    keyA: valueA
  resourceType2:
    keyB: valueB
```

etc. The Slack plugin chart currently doesn't support this, having a top level `podAnnotations` value (undocumented on
https://goteleport.com/docs/reference/helm-reference/teleport-plugin-slack/) only.

I need to be able to add annotations to the `ConfigMap` and `Deployment` so that our install of `stakater/reloader` picks up when the `roleToRecipients` map changes (because we've added a new role that goes to a non-default Slack channel) and restarts the container so that it fetches its new config.

This was a problem we ran into recently, where it took a bit of time to realise that Slack notifications for some Teleport roles were going to the wrong channel.

This PR brings the Slack plugin into line with the more recent versions of the core cluster and agent helm charts, and allows setting annotations on any of the created resources via a consistent block.

For backwards compatibility, the old `podAnnotations` value has been retained but the newer `annotations.pod` version takes priority.

### DIFF
--- a/charts/access/slack/README.md
+++ b/charts/access/slack/README.md
@@ -193,4 +193,41 @@ The following values can be set for the Helm chart:
     <td><code>"INFO"</code></td>
     <td>no</td>
   </tr>
+
+  <tr>
+    <td><code>annotations.config</code></td>
+    <td>
+      Annotations to add to the configmap.
+    </td>
+    <td>map</td>
+    <td><code>{}</code></td>
+    <td>no</td>
+  </tr>
+  <tr>
+    <td><code>annotations.deployment</code></td>
+    <td>
+      Annotations to add to the deployment.
+    </td>
+    <td>map</td>
+    <td><code>{}</code></td>
+    <td>no</td>
+  </tr>
+  <tr>
+    <td><code>annotations.pod</code></td>
+    <td>
+      Annotations to add to every pod created by the deployment.
+    </td>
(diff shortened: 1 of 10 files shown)


hugoShaka (approved):
Code looks good, can we add a quick test checking the annotations are correctly set? I think the slack chart already has a few smoke tests.

r0mant (approved):

dancmeyers:
Also I don't think they run on CI? Which I say because one of them was failing for me locally even on a clean repo with no changes ;)

hugoShaka:
> Also I don't think they run on CI? Which I say because one of them was failing for me locally even on a clean repo with no changes ;)

The `helm-unittest` plugin has several forks, all incompatible to some degree. The various versions also had several breaking changes. This repo has snapshots generated by an old plugin version, which can cause tests to fail locally if you don't have the exact version from the right fork. We'll need the fix the version and snapshots at some point, but this also involves changing the version used in the CI, which affects the main teleport repo, ...

dancmeyers:
👍 for the merge! I also have https://github.com/gravitational/teleport/pull/29353 over on the main repo, to document these new values on the website.

Issue #845, opened 2023-06-06
terraform: calls to `schemav1.CopyLoginRuleToTerraform` copy a lock by value


espadolini:
I'm not sure if this problem is more widespread than `go vet` can detect, or if the `gogoproto`-generated types are just more resilient to this, but this is the output of `go vet` from go 1.20.4 on the current tip of master (e4b2e509c5554b93e506cb3639a27071d11a3e49):

```
# github.com/gravitational/teleport-plugins/terraform/tfschema/loginrule/v1
terraform/tfschema/loginrule/v1/loginrule_terraform.go:346:56: CopyLoginRuleToTerraform passes lock by value: github.com/gravitational/teleport/api/gen/proto/go/teleport/loginrule/v1.LoginRule contains google.golang.org/protobuf/internal/impl.MessageState contains sync.Mutex
# github.com/gravitational/teleport-plugins/terraform/provider
terraform/provider/data_source_teleport_login_rule.go:69:49: call of schemav1.CopyLoginRuleToTerraform copies lock value: github.com/gravitational/teleport/api/gen/proto/go/teleport/loginrule/v1.LoginRule contains google.golang.org/protobuf/internal/impl.MessageState contains sync.Mutex
terraform/provider/resource_teleport_login_rule.go:134:49: call of schemav1.CopyLoginRuleToTerraform copies lock value: github.com/gravitational/teleport/api/gen/proto/go/teleport/loginrule/v1.LoginRule contains google.golang.org/protobuf/internal/impl.MessageState contains sync.Mutex
terraform/provider/resource_teleport_login_rule.go:177:49: call of schemav1.CopyLoginRuleToTerraform copies lock value: github.com/gravitational/teleport/api/gen/proto/go/teleport/loginrule/v1.LoginRule contains google.golang.org/protobuf/internal/impl.MessageState contains sync.Mutex
terraform/provider/resource_teleport_login_rule.go:254:49: call of schemav1.CopyLoginRuleToTerraform copies lock value: github.com/gravitational/teleport/api/gen/proto/go/teleport/loginrule/v1.LoginRule contains google.golang.org/protobuf/internal/impl.MessageState contains sync.Mutex
[...]

PR #875, opened 2023-07-21
fix: pass proto messages by pointer

nklaassen:
This commit updates the Terraform provider to use the latest version of protoc-gen-terraform which generates code to pass messages by pointer instead of by value.

Protobuf messages are in general not safe to pass by value since they may contain a mutex.

Depends on https://github.com/gravitational/protoc-gen-terraform/pull/33

Fixes https://github.com/gravitational/teleport-plugins/issues/845


Joerger (approved):

espadolini (approved):
this was necessary for non-gogo messages, right?

PR #909, opened 2023-09-07
CI: update terraform, teleport and linter versions used in pipelines

marcoandredinis:
It looks like our darwin builder has a cloud license, and some network settings can't be changed.
Remove that test case.

image

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -19,7 +19,7 @@ workspace:
 
 steps:
   - name: Run linter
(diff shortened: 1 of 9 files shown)


jof (approved):
LGTM

This actually fixes unit testing into a working state for this repo.

fheinecke (approved):
LGTM

We should probably setup renovate and/or dependabot in this repo at some point.

PR #914, opened 2023-09-07
Drone: test-linux: Fix Terraform install

### LINKED ISSUE #913: `apt-key add` is deprecated
`apt-key add` is deprecated and used in our Drone config.

Instead, `.gpg`/`.asc` files should be placed into `/etc/apt/trusted.gpg.d/` or `/usr/share/keyrings`

Currently, `terraform` installs in Drone seem to fail.

jof:
Fixes #913

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -33,9 +33,9 @@ steps:
       TELEPORT_VERSION: 13.3.6
     commands:
       - echo Testing plugins against Teleport $TELEPORT_VERSION
-      - apt update && apt install -y software-properties-common
-      - curl -fsSL https://apt.releases.hashicorp.com/gpg | apt-key add -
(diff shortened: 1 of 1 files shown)


wadells (approved):
Code LGTM.  How did you test the change?

jof:
> Code LGTM. How did you test the change?

Mostly messing around in a local docker.
I suppose I could try and trigger drone on this branch and see what it does 🤔

jof:
This is less-broken, but it seems like there maybe a broken test that was hiding: https://drone.platform.teleport.sh/gravitational/teleport-plugins/2991/1/3

wadells:
FWIW: I'm good with merging "less broken".  I don't want to make "green" the enemy of "better".

Joerger (approved):

marcoandredinis:
There's also this PR which has some overlap
https://github.com/gravitational/teleport-plugins/pull/909

PR #921, opened 2023-09-21
ci: Build darwin tag builds on linux

camscale:
Build the tag builds of the terraform provider and event-handler for
darwin on linux. These are pure Go programs so can easily be
cross-compiled using GOOS/GOARCH. The "tricky" part is the universal
binary, solved by using https://github.com/konoui/lipo - a Go version of
lipo (the tool that constructs universal binaries) that can run on
Linux.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -512,13 +512,6 @@ kind: pipeline
 type: exec
 name: tag-build-terraform-darwin
 
-concurrency:
-  limit: 1
-
-platform:
-  os: darwin
-  arch: amd64
-
 trigger:
   event:
     - tag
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -512,13 +512,6 @@ kind: pipeline
 type: exec
```
wadells, line 512:
```suggestion
type: kuberentes
```

camscale: Done, for both pipelines

wadells:
I've tested this at:

* terraform https://drone.platform.teleport.sh/gravitational/teleport-plugins/3012/2/4
* event handler https://drone.platform.teleport.sh/gravitational/teleport-plugins/3011/3/1

Both look good.

I downloaded the universal terraform binary and checked it on my mac, as well as the amd64 eventhandler plugin:

```console
walt@work plugins % ./teleport-event-handler/teleport-event-handler version
Teleport event handler v14.0.0 go1.21.1
walt@work plugins % ./terraform-provider-teleport version                  
This binary is a plugin. These are not meant to be executed directly.
Please execute the program that consumes these plugins, which will
load any plugins automatically
```

wadells (approved):

camscale:
Closing this in favour of https://github.com/gravitational/teleport-plugins/pull/922

PR #922, opened 2023-09-21
ci: Combine linux and darwin tag pipelines

camscale:
Combine the pipelines for tag builds of linux and darwin builds of the
terraform provider and event-handler. These are pure Go programs so can
easily be cross-compiled using GOOS/GOARCH. The "tricky" part is the
universal binary, solved by using https://github.com/konoui/lipo - a Go
version of lipo (the tool that constructs universal binaries) that can
run on Linux.

This removes Apple Mac machines from the release pipeline of the
plugins.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -442,7 +442,7 @@ volumes:
 ---
 kind: pipeline
 type: kubernetes
-name: tag-build-terraform-linux
+name: tag-build-terraform
 
 trigger:
   event:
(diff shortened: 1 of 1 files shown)


wadells:
Would you link your testing?  I think that is the most important merge criteria.

wadells (approved):

camscale:
terraform: https://drone.platform.teleport.sh/gravitational/teleport-plugins/3014
event-handler: https://drone.platform.teleport.sh/gravitational/teleport-plugins/3013

camscale:
I have downloaded these builds from the S3 bucket and run them on my Mac (arm64). I needed to first remove the quarantine bit (`xattr -d com.apple.quarantine <binary>`) from the downloaded binaries, and they both execute fine and produced a help message (event-handler) or the "this is a plugin" message (terraform provider).

camscale:
I have also downloaded the linux versions and tested them in the same way - executing the binaries to ensure they can run. All good, same results:

```
> ./terraform-provider-teleport
This binary is a plugin. These are not meant to be executed directly.
Please execute the program that consumes these plugins, which will
load any plugins automatically
```

```
> ./teleport-event-handler version
Teleport event handler v14.0.0-camh.1 go1.21.1
```

r0mant (approved):

PR #923, opened 2023-09-21
ci: Remove mac exec pipelines

wadells:
We no longer have any runners capable of running mac exec pipelines, so these will hang forever.

The former runner `buildbox-intel` died, as documented in https://github.com/gravitational/teleport.e/issues/2275.

This PR depends on https://github.com/gravitational/teleport-plugins/pull/922

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -44,90 +44,6 @@ steps:
 
 ---
 kind: pipeline
-type: exec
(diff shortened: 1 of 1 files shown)


camscale (approved):

zmb3 (approved):
🪦

PR #915, opened 2023-09-15
Update access plugins image to latest base distroless image

MattiasAng:
The current image contains several critical vulnerabilities for OpenSSL, libc and libssl.

Trivy output on current image:

```
public.ecr.aws/gravitational/teleport-plugin-slack:13.3.8 (debian 11.2)

Total: 48 (UNKNOWN: 0, LOW: 12, MEDIUM: 18, HIGH: 11, CRITICAL: 7)

┌───────────┬──────────────────┬──────────┬───────────────────┬──────────────────┬──────────────────────────────────────────────────────────────┐
│  Library  │  Vulnerability   │ Severity │ Installed Version │  Fixed Version   │                            Title                             │
├───────────┼──────────────────┼──────────┼───────────────────┼──────────────────┼──────────────────────────────────────────────────────────────┤
│ libc6     │ CVE-2021-33574   │ CRITICAL │ 2.31-13+deb11u2   │ 2.31-13+deb11u3  │ mq_notify does not handle separately allocated thread        │
│           │                  │          │                   │                  │ attributes                                                   │
│           │                  │          │                   │                  │ https://avd.aquasec.com/nvd/cve-2021-33574                   │
│           ├──────────────────┤          │                   │                  ├──────────────────────────────────────────────────────────────┤
│           │ CVE-2022-23218   │          │                   │                  │ glibc: Stack-based buffer overflow in svcunix_create via     │
│           │                  │          │                   │                  │ long pathnames                                               │
│           │                  │          │                   │                  │ https://avd.aquasec.com/nvd/cve-2022-23218                   │
│           ├──────────────────┤          │                   │                  ├──────────────────────────────────────────────────────────────┤
│           │ CVE-2022-23219   │          │                   │                  │ glibc: Stack-based buffer overflow in sunrpc clnt_create via │
│           │                  │          │                   │                  │ a long pathname                                              │
│           │                  │          │                   │                  │ https://avd.aquasec.com/nvd/cve-2022-23219                   │
│           ├──────────────────┼──────────┤                   ├──────────────────┼──────────────────────────────────────────────────────────────┤
│           │ CVE-2021-3999    │ HIGH     │                   │ 2.31-13+deb11u4  │ Off-by-one buffer overflow/underflow in getcwd()             │
│           │                  │          │                   │                  │ https://avd.aquasec.com/nvd/cve-2021-3999                    │
│           ├──────────────────┼──────────┤                   ├──────────────────┼──────────────────────────────────────────────────────────────┤
[...]

PR #924, opened 2023-09-25
Update base docker images

marcoandredinis:
Buddy PR https://github.com/gravitational/teleport-plugins/pull/915

### DIFF
--- a/access/Dockerfile
+++ b/access/Dockerfile
@@ -22,7 +22,7 @@ RUN --mount=type=cache,target=/go/pkg/mod --mount=type=cache,target=/root/.cache
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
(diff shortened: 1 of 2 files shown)


strideynet (approved):

strideynet:
Does this also need adjusting in the main Teleport repo for the plugins that have been moved ?

marcoandredinis:
I don't think so.
These are only used for creating the docker images for the plugins.
The ones that were moved live within the teleport binary, and should have their own base docker image.

reedloden (commented):
Instead of doing specific hashes, can we just use `gcr.io/distroless/base-debian12`? Specifically, please use a variable at the top of the Dockerfile for easy updating.

Take a look at gravitational/teleport#31620.

reedloden:
> Instead of doing specific hashes, can we just use `gcr.io/distroless/base-debian12`? Specifically, please use a variable at the top of the Dockerfile for easy updating.

I submitted https://github.com/gravitational/teleport-plugins/pull/928 as an alternative to this PR.

tcsc:
We should probably close this in favour of @reedloden's #928.

reedloden:
Addressed in https://github.com/gravitational/teleport-plugins/pull/928.

PR #928, opened 2023-09-25
Update base docker images to use Debian 12 distroless `static` image

reedloden:
Instead of chasing ever-changing commit hashes, just use the apppropriate tag for the distroless image. This aligns things to how `teleport` is handled (https://github.com/gravitational/teleport/pull/31620).

Additionally, standardize on Debian 12 for everything (instead of a mix of 11 and 12). Also, use `static` over `base`. This means no `glibc` or `libssl`, which should be fine for these plugins.

Alternative to #924 and #915.

### DIFF
--- a/access/Dockerfile
+++ b/access/Dockerfile
@@ -1,7 +1,8 @@
 # Build the plugin binary
 ARG GO_VERSION
+ARG BASE_IMAGE=gcr.io/distroless/static-debian12
 
-FROM golang:${GO_VERSION}-bullseye as builder
+FROM golang:${GO_VERSION}-bookworm as builder
 
 ARG ACCESS_PLUGIN
 ARG GITREF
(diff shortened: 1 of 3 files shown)


tcsc (approved):

jakule (approved):

espadolini:
This broke `public.ecr.aws/gravitational/teleport-plugin-slack:14.0.1` because the binary inside is very much not static:
```
teleport-plugin: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, Go BuildID=sb62wjCeHYTZN85Ku79e/28AjXaQ3faKD7cS4g4oB/AANtv5x_00Dhqk0pNj65/Zzx2amhxFyifyIEYnEzk, stripped
```

PR #936, opened 2023-09-29
Compile plugins as static (`CGO_ENABLED=0`)

### LINKED ISSUE #935: Access plugin containers are broken in v14.0.1
After #928 the containers are based on distroless-static, but the binaries we put inside them are still built with CGO, resulting in a (confusing) `exec: /usr/local/bin/teleport-plugin: not found` error because the `/lib64/ld-linux-x86-64.so.2` interpreter is not in the image.

We should revert #928 or ensure that the build is indeed static.

reedloden:
The change to distroless static in #928 broke plugin binaries, as the distroless image doesn't have glibc. However, there's no reason for the plugins to need glibc, so just change them to all `CGO_ENABLED=0`.

Additionally, some minor fixes:
* Change one place where `GO_VERSION` wasn't properly set/updated.
* Correct build process for `access/email`.
* Remove reference to previously removed `access/example`.

Fixes #935.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,13 +44,9 @@ access-msteams:
 access-pagerduty:
 	$(MAKE) -C access/pagerduty
 
-.PHONY: access-example
-access-example:
-	go build -o build/access-example ./access/example
-
 .PHONY: access-email
 access-email:
-	go build -o build/access-email ./access/email
+	$(MAKE) -C access/email
 
 # Build specific access plugin with docker
 .PHONY: docker-build-access-%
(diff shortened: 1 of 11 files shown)


espadolini (approved):

Makefile
marcoandredinis, line 1: There's a target that calls `go build` directly instead of relying on the plugin's makefile
```makefile
.PHONY: access-email
access-email:
	go build -o build/access-email ./access/email
```

I don't think it's used anywhere, but we should use the same methods as we have for the other plugins
```makefile
.PHONY: access-email
access-email:
	$(MAKE) -C access/email
```

Would you mind adding this change as well?

marcoandredinis (approved):

Makefile
reedloden: Good catch. Fixed. Also deleted the reference to the non-existent `access/example` build as well.

marcoandredinis (approved):

tcsc (commented):
For what it's worth, LGTM

PR #937, opened 2023-09-29
Pin Teleport Terraform Provider to Teleport major version

tigrato:
The previous constrain `>= (=teleport.version=)` allows Terraform to pick any version bigger than the specified version. This is a problem because `>= 12.0.0` can actually select `14.0.0` which breaks our client guarantees when connecting to Auth server.

This PR forces Terraform to pick any version `>=12.0.0` and `< 13.0.0`. The pessimist version constrain prevents Terraform provider to use versions outside of the specified major version.


marcoandredinis (approved):

marcoandredinis:
You should also run `make terraform-gen-tfschema` so that we can generate the file which then should be copied to `teleport` repo.

strideynet (approved):

marcoandredinis (approved):

Joerger (approved):

PR #948, opened 2023-10-10
Relax mandatory FluentD's key and cert

tigrato:
There are cases where the upstream service uses other CA certificates and shares it with other components. When this happens, it's not possible to configure fluentd to use mTLS.

This commit relaxes the mandatory cert-key usage for fluentD.


r0mant (approved):

webvictim (approved):
<3

PR #965, opened 2023-11-02
Migrate from `terraform-plugin-sdk` to `terraform-plugin-testing`

strideynet:
> We recommend migrating to terraform-plugin-testing to take advantage of new features of the testing module and to avoid importing the SDKv2 for providers that are built on the plugin Framework.

https://developer.hashicorp.com/terraform/plugin/testing/migrating


strideynet:
https://github.com/hashicorp/terraform-plugin-sdk/pull/920

PR #963, opened 2023-10-21
fix leaking go-routines in event-handler watcher

miguelvalerio:
## Description
We had a misconfiguration that broke the communication between the `event-handler` and `teleport`, which in our case lead to the memory to quickly grow to unreasonable values.

Upon further inspection, we believe that this was due to the fact that the `Events` call is creating a go-routine that is not terminating when the last event on a page was processed and calling `fetch` returns an error. Since in our case the error returned to the `EventsJob` was considered a connetion problem, this means that `Events` would be called yet again, leading to go-routines that would not exit until the connetion issue was fixed being continuously spawned.

This PR attempts to fix that

## Additional Info
Memory:
image

Logs:
```
2023-10-12T21:35:46+01:00       Stack Trace:
2023-10-12T21:35:46+01:00       Original Error: *trace.TraceErr breaker is tripped
2023-10-12T21:35:46+01:00       ERROR REPORT:
2023-10-12T21:35:46+01:00       ERRO   Failed to connect to Teleport Auth server. Reconnecting... error:[
2023-10-12T21:35:46+01:00       ERRO   Error ingesting Audit Log err:breaker is tripped event-handler/events_job.go:96
2023-10-12T21:35:46+01:00       User Message: breaker is tripped] event-handler/events_job.go:69
2023-10-12T21:35:46+01:00               /usr/local/go/src/runtime/asm_amd64.s:1598 runtime.goexit
2023-10-12T21:35:46+01:00               /workspace/event-handler/teleport_events_watcher.go:281 main.(*TeleportEventsWatcher).Events.func1
2023-10-12T21:35:46+01:00               /workspace/event-handler/teleport_events_watcher.go:141 main.(*TeleportEventsWatcher).fetch
2023-10-12T21:35:46+01:00               /workspace/event-handler/teleport_events_watcher.go:213 main.(*TeleportEventsWatcher).getEvents
2023-10-12T21:35:46+01:00               /go/pkg/mod/github.com/gravitational/teleport/api@v0.0.0-20230825181643-42843879008e/client/client.go:2227 github.com/gravitational/teleport/api/client.(*Client).SearchUnstructuredEvents
2023-10-12T21:35:46+01:00       Stack Trace:
2023-10-12T21:35:46+01:00       Original Error: *trace.TraceErr breaker is tripped
2023-10-12T21:35:46+01:00       ERROR REPORT:
2023-10-12T21:35:46+01:00       ERRO   Failed to connect to Teleport Auth server. Reconnecting... error:[
[...]

PR #966, opened 2023-11-03
[Buddy] fix leaking go-routines in event-handler watcher

Joerger:
Buddy PR for https://github.com/gravitational/teleport-plugins/pull/963 with edits to resolve remaining comments.


event-handler/teleport_events_watcher.go
```diff
@@ -20,6 +20,10 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/gravitational/trace"
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/net/context"
```
zmb3, line 25: Wrong context import?

Joerger:
I gave this some more thought and realized this one line fix is not a viable solution. This change allows the event handler to drop incomplete pages of events and move on to the next page. Instead, we need to update the event handler to only spawn a single goroutine at a time, instead of spawning new goroutines for new pages when we are still trying to finish a known page. Drafting for now as I won't have time to make this change immediately.

Joerger:
I gave another pass to this fix, and found it works after all. I've updated the tests to give better coverage to the event collection logic and cover the edge case I was concerned about.

zmb3 (approved):

PR #968, opened 2023-11-07
Update go to 1.21.4

marcoandredinis:


camscale:
> docker.io/library/golang:1.21.4-bookworm: not found

Looks like this PR is a little early. The latest tags on library/golang are 5 days old as of this moment, so are only up to 1.21.3. We'll just have to keep an eye on the image tags and wait until it's ready.

marcoandredinis:
Just restarted the job, should build now

camscale (approved):

greedy52 (approved):

PR #973, opened 2023-11-27
feat: update pods deployment labels

thib-mary:
## Describe your changes

Update labels in the pod template spec of the deployment to include the full set of labels and not just the selector labels.

### DIFF
--- a/charts/access/discord/templates/deployment.yaml
+++ b/charts/access/discord/templates/deployment.yaml
@@ -16,7 +16,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       labels:
-        {{- include "discord.selectorLabels" . | nindent 8 }}
+        {{- include "discord.labels" . | nindent 8 }}
     spec:
       {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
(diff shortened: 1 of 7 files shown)


camscale:
@thib-mary Would you describe your changes a little more please? Describe the change as you see it and why this should be done. Something like:

> Change the labels in the pod template spec to include the full set of labels defined in the helm chart
> and not just the selector labels. This includes the help chart, app version and managed-by labels.
> These are needed because ...

The reason I ask is because it seems wrong to set the `managed-by` label to `helm` as helm does not manage the pod. Some of the other labels may make sense.

However if this change does make sense, it needs to be made to all the helm charts and not just for slack. I've asked for the team's input into why the labels might have been limited to just the selector labels. I'll update this when I find out more.

tigrato (approved):

tigrato:
Hi @thib-mary,

Thanks for your contribution. We are currently unable to run our tests on
external contributions. As such, I'll be helping you validate and
merge your pull request. I've created an internal PR here:

#1013

Please wait for another approver and I'll get your
code merged. You will retain authorship of your commits.

PR #978, opened 2023-12-04
Terraform Provider: Use updated APIs for Users, Roles, & Auth Connectors

rosstimothy:
The resources now all expose a Create, Update, Upsert RPC which returns the resource and an error. The generated code now uses Create for creating resources and Upsert for updating resources. The Update RPC is not used because it enforces optimistic locking which breaks the ownership model expected by the provider. Additionally, the GetUser RPC now takes a context so the hack to accommodate it has been removed.

### - Do Not Merge Until v15 -
The APIs used in this PR are not being backported to v14 so this cannot land until the `gravitational/teleport-plugins` repo is updated to point to branch/v15 of `gravitational/teleport`. go.mod was updated in this PR to point at a commit of master from `gravitational/teleport` so that the functionality could be validated.


marcoandredinis (commented):
LGTM
Did you run this locally against a v15/master build?
The tests fail because we are probably using a v14.x version on the CI/CD

strideynet:
If we're switching to the APIs which return the created/updated resource, we can probably remove the logic which tries to Get the resource after the create/update.

hugoShaka (approved):

marcoandredinis (approved):

r0mant (approved):

Issue #985, opened 2023-12-15
Use of `subPath` when mounting Teleport identity secret in access plugins helm charts breaks reload


strideynet:
> Note: A container using a Secret as a [subPath](https://kubernetes.io/docs/concepts/storage/volumes#using-subpath) volume mount does not receive automated Secret updates.

This makes the charts not work great with the rotating credentials sourced from Machine ID.

PR #986, opened 2023-12-15
Helm: Improve compatibility with Machine ID

strideynet:
Closes https://github.com/gravitational/teleport-plugins/issues/985

## No longer use `subPath` when mounting the identity file secret

To use Machine ID with plugins, the identity file changes regularly. The use of `subPath` prevents this (from the k8s docs):

> Note: A container using a Secret as a [subPath](https://kubernetes.io/docs/concepts/storage/volumes#using-subpath) volume mount does not receive automated Secret updates.

## Enable credential watching by default

Provides `refresh_identity` by default to the configuration for plugins. There shouldn't be any adverse affects if Machine ID isn't being used.


strideynet:
I've tested the Discord access plugin locally with both Machine ID and an old style secret containing a one-off identity file as described by the docs.

hugoShaka (approved):

tigrato (approved):

zmb3 (approved):

PR #1002, opened 2024-01-11
Release 14.3.1

camscale:


go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/alecthomas/kong v0.2.22
 	github.com/go-resty/resty/v2 v2.11.0 // indirect
 	github.com/gogo/protobuf v1.3.2
-	github.com/google/uuid v1.3.1
```
fheinecke, line 12: Are the changes to this file intentional? This doesn't usually get changed on release

fheinecke (approved):

tigrato (approved):

go.mod
camscale: Huh? It's been changed on at least the last 5 releases. Go code gets updated as I understand it as the terraform provider has parts generated which also depends on the teleport API, so we probably get mod changes from there too.

I just ran the usual `make update-version VERSION=...` and raised the PR on its changes.

fheinecke: Ah sorry, I've just missed that on previous PRs. There are enough go.mod/go.sum changes in this one that it really stood out.

camscale: Most of it seems to be in the second `requires` section, which are the indirect dependencies. Something we depend on depends on a whole lot less!

PR #1007, opened 2024-01-16
Terraform provider: make version mandatory and immutable

hugoShaka:
This is a breaking change, merge this once we know we won't do any new v14 release.

To mitigate https://github.com/gravitational/teleport/issues/36773 we will enforce the resource version to be set starting with Teleport 15. This will ensure no surprise upgrade happens.

The documentation PR is here: https://github.com/gravitational/teleport/pull/36774

Note: protos from the docs PR were used to generate the schema in order to have the nice version messages and render a proper reference.


terraform/gen/referencedocs.go.tpl
```diff
@@ -67,6 +67,22 @@ provider "teleport" {
...
+by a resource version increase as version is used to set defaults on creation.
+However those changes will take effect as soon as Terraform will need to update
+the resource for any other reason.
+
+Since Teleport 15, you MUST set the version on each resource. This is not
+enforced on previous Teleport provider versions but we highly recommend doing
+so. If it's not specified, Terraform will pick the latest one by default. When
+this happens, you will face unexpected changes when doing major Terraform
+provider upgrades, potentially introducing new resource versions
+and changing the behaviour of your existing TF code.
```
tigrato, line 84: Worth mentioning that when dealing with roles, it can impact access rules when upgrading the provider across major versions

terraform/test/role_test.go
```diff
@@ -45,7 +45,7 @@ func (s *TerraformSuite) TestRole() {
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr(name, "kind", "role"),
 					resource.TestCheckNoResourceAttr(name, "spec.options"),
-					resource.TestCheckResourceAttr(name, "version", "v6"),
+					resource.TestCheckResourceAttr(name, "version", "v7"),
```
tigrato, line 48: Can we add a test that if we set the version on v6, it's kept as v6

terraform/tfschema/devicetrust/v1/device_terraform.go
```diff
@@ -174,11 +127,9 @@ func GenSchemaDeviceV1(ctx context.Context) (github_com_hashicorp_terraform_plug
 			Optional:    true,
 		},
 		"version": {
-			Computed:      true,
-			Description:   "Version is version",
-			Optional:      true,
-			PlanModifiers: []github_com_hashicorp_terraform_plugin_framework_tfsdk.AttributePlanModifier{github_com_hashicorp_terraform_plugin_framework_tfsdk.UseStateForUnknown()},
-			Type:          github_com_hashicorp_terraform_plugin_framework_types.StringType,
+			Description: "Version is the API version used to create the resource. It must be specified. Based on this version, Teleport will apply different defaults on resource creation or deletion. It ust be an integer prefixed by \"v\". For example: `v1`",
```
marcoandredinis, line 130: There's a typo at
`It ust be an` -> `It must be an`

I sent a comment to the [docs PR](https://github.com/gravitational/teleport/pull/36774)

marcoandredinis (commented):
I think we need to add specific tests for this

We should also have a docs page or an anchor to a doc that explains the change that users must do

terraform/gen/referencedocs.go.tpl
webvictim: I don't want to sound like a broken record but IMO we do also need some central reference for what role versions can be used, what is current, what the difference is between them, etc. We'll probably get a lot of questions about "what role version should I use" otherwise (more than we usually get)

hugoShaka: I wrote the tests, and I have bad news: we built a system with hysteresis 😭 .

A role upgraded to v6 from v5, and a role created from v6 with the same spec won't have the same rules.

Resources have their defaults set on initial creation, then never again. If you create a resource v5, you will have the v5 defaults, even after you upgrade to v6.

There's an interaction between Teleport resources without clear field ownership and how Terraform measures the existing state to do drift detection. Teleport edits the spec server-side, losing the original user intention. Then TF imports this amended resource back in its state, importing all the defaults and keeping them ad vitam.

The conclusion is that upgrading a resource between versions is not safe. As re-creating it from scratch will not lead to the same result. This also explains why so few people complained about their role being broken after upgrading to Teleport 13.

marcoandredinis (approved):

tigrato (approved):
We will need to somehow get a workaround for this otherwise for users that want to use Kubernetes resources other than pod, it will involve deleting several roles so the conversion can happen.
For now it will be fine but prepare for some support load since some kubernetes resources are a v7 only feature

strideynet (approved):

hugoShaka:
> We will need to somehow get a workaround for this otherwise for users that want to use Kubernetes resources other than pod, it will involve deleting several roles so the conversion can happen. For now it will be fine but prepare for some support load since some kubernetes resources are a v7 only feature

You're right. I will document how to upgrade a resource (create a new role v7, use it everywhere, remove rolev5). I also need to document best practices to dynamically use the correct role name and have Terraform correctly infer dependencies.

r0mant (approved):

Issue #1010, opened 2024-01-23
terraform provider loads `~/.tsh` profile creds by default


GavinFrazar:
### Expected behavior:

I set my provider config to use an identity file generated by `tbot`:
```terraform
provider "teleport" {
  # Update addr to point to your Teleport Cloud tenant URL's host:port
  addr               = "gavin-leaf.cloud.gravitational.io:443"
  identity_file_path = "./tbot-user/identity"
}
```

I expected `terraform apply` to use this credential for every API interaction.

I do not expect my `~/.tsh` profile to have any effect on the the terraform provider unless I've explicitly configured `profile_dir` or used `TF_TELEPORT_PROFILE_DIR` env var.

### Actual behavior

`terraform apply` loads credentials from BOTH the identity file and from `~/.tsh` if it exists (tsh profile).

Then the api client attempts to connect to the Teleport cluster using both credential sources simultaneously.

As a result, sometimes `terraform <cmd>` succeeds, and sometimes it would get an access denied error message, depending on which credential source "won" the connection race.

This was both extremely surprising and difficult to debug. I had to enable `TF_LOG=DEBUG` and search for the string `Using profile as the default auth method` (after reading our provider's source) to see it happening, as I noticed the intermittent failures only happened when I was logged into tsh.

PR #1011, opened 2024-01-23
terraform provider does not load `~/.tsh` creds by default

GavinFrazar:
Fixes https://github.com/gravitational/teleport-plugins/issues/1010

### DIFF
--- a/terraform/provider/provider.go
+++ b/terraform/provider/provider.go
@@ -300,7 +300,10 @@ func (p *Provider) Configure(ctx context.Context, req tfsdk.ConfigureProviderReq
 	}
 
 	if profileDir != "" || len(creds) == 0 {
-		log.WithField("dir", profileDir).WithField("name", profileName).Debug("Using profile as the default auth method")
+		log.WithFields(log.Fields{
+			"dir":  profileDir,
+			"name": profileName,
+		}).Debug("Using profile as the default auth method")
 		creds = append(creds, client.LoadProfile(profileDir, profileName))
 	}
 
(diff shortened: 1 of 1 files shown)


terraform/provider/provider.go
```diff
@@ -299,8 +299,10 @@ func (p *Provider) Configure(ctx context.Context, req tfsdk.ConfigureProviderReq
 		creds = append(creds, client.LoadIdentityFileFromString(string(decoded)))
 	}
 
-	log.WithField("dir", profileDir).WithField("name", profileName).Debug("Using profile as the default auth method")
-	creds = append(creds, client.LoadProfile(profileDir, profileName))
+	if profileDir != "" {
```
Joerger, line 302: We should still default to profile if no creds were provided. 
```suggestion
	if profileDir != "" || len(creds) == 0 {
```

terraform/provider/provider.go
```diff
@@ -299,8 +299,10 @@ func (p *Provider) Configure(ctx context.Context, req tfsdk.ConfigureProviderReq
 		creds = append(creds, client.LoadIdentityFileFromString(string(decoded)))
 	}
 
-	log.WithField("dir", profileDir).WithField("name", profileName).Debug("Using profile as the default auth method")
-	creds = append(creds, client.LoadProfile(profileDir, profileName))
+	if profileDir != "" || len(creds) == 0 {
+		log.WithField("dir", profileDir).WithField("name", profileName).Debug("Using profile as the default auth method")
```
tcsc, line 303: While you're already editing this:
```suggestion
		log.
			WithFields(log.Fields{
				"dir":  profileDir,
				"name": profileName,
			}).
			Debug("Using profile as the default auth method")
```

tcsc (approved):

GavinFrazar:
@zmb3 @Joerger PTAL

zmb3 (approved):

Joerger (approved):

Issue #1016, opened 2024-02-02
Terraform provider should verify the Teleport cluster version before applying


codingllama:
The idea is to avoid situations where users are running a newer (or just mismatch) provider against their cluster, which can have unforeseen and possibly destructive consequences.

PR #1017, opened 2024-02-02
bump teleport provider min version

hugoShaka:
Teleport provider must not connect to old Teleport version as they won't support newer RPCs.

Fixes https://github.com/gravitational/teleport-plugins/issues/1016

### DIFF
--- a/terraform/provider/provider.go
+++ b/terraform/provider/provider.go
@@ -42,7 +42,7 @@ import (
 
 const (
 	// minServerVersion is the minimal teleport version the plugin supports.
-	minServerVersion = "6.1.0-beta.1"
+	minServerVersion = "15.0.0-0"
 )
 
 type RetryConfig struct {


terraform/provider/provider.go
```diff
@@ -42,7 +42,7 @@ import (
 
 const (
 	// minServerVersion is the minimal teleport version the plugin supports.
-	minServerVersion = "6.1.0-beta.1"
+	minServerVersion = "15.0.0-0"
```
codingllama, line 45: Could we add this to a release checklist so we don't forget to bump the number in the future?

codingllama, line 45: What is the "-0" for? Is it to avoid maching betas/RCs?

codingllama (commented):
Thanks for the quick fix, Hugo!

r0mant (approved):

tigrato (approved):

terraform/provider/provider.go
hugoShaka: it is to match the beta/rc. Go's semver will consider 15.0.5-dev below 15.0.0 if you don't explicitly add the `-0` part

hugoShaka: I'll check with roman, I'm not sure where to put this but we definitely want to have this in our major release process

codingllama: Thanks.

PR #1019, opened 2024-02-06
add terraform support for Teleport servers

hugoShaka:
This PR implements: https://github.com/gravitational/teleport/issues/37624


terraform/tfschema/validators.go
```diff
@@ -228,3 +229,48 @@ func (v AnyOfValidator) Validate(ctx context.Context, req tfsdk.ValidateAttribut
...
+
+	value, ok := req.AttributeConfig.(types.String)
+	if !ok {
+		resp.Diagnostics.AddError("Version validation error", fmt.Sprintf("Attribute %v can not be converted to StringValue", req.AttributePath.String()))
+		return
+	}
+
+	if value.Null || value.Unknown {
+		if !slices.Contains(v.AllowedValues, "") {
+			resp.Diagnostics.AddError("Field validation error", fmt.Sprintf("Attribute %v (%v) is unset but empty string is not a valid value. (vXX)", req.AttributePath.String(), value.Value))
```
marcoandredinis, line 267: What does `(vXX)` stand for? Or was it a typo?

terraform/tfschema/validators.go
```diff
@@ -228,3 +229,48 @@ func (v AnyOfValidator) Validate(ctx context.Context, req tfsdk.ValidateAttribut
...
+
+// Validate performs the validation.
+func (v StringValueValidator) Validate(_ context.Context, req tfsdk.ValidateAttributeRequest, resp *tfsdk.ValidateAttributeResponse) {
+	if req.AttributeConfig == nil {
+		return
+	}
+
+	value, ok := req.AttributeConfig.(types.String)
+	if !ok {
+		resp.Diagnostics.AddError("Version validation error", fmt.Sprintf("Attribute %v can not be converted to StringValue", req.AttributePath.String()))
```
marcoandredinis, line 261: I'm guessing this `Version` was copy-pasted from another validator.

terraform/example/server.tf.example
```diff
@@ -0,0 +1,11 @@
+resource "teleport_server" "ssh_agentless" {
+    version = "v2"
+    sub_kind = "openssh"
+    metadata = {
+        name = "test"
```
marcoandredinis, line 5: Did you test this?

I was testing for the `openssh-ec2-ice` nodes and it doesn't work if we give it a non-uuid string.

An example that works:
```tf
resource "random_uuid" "ec2-random-name" {
}

resource "teleport_server" "my-dev-server" {
  version  = "v2"
  sub_kind = "openssh-ec2-ice"
  metadata = {
    name = "${random_uuid.ec2-random-name.result}"
  }
  spec = {
    addr     = "<ec2 private ip>:22"
    hostname = "my-dev-server"
    cloud_metadata = {
      aws = {
        account_id  = "<account-id>"
        instance_id = "<instance-id>"
        region      = "<region>"
        vpc_id      = "<vpc-id>"
        integration = "<awsoidc-integration-used-for-credentials>"
        subnet_id   = "<subnet-id>"
      }
    }
  }
}

```

hugoShaka: > Did you test this?
> 
> I was testing for the openssh-ec2-ice nodes and it doesn't work if we give it a non-uuid string.

Do you mean against a real AWS account? No, I only checked that Teleport was accepting the resource in the operator. I don't understand why the eice server name must be a uuid 🤔

marcoandredinis: It's a bit buried, but it is required to be an uuid because of this
https://github.com/gravitational/teleport/blob/a39c21716b40518f0433b619bb9828d127bd6b77/api/utils/route.go#L41

When I implemented EC2 ICE mode, I closely followed the OpenSSH/Agentless mode.
And I ended up making the same assumptions.

tigrato: I guess you shouldn't set the name. Instead, you should set the hostid

hugoShaka: I updated the TF example to not set the name, and added tests for both named and nameless openssh servers.

terraform/Makefile
```diff
@@ -117,7 +117,7 @@ ifeq ($(shell expr $(CURRENT_ULIMIT) \< 1024), 1)
 	@echo "ulimit -n is too low ($(CURRENT_ULIMIT)), please set ulimit -n 1024"
 	@exit -1
 endif
-	TF_ACC=true go test ./test -v
+	TF_ACC=true go test ./test -v # -run '^\QTestTerraform\E$$/^\QTestNamelessOpenSSHServer\E$$'
```
hugoShaka, line 120: TODO: remove

terraform/gen/plural_resource.go.tpl
```diff
@@ -116,9 +119,22 @@ func (r resourceTeleport{{.Name}}) Create(ctx context.Context, req tfsdk.CreateR
...
+{{- end}}
+
+{{if .HasCheckAndSetDefaults -}}
+	err = {{.VarName}}Resource.CheckAndSetDefaults()
+	if err != nil {
+	resp.Diagnostics.Append(diagFromWrappedErr("Error setting {{.Name}} defaults", trace.Wrap(err), "{{.Kind}}"))
+	return
+	}
+{{- end}}
+
```
hugoShaka, line 134: `CheckAndSetDefaults` and kind setting were moved before

```
id := {{.VarName}}Resource.Metadata.Name
```

because the name itself can be changed during this step (e.g. nameless openssh servers).

terraform/gen/main.go
```diff
@@ -108,6 +108,13 @@ type payload struct {
 	// existing resource when we're updating it. For example:
 	// "Spec.Audit.NextAuditDate" in AccessList resource
 	PropagatedFields []string
+	// Namespaced indicates that the resource get and delete methods need the
+	// deprecated namespace parameter (always the default namespace).
+	Namespaced bool
+	// ForceSetKind indicates that the resource kind must be forcefully set by the provider.
+	// This is required for weird resources (ServerV2) that support multiple kinds.
```
marcoandredinis, line 115: Don't call them weird, you might hurt their feelings

terraform/provider/resource_teleport_access_list.go
```diff
@@ -81,6 +81,13 @@ func (r resourceTeleportAccessList) Create(ctx context.Context, req tfsdk.Create
 		resp.Diagnostics.Append(diagFromWrappedErr("Error reading AccessList", trace.Errorf("Can not convert %T to AccessList: %s", accessListResource, err), "access_list"))
 		return
 	}
+
+err = accessListResource.CheckAndSetDefaults()
+	if err != nil {
+	resp.Diagnostics.Append(diagFromWrappedErr("Error setting AccessList defaults", trace.Wrap(err), "access_list"))
+	return
+	}
+
```
marcoandredinis, line 90: This is not gofmt'd

terraform/example/server.tf.example
marcoandredinis, line 1: Can we add an eice server as well?

terraform/protoc-gen-terraform-teleport.yaml
```diff
@@ -418,6 +455,10 @@ validators:
     - UseVersionBetween(2,2)
   SAMLConnectorV2.Spec:
     - UseAnyOfValidator("entity_descriptor", "entity_descriptor_url")
+  ServerV2.Version:
+    - UseVersionBetween(2,2)
+  ServerV2.SubKind:
+    - UseValueIn([]string{"openssh", "openssh-ec2-ice"})
```
marcoandredinis, line 461: :thinking: Do we need to use `[]string`?
For the `UseAnyOfValidator` we only need to pass the actual strings.

terraform/test/server_test.go
```diff
@@ -0,0 +1,305 @@
+/*
+Copyright 2015-2021 Gravitational, Inc.
```
marcoandredinis, line 2:
```suggestion
Copyright 2024 Gravitational, Inc.
```

terraform/example/server.tf.example
```diff
@@ -0,0 +1,43 @@
...
+resource "teleport_server" "ssh_agentless_eice" {
+  version = "v2"
+  sub_kind = "openssh-ec2-ice"
+  metadata = {
+    // It is recommended to put the instance ID as a name for EC2 Instance Connect
+    // When dialing to this instance, teleport will detect that this is an
+    // AWS instance ID an will contact this specific instance. This is more
+    // robust than host-based and IP-based dialing (because several server
+    // can have similar hostnames).
+    name = "i-0123456789abcdef"
```
marcoandredinis, line 26: I don't think that's the correct format
https://github.com/gravitational/teleport/blob/1c03445f51dec3ce8cef4141d705e92d10cebeb9/api/utils/aws/ec2.go#L34

```suggestion
    // It is recommended to put the account and instance ID as a name for EC2 Instance Connect
    // When dialing to this instance, teleport will detect that this is an
    // AWS instance ID an will contact this specific instance. This is more
    // robust than host-based and IP-based dialing (because several server
    // can have similar hostnames).
    name = "123456789012-i-0123456789abcdef"
```

marcoandredinis (approved):
Just left a comment

tigrato (approved):

r0mant (approved):
Bot.

PR #1040, opened 2024-03-20
CI: Update teleport version to 15.2.0

marcoandredinis:
Updates Teleport binary used in tests to 15.2.0

This new version of teleport has the following relevant changes:

### Teleport process logging uses `log/slog`
For terraform tests we start a teleport binary and parse its output to understand when the Auth/Proxy service started and what ports are they listening on.
We used a regex for that, but teleport migrated to `log/slog` and the regex no longer works.
Migration PR: https://github.com/gravitational/teleport/pull/38551
We had to fix the regex for integration tests in teleport: https://github.com/gravitational/teleport/pull/39315
Terraform Tests also use that library, so after the regex changed, we must upgrade Teleport CI version to get the new log format.

### Teleport API: `GetClusterNetworkingConfig` and `GetSessionRecordingConfig` never return a nil

When developing the `ClusterMaitenanceConfig` we had to include a nil check, because if it was never configured, `GetClusterMaintenanceConfig` would return a nil object.
This nil check was added to all SingleResource resources.

For `ClusterNetworkingConfig` and `SessionRecordingConfig`, the `Get` operation never returns a nil resource and `staticcheck` linter was yelling about it.
So, we had to create a new flag to ensure we only nil-checked the resources that can actually return a nil value.

### Teleport Resource Metadata
It is no longer recommended to use the `<Resource>.Metadata.ID` to check for cached responses.
We are now using the revision field.

During this change we also detected a miss-usage of an `error` variable and fixed that (could lead to a panic).

LoginRules didn't have the `Revision` field, so we added it here: https://github.com/gravitational/teleport.e/pull/3821

Unfortunately, that PR didn't merge in time for 15.2.0.
However, that's ok because LoginRules are not cached.
So, instead of waiting for a new release (15.2.1), we just removed the cache check.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,7 @@ require (
 	github.com/peterbourgon/diskv/v3 v3.0.1
 	github.com/sethvargo/go-limiter v0.7.2
 	github.com/sirupsen/logrus v1.9.3
-	github.com/stretchr/testify v1.8.4
+	github.com/stretchr/testify v1.9.0
 	github.com/tidwall/gjson v1.14.4
 	golang.org/x/exp v0.0.0-20231108232855-2478ac86f678
 	golang.org/x/net v0.21.0
@@ -287,7 +287,7 @@ require (
 	github.com/spf13/cast v1.5.1 // indirect
 	github.com/spf13/cobra v1.8.0 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
-	github.com/stretchr/objx v0.5.0 // indirect
+	github.com/stretchr/objx v0.5.2 // indirect
 	github.com/thales-e-security/pool v0.0.2 // indirect
 	github.com/tidwall/match v1.1.1 // indirect
 	github.com/tidwall/pretty v1.2.1 // indirect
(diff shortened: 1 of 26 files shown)


terraform/Makefile
```diff
@@ -103,7 +103,7 @@ release: build
 endif
 	tar -C $(BUILDDIR) -czf $(RELEASE).tar.gz .
 
-TERRAFORM_EXISTS := $(shell terraform -version 2>/dev/null | grep 'Terraform v1.5')
+TERRAFORM_EXISTS := $(shell terraform -version 2>/dev/null | grep 'Terraform v1.')
```
hugoShaka, line 106: Is this change intentional? Do we want to force a specific TF version or we're OK with everything 1.x?

hugoShaka (approved):

terraform/Makefile
marcoandredinis: Yes, but I can add it into another PR if you think it's better

I think requiring anything in the `1.x` range is fine, and I was getting annoyed with having to upgrade multiple things

go.mod
```diff
@@ -364,8 +364,8 @@ replace (
 	github.com/alecthomas/kingpin/v2 => github.com/gravitational/kingpin/v2 v2.1.11-0.20230515143221-4ec6b70ecd33
 	github.com/coreos/go-oidc => github.com/gravitational/go-oidc v0.1.1
 	github.com/gogo/protobuf => github.com/gravitational/protobuf v1.3.2-0.20201123192827-2b9fcfaffcbf
-	github.com/gravitational/teleport => github.com/gravitational/teleport v0.0.0-20240327020503-fc3b2b31dec6 // ref: tags/v15.1.10
-	github.com/gravitational/teleport/api => github.com/gravitational/teleport/api v0.0.0-20240327020503-fc3b2b31dec6 // ref: tags/v15.1.10
+	github.com/gravitational/teleport => github.com/gravitational/teleport v0.0.0-20240329210410-bb8bd77625f4 // ref: tags/v15.2.0
+	github.com/gravitational/teleport/api => github.com/gravitational/teleport/api v0.0.0-20240329210410-bb8bd77625f4 // ref: tags/v15.2.0
```
marcoandredinis, line 368: @hugoShaka What do you think about this?
I'm upgrading to 15.2.0 here so we get the new regex used in integration tests
But, usually, only release PRs change this

hugoShaka: I do this all the time to make the tests pass :)

The release PR will overwrite this anyway.

marcoandredinis:
@r0mant Can you please take a look? I need a code owners review

r0mant (approved):

PR #1042, opened 2024-03-20
Relax Terraform provider version re-creation condition

hugoShaka:
Fixes https://github.com/gravitational/teleport/issues/39638

This PR keeps the mandatory version attribute but does not trigger re-creation when it changes like the v14 provider was doing. We might enable re-creation back in a future version.

Relaxing the re-creation is safe on any non-computed field. So basically every resource but the role with kubernetes_resource set.

This change is required because it's blocking people from upgrading to v15. I will investigate other fixes for the roles with Kubernetes resources attrs.

### DIFF
--- a/terraform/protoc-gen-terraform-accesslist.yaml
+++ b/terraform/protoc-gen-terraform-accesslist.yaml
@@ -59,15 +59,6 @@ plan_modifiers:
     Metadata.name:
       - "github.com/hashicorp/terraform-plugin-framework/tfsdk.RequiresReplace()"
 
-    # Version MUST NOT change. Due to the way Terraform imports back the resource
-    # in its state (the provider relies on `USeStateForUnknown`) and the fact
-    # Teleport mixes the injected defaults with the original resource, defaults
-    # are set only on resource creation. Upgrading an existing resource will
-    # create a hybrid: a resource with the new version but old defaults.
-    AccessList.header.version:
-      - "github.com/hashicorp/terraform-plugin-framework/tfsdk.UseStateForUnknown()"
-      - "github.com/hashicorp/terraform-plugin-framework/tfsdk.RequiresReplace()"
-
 # This must be defined for the generator to be happy, but in reality all time
 # fields are overridden (because the protobuf timestamps contain locks and the
 # linter gets mad if we use raw structs instead of pointers).
--- a/terraform/protoc-gen-terraform-devicetrust.yaml
+++ b/terraform/protoc-gen-terraform-devicetrust.yaml
@@ -60,14 +60,6 @@ plan_modifiers:
     # Force to recreate resource if asset tag changes
     "DeviceV1.spec.asset_tag":
       - "github.com/hashicorp/terraform-plugin-framework/tfsdk.RequiresReplace()"
-    # Version MUST NOT change. Due to the way Terraform imports back the resource
-    # in its state (the provider relies on `USeStateForUnknown`) and the fact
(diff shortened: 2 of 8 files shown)


marcoandredinis (approved):

wadells:
For context, I was encountering the error where upgrading an accesslist to v15 required adding `version` but that required addition was triggering a recreate.  For resources like access lists that may be only partially terraformed this recreate is painful.

```
│ Error: Incorrect attribute value type
│ 
│   on access-lists.tf line 56, in resource "teleport_access_list" "access_list_reviewers":
│   56:   header = {
│   57:     metadata = {
│   58:       name = "${each.value.name}-reviewer"
│   59:     }
│   60:   }
│ 
│ Inappropriate value for attribute "header": attribute "version" is
│ required.
```

However, when I added the version (`v1`, pulled from `tctl`), I saw the access list s getting recreated.

```
# teleport_access_list.access_list["XYZ"] must be replaced
-/+ resource "teleport_access_list" "access_list" {
      ~ header = {
          ~ metadata = {
                name      = "XYZ"
              ~ namespace = "" -> (known after apply)
            }
          + version  = "v1" # forces replacement
        }
      ~ id     = "XYZ" -> (known after apply)
        # (1 unchanged attribute hidden)
    }
```

wadells (approved):
I don't know the code well enough to provide relevant feedback, but this has my approval in spirit.

tigrato (approved):

hugoShaka:
The tests are broken because of the relaxed condition. As this PR is already approved, I wrote a fix in another PR: https://github.com/gravitational/teleport-plugins/pull/1043

terraform/tfschema/role_plan_modifier.go
```diff
@@ -0,0 +1,108 @@
...
+	"fmt"
+
+	apitypes "github.com/gravitational/teleport/api/types"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	tftypes "github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+const (
+	DefaultRoleKubeModifierErrSummary  = "DefaultRoleKubeResources modifier failed"
+	DefaultRoleKubeModifierDescription = `This modifier re-render the role.spec.allow.kubernetes_resources from the user provided config instead of using the state.
```
camscale, line 30: s/re-render/re-renders/ or s/re-render/will re-render/

s/user provided/user-provided/

terraform/tfschema/role_plan_modifier.go
```diff
@@ -0,0 +1,108 @@
...
+		return
+	}
+	spec, ok := specRaw.(tftypes.Object)
+	if !ok {
+		resp.Diagnostics.AddError(DefaultRoleKubeModifierErrSummary, "Failed to cast 'spec' as a TF object.")
+		return
+	}
+	allowRaw, ok := spec.Attrs["allow"]
+	if !ok {
+		resp.Diagnostics.AddError(DefaultRoleKubeModifierErrSummary, "Failed to get 'spec' from TF object.")
```
camscale, line 94: s/spec/allow/ in error message

camscale (approved):
Just a few wording changes, but otherwise LGTM.

hugoShaka:
I spoke with @r0mant and we don't want to fix the role bug in v15 as this might be breaking for a small amount of users who already had roles upgraded from previous versions before reaching v15).

I will push the plan modifier in v16 and skip the broken test for now.

PR #1043, opened 2024-03-21
Ensure kube_resources always have the correct defaults, even after version change

hugoShaka:
image

This PR is required because relaxing the RequireReplace() causes the Kubernetes resource defaults bug to reappear.

For context:
- the kubernetes_resource field can be empty for the user
- the server and provider will set defaults
- TF will get mad if it's not aware that this can happen, we can tell TF by marking the field as "computed"
- When a field is computed, TF will store its result back into the state
- The state is then used for future reconciliations
- If you happen to change the role version, the defaults should change, but TF will not be aware and use the old defaults from the state
- You will have a new role with older defaults, causing inconsistencies and potentially granting too much access

We tried to solve this issue by marking the resource for replacement on version change, but this proved to be super disruptive. Not all resources can be deleted easily. AccessLists contain information about its members, and roles might have dependencies (you cannot delete a role if a user or an AL is referencing it, SSO are dynamic and we cannot remove them with TF).

This PR proposes a more advanced solution. Instead of using the TF state, we re-compute the defaults based on the configuration provided in the TF code. This is done with a Plan modifier.

### DIFF
--- a/terraform/tfschema/role_plan_modifier.go
+++ b/terraform/tfschema/role_plan_modifier.go
@@ -1,0 +1,91 @@
+package tfschema
+
+import (
+	"context"
+	"fmt"
+	apitypes "github.com/gravitational/teleport/api/types"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	tftypes "github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+const (
+	DefaultRoleKubeModifierErrSummary  = "DefaultRoleKubeResources modifier failed"
+	DefaultRoleKubeModifierDescription = `This modifier re-render the role.spec.allow.kubernetes_resources from the user provided config instead of using the state.
+The state contains server-generated defaults (in fact they are generated in the pre-apply plan).
+However, those defaults become outdated if the version changes.
+One way to deal with version change is to force-recreate, but this is too destructive.
+The workaround we found was to use this plan modifier.`
+)
+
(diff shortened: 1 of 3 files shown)


tigrato (approved):

terraform/tfschema/role_plan_modifier.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (d DefaultRoleKubeResourceModifier) Description(ctx context.Context) string {
+	return DefaultRoleKubeModifierDescription
+}
+
+func (d DefaultRoleKubeResourceModifier) MarkdownDescription(ctx context.Context) string {
+	return DefaultRoleKubeModifierDescription
+}
+
+func (d DefaultRoleKubeResourceModifier) Modify(ctx context.Context, req tfsdk.ModifyAttributePlanRequest, resp *tfsdk.ModifyAttributePlanResponse) {
```
marcoandredinis, line 52: Can we add godoc at least for this one?

marcoandredinis (commented):
Can we please add tests?

hugoShaka:
> Can we please add tests?

This is tested by `TestTerraform/TestRoleVersionUpgrade`. This is the reason why https://github.com/gravitational/teleport-plugins/pull/1042 tests are failing and cannot be merged: https://github.com/gravitational/teleport-plugins/actions/runs/8365719432/job/22904223261?pr=1042#step:6:379

marcoandredinis (approved):

Tarun5139:
Hello Team

PR #1044, opened 2024-03-28
Release 15.1.10

tcsc:


camscale (approved):

fheinecke:
@tcsc This release will definitely fail as is. See the Build release workflow output. This workflow is now the same for releases as it is for PRs, so if it fails on a PR then the release will also fail.

fheinecke (approved):

PR #1047, opened 2024-04-03
use the new bot API while keeping a similar behaviour

hugoShaka:
We removed the old CreateBot RPC on master, the provider doesn't compile anymore.

This PR is a quick fix: it uses the new RPC to create the bot. The long term fix is doing a breaking change and using the new bot resource correctly (generating from protobuf/go structs instead of writing the provider manually).

I need the Terraform provider to compile on master because I want to copy it there to get rid of the `teleport-plugins` repo.

### DIFF
--- a/terraform/provider/resource_teleport_bot.go
+++ b/terraform/provider/resource_teleport_bot.go
@@ -18,10 +18,10 @@ package provider
 
 import (
 	"context"
-	"time"
-
-	"github.com/gravitational/teleport/api/client/proto"
-	"github.com/gravitational/teleport/api/types/wrappers"
+	"fmt"
+	headerv1 "github.com/gravitational/teleport/api/gen/proto/go/teleport/header/v1"
+	machineidv1 "github.com/gravitational/teleport/api/gen/proto/go/teleport/machineid/v1"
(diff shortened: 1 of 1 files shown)


terraform/provider/resource_teleport_bot.go
```diff
@@ -60,30 +60,15 @@ func GenSchemaBot(ctx context.Context) (tfsdk.Schema, diag.Diagnostics) {
...
-				// management, so we'll instead make it required. Users can use
-				// a `random_password` resource with `teleport_provision_token`
-				// to achieve a similar result. See also:
-				// `example/bot.tf.example`
-				Required:    true,
+				// Implementation note: this is not used anymore, we can skip this
+				// This will go away eventually when we'll generate the bot provider instead
+				Optional:    true,
 				Sensitive:   true,
-				Description: "The bot joining token. If unset, a new random token is created and its name returned, otherwise a preexisting Bot token may be provided for IAM/OIDC joining.",
```
hugoShaka, line 77: Note: this description was wrong, it's not possible to unset the token in the TF provider.

strideynet (approved):

r0mant (approved):

PR #1059, opened 2024-04-30
Fix BoolOpts for ProvisionToken

strideynet:
Closes https://github.com/gravitational/teleport-plugins/issues/981

Fixes a fairly nasty bug in the CopyFromBoolOption helper whereby null values were not properly handled. Unfortunately, fixing this generally probably risks breaking folks who rely on the bad behaviour of this function. So for now, we've generated the provision token schema separately and with its own helpers. Eventually we can roll this new behaviour out to all resources.

A good thing to look at would be the different between the `custom_types.go` in this PR and the unchanged one on master.


terraform/Makefile
```diff
@@ -80,16 +92,15 @@ endif
 		--terraform_out=config=protoc-gen-terraform-devicetrust.yaml:./tfschema \
 		teleport/legacy/types/device.proto
 
-	@protoc \
-		-I$(API_MOD_PATH)/proto \
-		-I$(PROTOBUF_MOD_PATH) \
-		--plugin=$(GENTERRAFORMPATH)/protoc-gen-terraform \
-		--terraform_out=config=protoc-gen-terraform-accesslist.yaml:./tfschema \
-		teleport/accesslist/v1/accesslist.proto
+	# @protoc \
```
strideynet, line 95: Commented this out as the access lists are currently incompatible.

hugoShaka:
Superseded by https://github.com/gravitational/teleport-plugins/pull/1065

PR #1072, opened 2024-05-30
Enables debug logging for eventhandler by config and env variable

stevenGravy:
Provides a debug setting within the Teleport Event handler and the environment variable `FDFWD_DEBUG`.

```toml
[teleport]
addr = "teleport.example.com:443"
identity = "example"
debug = true
```

Will update docs and see about helm update after this is added.


stevenGravy:
Closing as this should have been in the teleport repo.

